PR #155, opened 2013-09-26
Make LR self contained (Windows only)

Tieske:
This PR is initially for discussion. 

It adds a Lua binary (5.2) to the `bin` directory, so LuaRocks runs completely independent of the Lua host installation, to reduce chances of interference. Basically the next step after moving the binaries out of the PATH (was also to reduce interference).

Not sure whether is worth the hassle. Just a possibility; maybe install Lua (5.1/5.2/Jit) itself as a rock as LR then no longer depends on Lua being present on the host.

please comment.


ildar:
I know I'm (very) late to get into the discussion, but binaries in git repos irritate me so I can't keep myself out of this.
@Tieske ,
> Though I agree with your comments; untidy and storing binaries being bad. But the alternative is a more complex setup, with limited maintenance capability/resources

You said earlier that it just developers to use git. What's so complicated for a developer to use subtree/subproject(s)?
> (Hisham basically). The gain would be appr. two dozen developers having less stuff in their repo (visibly only, the full history including binaries will still be in their Git repo!).

If you let binaries into the repo then sooner or later you will want to update them. Or (hopefully) you'll understand what a bad idea it was and delete them. Binaries expire. period.

And my last 5 cents: if you did that for win32 why stop? Add win64, MacOS, LuaJIT, Lua6, whatnot, etc.etc. for completeness(?)

PR #246, opened 2014-04-02
run LR itself on the Lua interpreter found during install (Windows)

Tieske:
run LR itself on the Lua interpreter found during install, not on the interpreter first in the system path. Lining up with the unix shell scripts.
See this comment https://github.com/keplerproject/luarocks/pull/243#discussion_r10999753


ignacio:
Can we close this?

Tieske:
yes, please close.

ignacio:
Oh, my bad. This is a pull request, not an issue. There no closing to do here. Sorry for the noise.

PR #252, opened 2014-05-01
[suggestion] INSTALL.BAT Cannot find lua (or anything else) 

andrewstarks:
```
C:\luarocks>install

C:\luarocks>rem=rem --[[
The system cannot find the path specified.
```

After this patch, it works correctly.

--Andrew

### DIFF
--- a/win32pack.bat
+++ b/win32pack.bat
@@ -8,7 +8,7 @@ mkdir %1\test
 mkdir %1\src
 xcopy /S/E .\test\*.* %1\test
 xcopy /S/E .\src\*.* %1\src
-xcopy /S/E .\win32\*.* %1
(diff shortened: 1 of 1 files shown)


hishamhm:
I just downloaded luarocks-2.2.0-win32.zip (on a Windows 8.1 VM with VC2013) and got the same error as @andrewstarks in his first message. Is our Windows installer broken (or perhaps the win32 zip in the server is the incorrect one?...)

hishamhm:
Aaand the zip file in http://keplerproject.github.io/luarocks/releases was really broken. Apparently I messed up during the move to the new site. Things _should_ be fixed now, but I'm off to actually test it.

PR #260, opened 2014-05-18
Error out on bad config files. Fixes #228

### LINKED ISSUE #228: [BUG] config.lua silently ignored when it fails
When the config file fails to load, it is silently ignored and defaults are used. This is ok if the file does not exist, but not ok if the file does exist but cannot be loaded due to a compilation error.

If the file fails to load, but does exist, a warning should be issued.

Tieske:


Tieske:
@hishamhm can we get this one settled? what is your opinion?

hishamhm:
Working on it! Thanks for pinging.

hishamhm:
I took the opportunity to clean up a bit of the mess in the code where I was using nil and false as returns to different error conditions. I think things look nicer now.

Took a while to get this done because I struggled a bit with the test suite.

Issue #262, opened 2014-05-26
[RFC] static build with LuaRocks 3.0


fperrad:
currently, Luarocks build/install shared libraries, for example on linux

```
luarocks install luafilesystem
```

build `lfs.so`, and install it in `/usr/local/lib`

I want that

```
luarocks install --static luafilesystem
```

build `lfs.a` and install it in `/usr/local/lib`
and install `lfs.h` in `/usr/local/inc`

The installation of include files request a new field in the rockspec.
The option `--static` has not effect on pure Lua modules.

xpol:
If this come with support corss compile for iOS and Android, that would be perfect.
Then I can use luarocks as a package manager for iOS and Android projects.

daurnimator:
This was something I wanted today. Has anyone worked on it?

daurnimator:
Oh wait.... was this merged in #264?

siffiejoe:
It was merged, but it is currently commented out:

 >   --[[ TODO disable static libs until we fix the conflict in the manifest, which will take extending the manifest format.

I missed the problem with the manifest, so I can't comment on that ...

jirutka:
I need this too. What is the problem with the manifest?

hishamhm:
@jirutka Because in the manifest file format each module name maps to a single file, like this:

```lua
      ["0.4.1.53-1"] = {
         {
            arch = "installed",
            commands = {},
            dependencies = {},
            modules = {
               gzip = "gzip.lua",
               zlib = "zlib.so"
            }
         }
      }
   },
```

If we build both `zlib.so` and `zlib.a`, only one of them will show up in the manifest. 

Thinking a bit more about it, the rest of the codebase could assume that, if a `.so` file is listed in the manifest, the corresponding `.a` file may or may not be present, and act accordingly. That would be one way of supporting static libs without breaking manifest compatibility.

DarthGandalf:
Hi, any progress here?

I made a workaround for this, which only works with "builtin" rockspecs, which adds the .c files listed in the rockspec to cmake of the project which uses lua as a build target which other targets can use.
https://github.com/AbyssEngine/AbyssEngine/blob/c2fedf572a4f9cf808af816aac6e6c238f28b915/cmake/add_lua_library.lua
https://github.com/AbyssEngine/AbyssEngine/blob/c2fedf572a4f9cf808af816aac6e6c238f28b915/cmake/Lualibs.cmake

justdie386:
[...]

PR #264, opened 2014-05-27
Create static libraries

siffiejoe:
Create static libraries (see #262) when using "builtin" build mode. Seems to work on Linux, Windows/MinGW, and Windows/VC (I haven't tested linking to the generated static libraries yet ...).

Currently no include files are generated ...

### DIFF
--- a/src/luarocks/build/builtin.lua
+++ b/src/luarocks/build/builtin.lua
@@ -50,7 +50,7 @@ end
 -- nil and an error message otherwise.
 function builtin.run(rockspec)
    assert(type(rockspec) == "table")
-   local compile_object, compile_library, compile_wrapper_binary --TODO EXEWRAPPER
+   local compile_object, compile_library, compile_static_library, compile_wrapper_binary --TODO EXEWRAPPER
 
    local build = rockspec.build
    local variables = rockspec.variables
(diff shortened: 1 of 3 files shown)


starius:
Ping.

Can I use this feature to embed Lua modules? There is a program using a rock which uses built-in build mode (e.g., LPeg). I want to get a static binary with no external dependencies. I see the following general solution (probably partly done in this pull request):

1. Create a static target in LuaRocks to prevent static-shared mixture. For example, for `lua-5.1` the static target is `lua-5.1-static`. 

2. A rock should produce a static library which incorporates C code and Lua code. Lua code can be converted into C code (lua2c) and linked together. This static library has only one function callable from outside: `rockopen_<rock>`. When Lua creates new `lua_State` it calls such a function to preload this module. `rockopen` function uses well known Lua mechanisms to enable `require "rock"`. For example, modify `package.preload`.

3. Finally Lua static executable is built. It is linked with all static libraries produced by rocks used. The only difference from normal lua.exe is that it calls functions `rockopen_<rock>` after creating a Lua state.

4. When building libs, C sources are changed to replace `lua_newstate` (and `luaL_newstate`) with a wrapper calling the original function plus `rockopen`s. This enables using rocks from modules creating additional Lua states (e.g., `llthreads`).

Current rockspec format has enough information to implement all these cool features in LuaRocks and to build static executables from Lua scripts with just `luarocks make --static` without modifying rocks.

hishamhm:
> Current rockspec format has enough information to implement all these cool features in LuaRocks and to build static executables from Lua scripts with just `luarocks make --static` without modifying rocks.

...if the rockspec uses the `builtin` type. This is a big "if" (even if 80% of all rocks are `builtin`). Torch people use `build.type="cmake"` a lot, for example.

hishamhm:
Merged in `luarocks-3` branch! As promised in... 2014 :-P

PR #309, opened 2015-02-15
Removed some unused and global variables

mpeterv:
Some tiny fixes. There are a lot more unused variables around but it's not obvious for me how to fix them. E.g. there are a lot of unused error messages (`local ok, err = ...`; `err` is then unused) which should probably be reported instead of simply being removed.

### DIFF
--- a/src/luarocks/manif_core.lua
+++ b/src/luarocks/manif_core.lua
@@ -9,7 +9,6 @@ local persist = require("luarocks.persist")
 local type_check = require("luarocks.type_check")
 local dir = require("luarocks.dir")
 local util = require("luarocks.util")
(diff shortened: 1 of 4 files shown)


hishamhm:
Thanks! Did you check unused variables using a tool? Could you generate a list so I could go through them?

mpeterv:
I used [luacheck](https://github.com/mpeterv/luacheck), here is the list: https://gist.github.com/mpeterv/13457c89e5113d418c14

hishamhm:
Thank you! I'll look into it. BTW, I just integrated luacheck into my text editor :) 
https://github.com/hishamhm/dit/tree/master/scripts

mpeterv:
Awesome!

PR #326, opened 2015-03-18
Fix url redacting when Luasocket is used

mpeterv:
Move definition of redact_api_url up so that it is visible in both
Luasocket and Luasocket-less versions of Api:request


hishamhm:
Oops, thanks for catching it!

Issue #325, opened 2015-03-18
more portable cmake build type support.


xpol:
Currently we use 

```Lua
fs.execute_string(rockspec.variables.CMAKE.." . " ..args
fs.execute_string(rockspec.variables.MAKE.." -fMakefile")
fs.execute_string(rockspec.variables.MAKE.." -fMakefile install")
```

Which use make to build the module.

But cmake have more portable way to build project:

```Bash
cmake -Bbuild -H.
cmake --build build --target install --config Release
```
This will let cmake to choose best generator and build it in a portable way.

This solution have follow advantages:

1. One command to build all platform (with cmake installed).
2. Better for windows, we don't need to run under MSVC command prompt

xpol:
The cmake_generator = "MinGW Makefiles" for windows is not always work.

For me I have a luajit compiled with msvc.

I think when can pass it from command line like:

```
luarocks cmake -G "Visual Studio 12"
```

xpol:
Tested the follow code almost work expect I don't know how to get the path where sould I install dlls and lua files:

```
   if not fs.execute_string(rockspec.variables.CMAKE.." -H. -Bluarocks.build") then
      return nil, "Failed cmake."
   end

   if not fs.execute_string(rockspec.variables.CMAKE.." --build luarocks.build --config Release") then
      return nil, "Failed building."
   end

   if not fs.execute_string(rockspec.variables.CMAKE.." --build luarocks.build --target install --config Release") then
      return nil, "Failed installing."
   end
   return true

```

ignacio:
Do `variables.LUADIR` and `variables.LIBDIR` do what you want?

ignacio:
Also, I think there are a few rocks that currently use the cmake build backend. Does this change affect that rocks? (Although I can't find them. @hishamhm how do you grep over the entire main repository?)

siffiejoe:
@ignacio: Currently only

*  lua-http-parser-1.0-1.rockspec
*  lua-sdl2-2.0.3.3.1-1.rockspec
*  lua-sdl2-2.0.3.3-1.rockspec

from the main repository use the cmake backend. (I wrote a small Lua script that downloads all rockspecs referenced in a given manifest file.)

hishamhm:
@xpol Passing the path where to install .dll and .lua files is up for the rockspec. I believe it depends on how the CMakefile is set up. See for example:

[...]

PR #327, opened 2015-03-18
Better cmake support.

xpol:
1. Let cmake choose the generator.
2. use `cmake --build` to build without needs of  make (nmake mingwin-make ...)

This is a solution for #325 .


hishamhm:
Great, thanks! Just tested here on Linux with the lua-sdl2 rockspec and it worked fine.

PR #329, opened 2015-03-19
Tiny persist.lua refactorings

mpeterv:
This fixes some small issues with `luarocks.persist` module:

* `persist.load_into_table` (used for loading rockspecs) used to read file twice on Lua 5.2/5.3;
* `persist.save_from_table` (used for saving manifests, etc.) used to produce broken code when table contained strings with backslashes or Windows-like newlines.

### DIFF
--- a/src/luarocks/persist.lua
+++ b/src/luarocks/persist.lua
@@ -9,8 +9,14 @@ package.loaded["luarocks.persist"] = persist
 
 local util = require("luarocks.util")
 
+--- Load and run a Lua file in an environment.
+-- @param filename string: the name of the file.
+-- @param env table: the environment table.
+-- @return (true, any) or (nil, string, string): true and the return value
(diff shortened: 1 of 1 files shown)


hishamhm:
Cool!

Since I gave you commit access recently, I take it as a request for code review. :) The write_table change looks good, thanks!

For fixes like the `load_into_table` change, feel free to commit directly. (In fact, since this whole PR fixes behavior only and doesn't change APIs, I think you could have committed directly the whole thing, but thanks for asking anyway :) )

mpeterv:
> For fixes like the load_into_table change, feel free to commit directly. (In fact, since this whole PR fixes behavior only and doesn't change APIs, I think you could have committed directly the whole thing,

Great, will do! I just decided to be careful the first time since some teams require that even small changes are approved by at least one collaborator other than the author.

PR #340, opened 2015-03-26
Revert incorrect remove of cmake_generator support.  ...

xpol:
Only windows msvc default cmake_generator are removed.
Windows mingw and cygwin cmake_generator default are kept.


hishamhm:
Thank you!

PR #343, opened 2015-04-01
Add options to hide the MSVC tools' startup logo.

xpol:


hishamhm:
@ignacio seems to be using this flag in his AppVeyor scripts and it's not the first time I see `/nologo` thrown around, so I guess it's ok to merge. Anyone has comments?

PR #345, opened 2015-04-03
Add support for git submodules

mpeterv:
When fetching rock sources using git, run `git submodule update --init --recursive` after cloning. When using git >= 1.8.4, also pass `--depth=1` to fetch only the last commit of each submodule.

### DIFF
--- a/src/luarocks/fetch/git.lua
+++ b/src/luarocks/fetch/git.lua
@@ -7,20 +7,48 @@ local unpack = unpack or table.unpack
 
 local fs = require("luarocks.fs")
 local dir = require("luarocks.dir")
+local deps = require("luarocks.deps")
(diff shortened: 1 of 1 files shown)


daurnimator:
Can't we just pass --recursive to clone?

mpeterv:
@daurnimator `--recursive` is only supported since git 1.6.5, and it ignores `--depth` (that is, `git clone --recursive --depth=1` clones main repo shallowly, but not the submodules.

hishamhm:
This can be merged only in the LuaRocks 3.0 tree when we bump up the rockspec_format version number

mpeterv:
@hishamhm I see that this change can be breaking, but why rockspec_format must be bumped? I didn't add any rockspec options.

PR #346, opened 2015-04-08
bail out on bad config files, fixes #228

### LINKED ISSUE #228: [BUG] config.lua silently ignored when it fails

Tieske:


src/luarocks/persist.lua
```diff
@@ -53,10 +53,10 @@ end
 -- @param filename string: the name of the file.
 -- @param tbl table or nil: if given, this table is used to store
 -- loaded values.
--- @return (table, table) or (nil, string, string): a table with the file's
--- assignments as fields and set of undefined globals accessed in file,
--- or nil, an error message and an error code ("open", "load" or "run") in
--- case of errors.
+-- @return table or (nil, string, string): a table with the file's assignments
```
mpeterv: Why did you remove info on the second normal return value?

hishamhm: My bad, that comment was lingering in the wrong function.

Issue #287, opened 2014-10-02
Command to install dependencies


tbastos:
I would like to use LuaRocks to install the dependencies listed in a rockspec without starting a full build/install. This is handy in build scripts that must integrate LuaRocks with other build tools.

For example, I need to:
1. Install project dependencies.
2. Build and run tests using other tools (not LuaRocks).
3. Only then use LuaRocks to create a rock.

I want to avoid repeating my dependencies both in the rockspec and in the build script (where I'd have to write many "sudo luarocks install x" lines).

In Ruby I may have a "Gemfile" within a project and run the command `bundle` to install dependencies and prepare the environment to build the project. Is there anything similar in luarocks?

Thanks

ignacio:
A new flag (--only-deps ?) should be added in order to do that. It should be forwarded to the build command as well and it should stop after fulfilling the dependencies.

@hishamhm is that correct?

ignacio:
Also, you may give [Moonbox](https://github.com/kernelp4nic/moonbox) a go.

hishamhm:
Yes, this could be possibly added as a flag to `build` without adding a lot of code.

ignacio:
Fixed by #349

PR #349, opened 2015-04-14
Adds --only-deps flag to the 'build' command.

ignacio:
As discussed in #287, adds a new flag (--only-deps) to the 'build' command, so only the dependencies of a rock are installed, and not the rock itself.

Feedback is welcome. Specially, I don't know if I should add this [also here](https://github.com/keplerproject/luarocks/blob/master/src/luarocks/build.lua#L371). I think it is best not to add it there, since 'install' removes the installed rock and we don't want that.

### DIFF
--- a/src/luarocks/install.lua
+++ b/src/luarocks/install.lua
@@ -26,6 +26,8 @@ or a filename of a locally available rock.
                     rock after installing a new one. This behavior can
                     be made permanent by setting keep_other_versions=true
                     in the configuration file.
+
+--only-deps         Installs only the dependencies of the rock.
 ]]..util.deps_mode_help()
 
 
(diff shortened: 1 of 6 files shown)


ignacio:
Sorry for all the commits. I intend to squash them before merging.

ignacio:
Appveyor build failed due to a transient network error.

hishamhm:
If we're adding --only-deps, I think it needs to be added to `luarocks install` as well. In luarocks.install this would be better as a separate function, since reusing install.install_binary_rock would require greater changes.

ignacio:
Thanks. Will look into that. Also, any tips on the test? I have the problem that if the first command (luarocks build --only-deps) fails, the test passes by mistake. Is it ok to add something like set -e in a function?

hishamhm:
I think chaining them with && does the trick.

```
test_build_only_deps_rockspec() { $luarocks download --rockspec lxsh ${verrev_lxsh} && $luarocks build ./lxsh-${verrev_lxsh}.rockspec --only-deps && { $luarocks show lxsh; [ $? -ne 0 ]; }; }
```

I've been writing all tests as single-lines because I used to grep the script, pipe and run it, but I think it's not necessary anymore.

Ah, please add them to testing.lua as well. I know it's unfinished and we still can't run it (so you'll have to guess the API a little), but I don't want to keep the list of tests out of sync until we finish the port and retire the bash version.

ignacio:
How about now?

src/luarocks/install.lua
```diff
@@ -109,6 +111,43 @@ function install.install_binary_rock(rock_file, deps_mode)
...
+   local rockspec, err, errcode = fetch.load_rockspec(path.rockspec_file(name, version))
+   if err then
+      return nil, "Failed loading rockspec for installed package: "..err, errcode
+   end
+
+   ok, err, errcode = deps.fulfill_dependencies(rockspec, deps_mode)
+   if err then return nil, err, errcode end
+
+   util.printout()
+   util.printout("Succesfully nstalling dependencies for " ..name.." "..version)
```
hishamhm: typo! → _i_nstalling

test/testing.lua
```diff
@@ -140,6 +140,18 @@ local tests = {
          and rm "./validate-args-${verrev_validate_args}.rockspec"
    end,
    test_build_supported_platforms = function() return run "$luarocks build lpty" end,
+   test_build_only_deps_rockspec = function()
+      return run "$luarocks download --rockspec lxsh ${verrev_lxsh}"
+         and run "$luarocks build ./lxsh-${verrev_lxsh}.rockspec --only-deps"
+         and run "$luarocks show lxsh; [ $? -ne 0 ]; };"
```
hishamhm: and (not run "$luarocks show lxsh")

test/testing.sh
```diff
@@ -262,6 +263,7 @@ mkdir -p "$testing_server"
    get "$luarocks_repo/lua-path-0.2.3-1.src.rock"
    get "$luarocks_repo/lua-cjson-2.1.0-1.src.rock"
    get "$luarocks_repo/luacov-coveralls-0.1.1-1.src.rock"
+   get "$luarocks_repo/luacov-coveralls-0.1.1-1.src.rock"
```
hishamhm: duplicated line?

hishamhm:
Apart from the minor notes above, looks good to merge!

test/testing.sh
ignacio: So it seems.

ignacio:
Thanks a lot!

PR #350, opened 2015-04-14
Add git+https support.

jszakmeister:


jszakmeister:
FWIW, I modeled this after what was done in `hg_https.lua`.  I hope it's the right approach.

mpeterv:
LGTM. @hishamhm by the way, why `--depth=1` is not used when fetching `git+http://`? [1](https://github.com/keplerproject/luarocks/blob/0fe8556ced63d714e0cd371a8511c8dadce7874e/src/luarocks/fetch/git_http.lua#L24) [2](https://github.com/keplerproject/luarocks/blob/0fe8556ced63d714e0cd371a8511c8dadce7874e/src/luarocks/fetch/git.lua#L57)

ignacio:
@mpeterv because it depends on the git server. See [this page](http://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols) on protocols. The old "dumb" protocol does not support the _depth_ flag.

~~~~
$ git --version
git version 1.9.5.msysgit.0
$ git clone --depth=1 http://luajit.org/git/luajit-2.0.git luajit-2.0
Cloning into 'luajit-2.0'...
fatal: dumb http transport does not support --depth
~~~~

However, some git servers implement a "smart" http protocol. Github, for instance, supports the --depth flag with no problems.

~~~~
$ git clone --depth=1 http://github.com/keplerproject/luarocks.git luarocks
Cloning into 'luarocks'...
remote: Counting objects: 193, done.
remote: Compressing objects: 100% (169/169), done.
remote: Total 193 (delta 16), reused 123 (delta 11), pack-reused 0
Receiving objects: 100% (193/193), 3.29 MiB | 158.00 KiB/s, done.
Resolving deltas: 100% (16/16), done.
Checking connectivity... done.
~~~~

I couldn't find a way to determine the version of the git server, so if we were to support this flag over these transports, maybe we could have a whitelist of known git services that support the --depth flag but I don't know if it is worth the hassle.  /cc @hishamhm

PR #360, opened 2015-04-20
Print a warning and stop if the old url is present.

ignacio:
When trying to upload, and your _upload_config.lua_ file still has the old url, print a warning and exit (not so cleanly, it wrongly reports that you're missing the api key, which, probably, you're not).

~~~
$ luarocks upload rockspecs/nozzle-git-1.rockspec

Warning: The main rocks repository is now located at https://luarocks.org
Edit your upload configuration at 'C:/Users/Ignacio/AppData/Roaming/luarocks/upload_config.lua'

Error: You need an API key to upload rocks.
Navigate to https://luarocks.org/settings to get a key
and then pass it through the --api-key=<key> flag.
~~~


ignacio:
This is not needed anymore.

PR #361, opened 2015-04-20
Smooth upgrade

ignacio:
When trying to upload, and your _upload_config.lua_ file still has the old url, rewrites the configuration to use the new url.


ignacio:
This is not needed anymore.

Issue #367, opened 2015-05-11
"libraries" parsing should allow for custom flags when value starts with "-"


gaspard:
In "libraries", If the library name starts with "-", it should not be added to the link command with an extra "-l". This bug is a show stopper for libraries requiring native frameworks on Mac OS X.

If you provide something like this:
````
 macosx = {
      modules = {
        ['lens.core'] = {
          sources = {
            [13] = 'src/macosx/poller.mm',
          },
          libraries = {'stdc++', '-framework Foundation', '-framework Cocoa', 'objc'},
        },
      },
    },    
````

You can an error like this:

````
ld: library not found for -l-framework Foundation
````

I can provide a patch and pull request if needed.

deepakjois:
I just ran into this requirement when trying to create LuaRocks module that needs to be compiled with `-framework CoreFoundation` on OS X. I hope this will be possible in LuaRocks 3.

FWIW, I checked node-gyp and how it handles this requirement for OS X. Here is the relevant code:
https://github.com/nodejs/node-gyp/blob/f6eab1f9e4b160b60b2736bb01a09c83785ba979/gyp/pylib/gyp/xcode_emulation.py#L1024

osch:
I tried to solve the above problem, see https://github.com/osch/luarocks-build-extended#luarocks-build-extended

Issue #129, opened 2013-04-30
Add custom compilation flags


mtourne:
I wish it would be possible to add custom CFLAGS from the command line, or in the rockspec file directly.

Currently setting CFLAGS=-g to add debug symbol to a particular lib will overwrite _all_ of the original CFLAGS and it won't build anymore.

Or maybe I've missed an obvious way to do that ?

jasonwilliams200OK:
> C99 is 15 years old and all the do is partial suport of C99 Library in MSVC13

In VS2015, the story is different. C99 support is almost done except for (majorly) tgmath.h which isn't popular anyway. Even C11 support is partially available. Some niche nix libraries which were never made possible to compile with VS before, have started to support VS2015 by strictly using ISO C99 standard features 

So it is now very possible if VS2015 is set as minimum requirement to build for Windows (XP-10) without sacrificing your existing C99 code and doubling the effort. If you are someone who don't want to install VS2015, then you can either use MinGW/Clang/CigWin to build for Windows or configure AppVeyor CI to build and make artifacts available for you to download from CI server. I don't see any point of keep supporting VS2003-VS2013 at this point in 2015 for any library dev, who prefers C99. 

C-support can only get better in future versions of VS.

craigbarnes:
> Currently setting CFLAGS=-g to add debug symbol to a particular lib will overwrite all of the original CFLAGS and it won't build anymore.

If a lib is using the `CFLAGS` variable to store essential compiler flags, then that lib's build system is broken.

> If there are C compiler options that must be used for proper compilation of certain files, do not include them in `CFLAGS`. Users expect to be able to specify `CFLAGS` freely themselves. Instead, arrange to pass the necessary options to the C compiler independently of `CFLAGS`, by writing them explicitly in the compilation commands or by defining an implicit rule, like this:
>
>```
>CFLAGS = -g
>ALL_CFLAGS = -I. $(CFLAGS)
>.c.o:
>        $(CC) -c $(CPPFLAGS) $(ALL_CFLAGS) $<
>```

-- https://www.gnu.org/prep/standards/html_node/Command-Variables.html

This convention goes back decades. Most Linux distro packaging scripts overwrite `CFLAGS` by default.

daurnimator:
I think this issue deserves some attention. @hishamhm is this now possible with the rockspec 3.0 format?

osch:
I tried to solve the above problem, see https://github.com/osch/luarocks-build-extended#luarocks-build-extended

[...]

PR #368, opened 2015-05-11
Enable platform specific raw flags for compile and link steps with 'cflags' and 'lflags'.

gaspard:
Hi there !

This fixes issues #367 and #129.

It is a clean change that does not break compatibility with previous luarocks versions.

This change would allow releasing libraries such as lui (minimal ui library) and lens (scheduler) as these require Mac OS X frameworks. I have some other installation issues for Windows that could be solved by using platform specific flags.

Cheers !

Gaspard

### DIFF
--- a/src/luarocks/build/builtin.lua
+++ b/src/luarocks/build/builtin.lua
@@ -68,16 +68,18 @@ function builtin.run(rockspec)
    end
 
    if cfg.is_platform("mingw32") then
-      compile_object = function(object, source, defines, incdirs)
+      compile_object = function(object, source, defines, incdirs, rawflags)
          local extras = {}
          add_flags(extras, "-D%s", defines)
          add_flags(extras, "-I%s", incdirs)
+         add_flags(extras, "%s", rawflags)
          return execute(variables.CC.." "..variables.CFLAGS, "-c", "-o", object, "-I"..variables.LUA_INCDIR, source, unpack(extras))
       end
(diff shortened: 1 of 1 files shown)


hishamhm:
Hi!

> It is a clean change that does not break compatibility with previous luarocks versions.

Unfortunately it does break compatibility because it extends the grammar of the rockspec format. In other words, rockspecs using this feature will not work on older LuaRocks releases.

I've been meaning to create the luarocks-3 branch for a while now (in which features like this will be welcome). I won't be able to do it this week as I'm focusing on other work right now, but my plan is to do it later this month (preferrably before the start of the Google Summer of Code coding phase, so that the student works his branch off the new "hot" codebase).

hishamhm:
(Never mind the Travis CI failure. It was for an unrelated reason)

I'll keep this on hold until we get the new branch. Thanks for submitting!

gaspard:
Hmmm... I am not sure it "breaks compatibility" because if these flags are
not present (as in previous versions), they are simply ignored.

Or do you mean people submitting build tables that contain keys not in the
current spec ? In this case, yes, weird things could happen but it seems
very far fetched.

But maybe, you simply mean that it changes the semantics and therefore
needs a new version number (as in http://semver.org/).

Cheers and thanks for responding so fast :-)

Gaspard



Gaspard Bucher
teti sàrl

PR #379, opened 2015-05-31
Ignore missing build table when using --only-deps

Alloyed:
When using `luarocks build --only-deps`, don't check to see if the rockspec has a build table, because it won't get built anyways.
This is useful for users who would like to use luarocks exclusively for dependency management, and not to produce a final .rock

### DIFF
--- a/src/luarocks/build.lua
+++ b/src/luarocks/build.lua
@@ -169,10 +169,14 @@ function build.build_rockspec(rockspec_file, need_to_fetch, minimal_mode, deps_m
    local rockspec, err, errcode = fetch.load_rockspec(rockspec_file)
    if err then
       return nil, err, errcode
-   elseif not rockspec.build then
-      return nil, "Rockspec error: build table not specified"
-   elseif not rockspec.build.type then
-      return nil, "Rockspec error: build type not specified"
+   end
+
+   if not build_only_deps then
+      if not rockspec.build then
+         return nil, "Rockspec error: build table not specified"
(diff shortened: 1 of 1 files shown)


ignacio:
Why would you use LuaRocks this way? A "real world" use case for this would be nice.

hishamhm:
@Alloyed, can't you simply use 
```
build = { type = "none" }
```
?

Alloyed:
`build = { type = 'none' }` works perfectly, actually.

That said, if it's intended that a rockspec always have an explicit build table, it'd be nice if `luarocks lint` also enforced it and if the docs said something along those lines.

>Why would you use LuaRocks this way? A "real world" use case for this would be nice.
I plan on using luarocks with [LOVE](http://love2d.org), which uses its own package format. One of the requirements is that love packages be standalone, ie. they can't depend on any external packages. So in order to keep using the usual luarocks infrastructure, I install everything besides my code into a project-local rocks tree, and then handle the rest manually.

ignacio:
@hishamhm do you want to add a check in the lint command for this?

hishamhm:
sure, extra linting is always welcome

ignacio:
Will making the "build" key mandatory in type_check.lua do?

hishamhm:
I believe so

ignacio:
Please take a look at #394. Added new test for it.

ignacio:
Closed in favor of #394

PR #383, opened 2015-06-08
cmake backend: Generate 64 bits build when appropiate

### LINKED ISSUE #382: cmake build does not generate 64 bit build for windows.
cmake build does not generate 64 bit build for windows when luarocks installed with a 64 bit lua.

ignacio:
Adds -DCMAKE_GENERATOR_PLATFORM=x64 when needed. It seems that CMake does not do that by default (although I might be wrong on that).

fixes #382

### DIFF
--- a/src/luarocks/build/cmake.lua
+++ b/src/luarocks/build/cmake.lua
@@ -37,16 +37,16 @@ function cmake.run(rockspec)
 
    -- Execute cmake with variables.
    local args = ""
-   if cfg.cmake_generator then
+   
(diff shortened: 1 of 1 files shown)


ignacio:
Again, this is not ok. The decision to choose 32/64 target depends on whether the Lua binary in use is 32/64 bits.

ignacio:
After #397 has been merged, this one is good to go.

ignacio:
Rebased to master.

PR #384, opened 2015-06-10
Add version-specific overrides, similar to platform-specific overrides

dpapavas:
This pull request adds version-specific overrides, which work and are implemented in a way almost identical to platform-specific overrides.  This feature would allow one to tailor the build process according to the Lua version.

My use case for this is, that I'd like to be able to append the Lua version to an executable script that's installed alongside a module, so that multiple versions of the script, for multiple vesions of Lua can be installed side-by-side in the same bin directory.  I imagine other uses will exist.

Note that as the overrides are applied after platform overrides one should be able to combine these, putting version overrides within platform overrides.

### DIFF
--- a/src/luarocks/fetch.lua
+++ b/src/luarocks/fetch.lua
@@ -212,6 +212,12 @@ function fetch.load_local_rockspec(filename, quick)
    util.platform_overrides(rockspec.source)
    util.platform_overrides(rockspec.hooks)
 
+   util.version_overrides(rockspec.build)
+   util.version_overrides(rockspec.dependencies)
+   util.version_overrides(rockspec.external_dependencies)
+   util.version_overrides(rockspec.source)
+   util.version_overrides(rockspec.hooks)
+
    local basename = dir.base_name(filename)
    if basename == "rockspec" then
       rockspec.name = rockspec.package:lower()
--- a/src/luarocks/util.lua
+++ b/src/luarocks/util.lua
@@ -280,6 +280,31 @@ function util.platform_overrides(tbl)
    tbl.platforms = nil
 end
 
+--- Perform version-specific overrides on a table.
+-- This works exactly like platform-specific overrides, with the
+-- exception that the keys of the "versions" table are matched against
+-- the cfg.lua_version field.  If a match is found, all fields in the
+-- matched table overwrite those of tbl, as for platform-specific
+-- overrides.
+-- @param tbl table or nil: Table which may contain a "versions" field;
+-- if it doesn't (or if nil is passed), this function does nothing.
+function util.version_overrides(tbl)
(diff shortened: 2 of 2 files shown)


hishamhm:
Sorry about the long delay in reviewing this. 

This is a great feature for LuaRocks 3. I'm looking into merging this to the `luarocks-3` branch. But there is still some work needed:

[ ] Support in `luarocks.type_check` to `versions` entries, similar to the one we have for `platforms`.
[ ] Some tests in the `test/testing.sh` test suite (to test coverage of the added parts — some testing would have caught the absence of type checking above)

dpapavas:
Is there anything you'd like me to look into?

hishamhm:
Yes, definitely! Thanks for the offer! I can't focus on this right now, so if you implement the two missing parts I outlined above (type check and tests), this feature will be ready to merge into `luarocks-3`.

dpapavas:
Sorry for taking so long to report back.  I've added the type checks and tested a bit (manually), to make sure they work.  I've also tested the feature by adding the following to my rockspec for `luaprompt`:

```
versions = {
       ["5.3"] = {
          install = {
             bin = {
                ["luap-5.3"] = "luap.lua"
             }
          }
       },

       ["5.2"] = {
          install = {
             bin = {
                ["luap-5.2"] = "luap.lua"
             }
          }
       },

       ["5.1"] = {
          install = {
             bin = {
                ["luap-5.1"] = "luap.lua"
             }
          }
       }
```

This had the intended result of adding a luap-5.x executable to `/usr/local/bin`, in addition to the `luap` executable copied by the global `install` field:

```
    install = {
        bin = {
            luap = 'luap.lua'
        },
    },
```

By the way, there's a minor problem, in that each different version, in overwriting the possibly existing `luap` script, creates a `luap~` backup, which ends up polluting the `bin` directory.  Is there any way of disabling this backup generation?

I noticed that the `platforms` field of the `build` table, is included in the `type_check.rockspec_order` table.  As far as I can tell, this table specifies any needed ordering of fields, when writing back rockspecs during the `write_rockspec` and `new_version` commands.  I don't see why the `versions` field would be sensitive to order, but then again I don't see why that would be the case for the `platforms` field either.  I tried `new_version` on a my `versions`-enabled luaprompt rockspec and it seemed to work fine.  Let me know if I'm missing anything.

Finally, regarding testing, I don't see any special testing of the platform override functionality, so I'm a bit unclear on how to proceed.  I see that I'll need a rockspec to test with, but there's none with a `versions` field already available, of course.  Should I put one in test/testfiles?  Is there perhaps some other approach I'm missing?  Also should I just test that `luarocks` can read a properly formatted `versions` field, or should I also make sure that the overrides are in fact implemented?

PR #385, opened 2015-06-16
Read os variables from config files

Tieske:
added 2 functions to the global environment of the config-file loader sandbox; `os_getenv()` and `__dump_env()`

The first is the same as the `os.getenv()` standard function. I didn't create a sub table as it is passed by reference and would allow one to override it, which might then possibly be reused for loading another config file.

The latter is for setup/debug purposes and will dump the environment passed to the config file. So adding `__dump_env()` line to `config.lua` and then execute `luarocks` to try it. It will dump all available input for the config file.

example results;
````
Microsoft Windows [versie 6.1.7601]
Copyright (c) 2009 Microsoft Corporation. Alle rechten voorbehouden.

C:\Users\Thijs>luarocks
global environment = {
   ["program_series"] = "2.2";
   ["lua_version"] = "5.1";
   ["errorcodes"] = {
      ["OK"] = 0;
      ["CONFIGFILE"] = 3;
      ["CRASH"] = 99;
      ["PERMISSIONDENIED"] = 2;
      ["UNSPECIFIED"] = 1;
   };
   ["verbose"] = "false";
   ["__dump_env"] = "function: 004821F8, defined in (150-154)@C:\\Program Files (x86)\\LuaRocks\\2.2\\lua\\luarocks\\cfg
.lua";
   ["site_config"] = {
      ["LUA_INCDIR"] = "C:\\Program Files (x86)\\Lua\\5.1\\include\\";
      ["LUAROCKS_UNAME_M"] = "x86";
      ["LUAROCKS_SYSCONFIG"] = "C:\\Program Files (x86)\\LuaRocks\\config.lua";
      ["LUAROCKS_ROCKS_TREE"] = "C:\\Program Files (x86)\\LuaRocks\\systree";
      ["LUAROCKS_MD5CHECKER"] = "md5sum";
      ["LUAROCKS_DOWNLOADER"] = "wget";
      ["LUA_INTERPRETER"] = "lua.exe";
      ["LUA_LIBDIR"] = "C:\\Program Files (x86)\\Lua\\5.1\\bin\\";
      ["LUAROCKS_PREFIX"] = "C:\\Program Files (x86)\\LuaRocks";
      ["LUAROCKS_UNAME_S"] = "MINGW";
      ["LUA_BINDIR"] = "C:\\Program Files (x86)\\Lua\\5.1\\bin\\";
   };
   ["program_version"] = "scm";
   ["home"] = "C:\\Users\\Thijs\\AppData\\Roaming";
   ["rocks_trees"] = {
      [1] = {
         ["name"] = "user";
         ["root"] = "C:\\Users\\Thijs\\AppData\\Roaming/luarocks";
      };
      [2] = {
         ["name"] = "system";
         ["root"] = "C:\\Program Files (x86)\\LuaRocks\\systree";
      };
   };
   ["variables"] = {
      ["LUALIB"] = "lua51.dll";
      ["MSVCRT"] = "m";
   };
   ["os_getenv"] = "function: 008E99B0, C function";
   ["major_version"] = "2.2";
   ["home_tree"] = "C:\\Users\\Thijs\\AppData\\Roaming/luarocks/";
};


LuaRocks scm, a module deployment system for Lua

NAME
        C:\Program Files (x86)\LuaRocks\2.2\luarocks.lua - LuaRocks main command-line interface

<snip>
````

### DIFF
--- a/src/luarocks/cfg.lua
+++ b/src/luarocks/cfg.lua
@@ -145,28 +145,7 @@ end
 cfg.variables = {}
 cfg.rocks_trees = {}
 
--- some extras for the global environment in the config files;
-cfg.os_getenv = os.getenv
-cfg.__dump_env = function()
-  -- debug function, calling it from a config file will show all 
-  -- available globals to that config file
-  print(util.show_table(cfg, "global environment"))
-end
-
-sys_config_file = site_config.LUAROCKS_SYSCONFIG or sys_config_dir.."/config-"..cfg.lua_version..".lua"
-do
-   local err, errcode
-   sys_config_ok, err, errcode = persist.load_into_table(sys_config_file, cfg)
-   if (not sys_config_ok) and errcode == "open" then -- file not found, so try alternate file
-      sys_config_file = sys_config_dir.."/config.lua"
-      sys_config_ok, err, errcode = persist.load_into_table(sys_config_file, cfg)
-   end
-   if (not sys_config_ok) and errcode ~= "open" then -- either "load" or "run"; bad config file, bail out with error
-      io.stderr:write(err.."\n")
-      os.exit(cfg.errorcodes.CONFIGFILE)
-   end
-end
-
+-- The global environment in the config files;
 local env_for_config_file
 env_for_config_file = {
    home = cfg.home,
@@ -174,12 +153,26 @@ env_for_config_file = {
    platform = util.make_shallow_copy(detected),
    processor = proc,
    os_getenv = os.getenv, 
-   __dump_env = function()
+   dump_env = function()
      -- debug function, calling it from a config file will show all 
      -- available globals to that config file
      print(util.show_table(env_for_config_file, "global environment"))
    end,
 }
+
+sys_config_file = site_config.LUAROCKS_SYSCONFIG or sys_config_dir.."/config-"..cfg.lua_version..".lua"
+do
+   local err, errcode
+   sys_config_ok, err, errcode = persist.load_into_table(sys_config_file, env_for_config_file)
+   if (not sys_config_ok) and errcode == "open" then -- file not found, so try alternate file
+      sys_config_file = sys_config_dir.."/config.lua"
+      sys_config_ok, err, errcode = persist.load_into_table(sys_config_file, env_for_config_file)
+   end
+   if (not sys_config_ok) and errcode ~= "open" then -- either "load" or "run"; bad config file, bail out with error
+      io.stderr:write(err.."\n")
+      os.exit(cfg.errorcodes.CONFIGFILE)
+   end
+end
 
 if not site_config.LUAROCKS_FORCE_CONFIG then
 
--- a/src/luarocks/cfg.lua
+++ b/src/luarocks/cfg.lua
@@ -145,7 +145,7 @@ end
 cfg.variables = {}
 cfg.rocks_trees = {}
 
--- some extras for the global enviornment in the config files;
+-- some extras for the global environment in the config files;
 cfg.os_getenv = os.getenv
 cfg.__dump_env = function()
   -- debug function, calling it from a config file will show all 
--- a/src/luarocks/cfg.lua
+++ b/src/luarocks/cfg.lua
@@ -145,6 +145,14 @@ end
 cfg.variables = {}
 cfg.rocks_trees = {}
 
+-- some extras for the global enviornment in the config files;
+cfg.os_getenv = os.getenv
+cfg.__dump_env = function()
+  -- debug function, calling it from a config file will show all 
+  -- available globals to that config file
+  print(util.show_table(cfg, "global environment"))
+end
+
 sys_config_file = site_config.LUAROCKS_SYSCONFIG or sys_config_dir.."/config-"..cfg.lua_version..".lua"
 do
    local err, errcode
@@ -159,11 +167,18 @@ do
    end
 end
 
-local env_for_config_file = {
+local env_for_config_file
+env_for_config_file = {
    home = cfg.home,
    lua_version = cfg.lua_version,
    platform = util.make_shallow_copy(detected),
    processor = proc,
+   os_getenv = os.getenv, 
+   __dump_env = function()
+     -- debug function, calling it from a config file will show all 
+     -- available globals to that config file
+     print(util.show_table(env_for_config_file, "global environment"))
+   end,
 }
 
 if not site_config.LUAROCKS_FORCE_CONFIG then

### CODE REVIEW

File: src/luarocks/cfg.lua [hunk 1]
```diff
@@ -145,6 +145,14 @@ end
 cfg.variables = {}
 cfg.rocks_trees = {}
 
+-- some extras for the global enviornment in the config files;
```
hishamhm: typo: environment

src/luarocks/cfg.lua
```diff
@@ -159,11 +167,18 @@ do
...
 end
 
-local env_for_config_file = {
+local env_for_config_file
+env_for_config_file = {
    home = cfg.home,
    lua_version = cfg.lua_version,
    platform = util.make_shallow_copy(detected),
    processor = proc,
+   os_getenv = os.getenv,
```
hishamhm: maybe `os = { getenv = os.getenv }`, so that users get a familiar interface?

src/luarocks/cfg.lua
```diff
@@ -159,11 +167,18 @@ do
...
 
-local env_for_config_file = {
+local env_for_config_file
+env_for_config_file = {
    home = cfg.home,
    lua_version = cfg.lua_version,
    platform = util.make_shallow_copy(detected),
    processor = proc,
+   os_getenv = os.getenv, 
+   __dump_env = function()
```
hishamhm: I'm not a fan of using "__" as a prefix (clashes with Lua's standard stuff in other places).

src/luarocks/cfg.lua
ignacio: @Tieske decided against that in order to not let a user change os.getenv for another thing.

src/luarocks/cfg.lua
```diff
@@ -145,6 +145,14 @@ end
...
 cfg.rocks_trees = {}
 
+-- some extras for the global enviornment in the config files;
+cfg.os_getenv = os.getenv
+cfg.__dump_env = function()
+  -- debug function, calling it from a config file will show all 
+  -- available globals to that config file
+  print(util.show_table(cfg, "global environment"))
+end
+
```
hishamhm: are these necessary? `cfg` is only used internally by LuaRocks where these things are already available.

src/luarocks/cfg.lua
hishamhm: ah yes, I saw the code for the pull request first and then read the description above later

hishamhm:
Your example dump got me confused. That global environment shown is the full environment of `cfg`, right? Not the one from `env_for_config_file`.

src/luarocks/cfg.lua
Tieske: fixed that, coming up!

src/luarocks/cfg.lua
Tieske: I tend to use it only for some 'hidden' debug stuff, any preferred alternative?

Tieske:
Glad you asked, I was wondering the same. It is `cfg`, because the system config files are loaded [here](https://github.com/Tieske/luarocks/blob/master/src/luarocks/cfg.lua#L156-168) using the `cfg` table.

`env_for_config_file` is only used [further down](https://github.com/Tieske/luarocks/blob/master/src/luarocks/cfg.lua#L188-197)

I was wondering whether there was a reason for this, or that the system config file also should have been loaded using `env_for_config_file` (and hence its definition should move up).

So is there a reason?

hishamhm:
> So is there a reason?

No, I think this happened by accident over time. If we look at the source code waaay waaay back...

https://github.com/keplerproject/luarocks/blob/a88d6f2eeba2b3355c33fac6d736cf6086342f47/src/luarocks/cfg.lua

...the sys config file was loaded over the module table, and the home config file was loaded over a new table, as they are now. *But* the sys config table at that point contained only `variables` and `rocks_trees`, and the home overrides table was simply `{ home = home }`.

Turns out that we started to add stuff to the module table earlier on, and those "leaked" into the sys config loading environment.

So yes, you're right: sys config loading should use `env_for_config_file`.

Tieske:
I'll have a stab at fixing it (in this PR). Meanwhile, what's the preferred alternative for the `__` prefix of `__dump_env()` ?

hishamhm:
I'd just leave it as dump_env (and would later remove the functions from the table so they are not merged into cfg).

Tieske:
> and would later remove the functions from the table so they are not merged into cfg

that won't be necessary, as the table will be used as a metatable on the global environment, so its content can be accessed, but modifications to it will land in the actual global environment

Tieske:
updated it. One other thing comes to mind; the processor architecture is listed, shouldn't we also export 32/64-bitness of the Lua engine used to the config files?

ignacio:
Will it also be available in `luarocks.cfg` ? I'd love that. I had to [hack my way](https://github.com/ignacio/luarocks/blob/22281450a85a40aa2bd600d2fccf16c9e3261c19/src/luarocks/build/cmake.lua#L43) to get that info.

siffiejoe:
> One other thing comes to mind; the processor architecture is listed, shouldn't we also export 32/64-bitness of the Lua engine used to the config files?

Maybe that should be part of the architecture (same with the C runtime version used), so that binary rocks work again on Windows ...

hishamhm:
Isn't it already available via `processor`?

Tieske:
not on my system;
````
C:\Users\Thijs>luarocks
global environment = {
   ["processor"] = "x86";
   ["platform"] = {
      ["windows"] = "true";
      ["mingw32"] = "true";
   };
   ["verbose"] = "false";
   ["os_getenv"] = "function: 005199B0, C function";
   ["home"] = "C:\\Users\\Thijs\\AppData\\Roaming";
   ["rocks_trees"] = {
      [1] = {
         ["name"] = "user";
         ["root"] = "C:\\Users\\Thijs\\AppData\\Roaming/luarocks";
      };
      [2] = {
         ["name"] = "system";
         ["root"] = "C:\\Program Files (x86)\\LuaRocks\\systree";
      };
   };
   ["variables"] = {
      ["LUALIB"] = "lua51.dll";
      ["MSVCRT"] = "m";
   };
   ["lua_version"] = "5.1";
   ["dump_env"] = "function: 004822C0, defined in (156-160)@C:\\Program Files (x86)\\LuaRocks\\2.2\\lua\\luarocks\\cfg.l
ua";
};


LuaRocks scm, a module deployment system for Lua

NAME
        C:\Program Files (x86)\LuaRocks\2.2\luarocks.lua - LuaRocks main command-line interface

SYNOPSIS
        C:\Program Files (x86)\LuaRocks\2.2\luarocks.lua [--from=<server> | --only-from=<server>] [--to=<tree>] [VAR=VAL
UE]... <command> [<argument>]

<snip>
````

Tieske:
The Windows installer fetches this info using the pe-parser module, so it's just a matter of writing it to the config file. Can it also be reliably fetched on other os's ?

ignacio:
Ah, @Tieske is right. What I did in CMake is wrong. I can't assume that if I'm on a 64 bit system, I have to build the 64 bits version of a rock. It has to match the bitness of Lua.

hishamhm:
I was assuming the processor string would say "x86" meaning 32-bit binaries and "x86_64" meaning 64-bit binaries. Wouldn't that suffice?

ignacio:
@hishamhm I think it's not enough. In my case, my system is 64 bits but I have Lua compiled as 32 bits, so when building rocks, I have to also build them in 32 bits.
The same could happen on Linux, no?

hishamhm:
but in this case the `processor` variable says "x86" (see @Tieske's output up there), not "x86_64". For all purposes, a 32-bit Lua/LuaRocks thinks it's running on a 32-bit machine...

ignacio:
This is the output on my machine. Notice my 'processor' key:

~~~~
global environment = {
   ["processor"] = "x86_64";
   ["platform"] = {
      ["windows"] = "true";
   };
   ["show_downloads"] = "true";
   ["verbose"] = "false";
   ["os_getenv"] = "function: 00869E90, C function";
   ["home"] = "C:\\Users\\Ignacio\\AppData\\Roaming";
   ["rocks_trees"] = {
      [1] = {
         ["name"] = "user";
         ["root"] = "C:\\Users\\Ignacio\\AppData\\Roaming/luarocks";
      };
      [2] = {
         ["name"] = "system";
         ["root"] = "c:\\luarocks\\systree";
      };
   };
   ["variables"] = {
      ["LUALIB"] = "lua5.1.lib";
      ["MSVCRT"] = "MSVCRT";
   };
   ["lua_version"] = "5.1";
   ["dump_env"] = "function: 004A6440, defined in (158-162)@c:\\LuaRocks\\2.2\\lua\\luarocks\\cfg.lua";
};
~~~~

Tieske:
My result is on an AMD 64 bit cpu (running Win7 64bit, with Lua in 32bit). And the output listed;
````
C:\Users\Thijs>luarocks
global environment = {
   ["processor"] = "x86";
<snip>
````
My system config says; `AMD Turion(tm) 64 X2 Mobile Technology TL-58  1.90GHz`

Here's the uname output;
````
C:\Program Files (x86)\LuaRocks\2.2\tools>uname -s
windows32

C:\Program Files (x86)\LuaRocks\2.2\tools>uname -m
i786-pc

C:\Program Files (x86)\LuaRocks\2.2\tools>
````

So this seems unreliable. And I think @ignacio is right in that we need the state of the Lua executable, being 32/64 bit.

ignacio:
I'm toying with the idea of extracting that info in runtime, from the bytecode itself. So far, I have it working for Lua 5.1, 52 and 5.3. LuaJIT is not that straightforward, but, straight from the "ugly hacks" department:

~~~~lua
local function bitness()
	local dump = string.dump(function() end)

	local size
	
	-- check signature
	if dump:sub(2, 4) == "Lua" then
		local version = dump:sub(5,5):byte()
	
		if version == 0x51 or version == 0x52 then
			size = dump:sub(9,9):byte()
		elseif version == 0x53 then
			size = dump:sub(14,14):byte()
		else
			error("Unknown Lua version (".._VERSION..")")
		end
	elseif dump:sub(2, 3) == "LJ" then
		local ffi = require "ffi"
		size = ffi.sizeof(ffi.new("char*"))
	else
		error("Unknown Lua version")
	end
	
	return (size == 4 and 32) or
				(size == 8 and 64) or error("Unknown bitness")
end
~~~~

That was tested with Lua 5.1.5, 5.2.4, 5.3.1 and LuaJIT 2.0.4 and 2.1, in both x86 and x64 builds and produced the correct results.

Tieske:
That assumes that a binary is being compiled for the same Lua as the one that runs LuaRocks. I don't think you can make that assumption.

The c runtime is also determined at install time, so I think it should be done at install time. Then the config file is one coherent piece that determines the outcome of a binary compilation step.

ignacio:
@Tieske You're right. But in that case, at least the installer on Windows is wrong since it extracts the c runtime and the architecture from the interpreter and not from the library, which is what you'll end up linking with, right?

Tieske:
You mean from `lua.exe` instead of `lua51.dll` for example? It extracts this from Lua.exe, but if not found it will traverse its dll dependencies, and also check lua51.dll in that case until it find the runtime.

In general I would figure it is save to assume that if the interpreter is 32 or 64 bit, that then the library is also? no?

BTW; detecting processors on windows; http://blogs.msdn.com/b/david.wang/archive/2006/03/26/howto-detect-process-bitness.aspx 
If the process is 32 bit, then the returned result is usually 32 bit (`x86`) also. Because of the wow compatibility layer. I think that's the reason we should not trust `uname` on windows

ignacio:
Thanks for the link Thijs.
What I meant is to go straight to the dll to extract the bitness info. Per [your comment](https://github.com/keplerproject/luarocks/pull/385#issuecomment-113429972), we can't rely on which interpreter is used, we need to rely on the library. 

Imagine this (imaginary) scenario. I have a Lua interpreter that statically links Lua runtime and is built with Vs2008, which I only use to run LuaRocks, but I have a Lua runtime built with vs2013, because when I build rocks, I want them to link with it, because that's what I use in my app.
So I install LuaRocks using different values for /BIN, /INC and /LIB flags.

In that case, the C runtime and bitness info that matters is the one from what is configured by /LIB.

Does this make sense or I got everything wrong?

Tieske:
Exactly!

But... I take a slightly different approach. When installing LuaRocks, the installer looks for the interpreter, checks the runtime (and should also check bitness). The defaults should simply be to collect them from the interpreter found. And then write them to a config file.

Now your scenario described above, is not the default one, but should be possible. So in that case you can manually tweak the config.lua file, update the file locations, the lua lib, the bitness and the runtime to whatever target you need.

Now that LR installation will (independently of the Lua engine it runs on) compile binaries according to the settings in the config file, and hence to your own specified target.

I think that this is the way it ought to work.

Tieske:
an alternative might be to only list the lualib used in the config file by default. And at runtime dynamically determine the bitness and runtime, before any compilation starts. With the option to hardcode them in the config file.

but my personal preference would be to detect at install time, for simplicity sake. I suspect that the number of people with an installation that might benefit from the dynamic solution is hovering somewhere slightly above zero.

ignacio:
Agreed that "at install time" seems better / simpler. I thought that calling the pe-parser on the dll would be a lot simpler, but what we're detecting is the .lib, and the format is not PE. However, if we were to also look for the .dll, I think that would be even better. No config file editing afterwards.

So, if install time it is, that leaves us with the question on what to do on Linux (and bsd and stuff).

ignacio:
@Tieske I think that I've messed up :(  My system was reporting x86_64 because I edited a file by hand when dealing with #382. Terribly sorry about any confusion and waste of your time I could've made. 

I think what confused me the most is that on Windows, whenever "processor", LUAROCKS_UNAME_M, etc is mentioned, we're talking about the bitness of the detected Lua interpreter ([see here]()) and not the actual CPU arch.

So the info you @Tieske requested [here](https://github.com/keplerproject/luarocks/pull/385#issuecomment-113203489) is already there, but the name is (at least for me) misleading.

So, what do we expect to have in that key?
Also, I think that on Linux (64 bits) + Lua (32 bits) won't work.

Tieske:
I think we should close this PR by merging it. We've gone off topic quite a bit, lets continue this in #390

ignacio:
Good idea.

PR #389, opened 2015-06-18
Checks if tool is installed (svn, cvs, hg)

ignacio:
As #387, but for CVS, Mercurial and Subversion.

### DIFF
--- a/src/luarocks/fetch/cvs.lua
+++ b/src/luarocks/fetch/cvs.lua
@@ -20,9 +20,14 @@ function cvs.get_sources(rockspec, extract, dest_dir)
    assert(type(rockspec) == "table")
    assert(type(dest_dir) == "string" or not dest_dir)
 
+   local cvs_cmd = rockspec.variables.CVS
+   if not fs.execute_quiet(cvs_cmd, "--version") then
+      return nil, "'"..cvs_cmd.."' program not found. Is CVS installed? You may want to edit variables.CVS"
+   end
+
    local name_version = rockspec.name .. "-" .. rockspec.version
    local module = rockspec.source.module or dir.base_name(rockspec.source.url)
-   local command = {rockspec.variables.CVS, "-d"..rockspec.source.pathname, "export", module}
+   local command = {cvs_cmd, "-d"..rockspec.source.pathname, "export", module}
    if rockspec.source.tag then
       table.insert(command, 4, "-r")
       table.insert(command, 5, rockspec.source.tag)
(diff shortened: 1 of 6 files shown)


src/luarocks/fetch/cvs.lua
```diff
@@ -20,9 +20,14 @@ function cvs.get_sources(rockspec, extract, dest_dir)
    assert(type(rockspec) == "table")
    assert(type(dest_dir) == "string" or not dest_dir)
 
+   local cvs_cmd = rockspec.variables.CVS
+   if not fs.execute_quiet(cvs_cmd, "--version") then
+      return nil, "'"..cvs_cmd.."' program not found. Is CVS installed? You may want to edit variables.CVS"
```
Tieske: wouldn't it be clearer to write; "You may want to edit the 'variables.CVS' value in file 'config.lua' "

hishamhm: alternatively, `"'"..cvs_cmd.."' program not found. Please make sure it is available in your PATH".` _might_ be an easier instruction to follow (not sure).

siffiejoe: It doesn't have to be in PATH if it's an absolute path.

ignacio: Sure. I copied this message from the cmake builder. Considering that:
1) The user just wants to install a rock. She might have no idea what "hg" is.
2) The tool's executable may not indicate the tool's name (hg -> Mercurial)

What about:
'foo' program not found. Make sure Bar is installed and is available in your PATH"

src/luarocks/fetch/hg.lua
```diff
@@ -21,9 +21,13 @@ function hg.get_sources(rockspec, extract, dest_dir)
...
 
    local hg_cmd = rockspec.variables.HG
+   if not fs.execute_quiet(hg_cmd, "--version") then
+      return nil, "'"..hg_cmd.."' program not found. Is Mercurial installed? You may want to edit variables.HG"
+   end
+
    local name_version = rockspec.name .. "-" .. rockspec.version
    -- Strip off special hg:// protocol type
-   local url = rockspec.source.url:gsub("^hg://", "") 
+   local url = rockspec.source.url:gsub("^hg://", "")
```
ignacio: Sorry @hishamhm that trailing whitespace trigger my OCD :smile:

hishamhm: np :)

src/luarocks/fetch/cvs.lua
hishamhm: > It doesn't have to be in PATH if it's an absolute path.

Sure, I was just catering for the common case. If git is not found, most likely it is simply not installed and installing via the OS's usual methods will put it there.

Tieske: not on Windows...

ignacio:
So, do you people think my proposed wording is ok?

hishamhm:
The refactored function is a good idea and I think the wording is good.

PR #393, opened 2015-06-24
add target cpu info to configuration

Tieske:
- site_config remains local, no longer exposed
- added field 'target_cpu' to `cfg` and config file environments

as discussed in https://github.com/keplerproject/luarocks/issues/390#issuecomment-114852615

### DIFF
--- a/src/luarocks/cfg.lua
+++ b/src/luarocks/cfg.lua
@@ -32,8 +32,6 @@ if not ok then
    site_config = {}
 end
 
-cfg.site_config = site_config
-
 cfg.program_version = "scm"
 cfg.program_series = "2.2"
 cfg.major_version = (cfg.program_version:match("([^.]%.[^.])")) or cfg.program_series
@@ -87,6 +85,7 @@ elseif proc:match("amd64") or proc:match("x86_64") then
 elseif proc:match("Power Macintosh") then
    proc = "powerpc"
 end
+cfg.target_cpu = proc
 
 if system == "FreeBSD" then
    detected.unix = true
@@ -151,7 +150,8 @@ env_for_config_file = {
    home = cfg.home,
    lua_version = cfg.lua_version,
    platform = util.make_shallow_copy(detected),
-   processor = proc,
+   processor = cfg.target_cpu,   -- remains for compat reasons
+   target_cpu = cfg.target_cpu,  -- replaces `processor`
    os_getenv = os.getenv, 
    dump_env = function()
      -- debug function, calling it from a config file will show all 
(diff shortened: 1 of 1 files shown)


src/luarocks/cfg.lua
```diff
@@ -630,4 +630,13 @@ function cfg.is_platform(query)
...
 
+--[[
+function cfg.dump()
+  for k,v in pairs(defaults) do
+    if cfg[k] then end -- nothing todo, just touch it and the __index will make a local copy in `cfg` 
+  end
+  print(util.show_table(cfg, "Dump of `cfg` table:"))
+end
+cfg.dump() --]]
+
```
hishamhm: not a fan of leaving debug stuff around as comments like that...

Tieske:
oops... fixed!

ignacio:
I had a go at testing this but I think something is wrong. luarocks.cfg seems to be overriding my config file with wrong values.

This is what the installer found:
~~~~
==========================
== System check results ==
==========================

Will configure LuaRocks with the following paths:
LuaRocks        : c:\luarocks\2.2
Config file     : c:\luarocks\config.lua
Rocktree        : c:\luarocks\systree

Lua interpreter : c:\lua\X64\5.1.5\bin\\lua.exe
    binaries    : c:\lua\X64\5.1.5\bin\
    libraries   : c:\lua\X64\5.1.5\lib\
    includes    : c:\lua\X64\5.1.5\include\
    architecture: x86_64
    binary link : lua51.lib with runtime MSVCR120.dll
~~~~

This is part of my config file:
~~~~
variables = {
    MSVCRT = 'MSVCR120',
    LUALIB = 'lua51.lib'
}
~~~~

And it seems that https://github.com/keplerproject/luarocks/blob/master/src/luarocks/cfg.lua#L342 is overriding it with the wrong value.

ignacio:
Ok... nevermind, I saw that "remains; configuring targets from config.lua". Until that is in place, not even the builtin build backend works on Windows. (MSVCRT variable is nil)

Tieske:
No... you're right. But the problem was introduced in #385. The config file was directly loaded into `cfg`, which exposed `cfg` to the config file. After changing that, the content of the config file was no longer merged into `cfg`, because that should now be an extra step.

Fixed it, PR is upcoming

Tieske:
see PR #397 

please give this another try, it should also allow updating targets if you add `target_cpu = x86_64` in `config.lua` for example.

ignacio:
Thanks. Will check that first thing tomorrow morning.

ignacio:
@Tieske Just checked and it is working fine. Also overriding target_cpu on the config file. The PR is good to merge.

Thanks!

PR #395, opened 2015-06-24
Addon prototype

xiaq:


ignacio:
Wow :) Is there a document I could read describing the addon system?

hishamhm:
@ignacio we're not even halfway through the GSoC project yet (and the design is still in flux! (and impressions/feedback are certainly welcome!) ), but documentation is part of the plan, don't worry :)

src/luarocks/type_check.lua
```diff
@@ -170,6 +183,43 @@ local manifest_types = {
...
+      return nil, "Field "..context..field.." already exists"
+   end
+   tbl[field] = value
+end
+
+--- Add a new field to rockspec_types. Must not conflict with existing field.
+-- @param field string: see add_field.
+-- @return boolean or (nil, string), see add_field.
+function type_check.add_rockspec_field(field, typetbl)
+   add_field(rockspec_types, field, typetbl, "")
```
hishamhm: missing return to propagate the output of add_field

ignacio:
I was looking for a "rough design" or something, to avoid talking nonsense :D (more than the usual amount).

xiaq:
@ignacio Hi! I didn't expect feedback this soon :)

I have yet to write a formal doc, but here is a quick view of how it works now:

Addon `$a` is the lua module `luarocks.addon.$a` exporting two functions `load` and `run`.

1. The `run` function is called when the user calls `luarocks $a`, the eponymous command. It is passed the same arguments with the `run` functions of builtin commands.

2. The `load` function is called when the user has `addons = { ..., $a, ... }` in the rockspec. It can then do various things, including introducing new rockspec fields and hooks that are triggered at various stages of LuaRocks execution. These are done through calls to functions in the new `luarocks.api` module.

As examples, see the following sample addons:

1. [luarocks-addon-test](https://github.com/xiaq/luarocks-addon-test) introduces a new field called `test` that has the same format as `build.modules`. For each source `blah.lua` it tries to run `blah_test.lua` after building the rock.

2. [luarocks-addon-moonc](https://github.com/xiaq/luarocks-addon-moonc) introduces a new field called `moonc` that has the same format as `build.modules`. For each source `blah.lua` that does not exist but where `blah.moon` exists, it tries to compile the latter before building the rock.

There is also a sample rock [dinner](https://github.com/xiaq/dinner) that uses the above two addons.

ignacio:
@xiaq thanks! That'll do.

src/luarocks/fetch.lua
```diff
@@ -244,21 +271,46 @@ function fetch.load_local_rockspec(filename, quick)
...
-      for i = 1, #rockspec.dependencies do
-         local parsed, err = deps.parse_dep(rockspec.dependencies[i])
-         if not parsed then
-            return nil, "Parse error processing dependency '"..rockspec.dependencies[i].."': "..tostring(err)
-         end
-         rockspec.dependencies[i] = parsed
-      end
-   else
-      rockspec.dependencies = {}
+   local ok, err = parse_deps(rockspec, "build_dependencies")
```
hishamhm: I understand you're passing the table name as a string so you can do rockspec[field] = {} in line 188 above, but I'd just initialize both tables here instead and simplify the function above.
```lua
rockspec.dependencies = rockspec.dependencies or {}
local ok, err = parse_deps(rockspec.dependencies)
-- etc.
```
Minor thing, but it makes it clearer when reading that `parse_deps` isn't modifying the rest of the rockspec.

src/luarocks/fetch.lua
```diff
@@ -244,21 +271,46 @@ function fetch.load_local_rockspec(filename, quick)
...
+      return nil, err
+   end
+
+   -- Load addons
+   addon.reset()
+   if rockspec.addons then
+      for _, name in ipairs(rockspec.addons) do
+         local ok, err = addon.load(name)
+         if not ok then
+            return nil, filename..": cannot load addon "..addon..": "..err
```
hishamhm: should be `return nil, filename..": cannot load addon "..name..": "..err` — got a crash there

xiaq:
@hishamhm Fixes pushed.

src/luarocks/addon.lua
```diff
@@ -0,0 +1,94 @@
...
+   if i then
+      return get(tbl[field:sub(1,i-1)], field:sub(i+1))
+   end
+   return tbl[field]
+end
+
+function addon.handle_rockspec(rockspec)
+   for k, v in pairs(rockspec_field_registry) do
+      if v.callback then
+         field_value = get(rockspec, k)
```
mpeterv: `field_value` should be local?

PR #396, opened 2015-06-24
Add support for git submodules

mpeterv:
Version of #345 for luarocks-3 branch, also rebased.

### DIFF
--- a/src/luarocks/fetch/git.lua
+++ b/src/luarocks/fetch/git.lua
@@ -110,15 +110,18 @@ function git.get_sources(rockspec, extract, dest_dir, depth)
       end
    end
 
-   command = {fs.Q(git_cmd), "submodule", "update", "--init", "--recursive"}
+   -- Fetching git submodules is supported only when rockspec format is >= 3.0.
+   if deps.format_is_at_least(rockspec, "3.0") then
+      command = {fs.Q(git_cmd), "submodule", "update", "--init", "--recursive"}
 
-   if git_supports_shallow_submodules(git_cmd) then
-      -- Fetch only the last commit of each submodule.
-      table.insert(command, 5, "--depth=1")
-   end
+      if git_supports_shallow_submodules(git_cmd) then
+         -- Fetch only the last commit of each submodule.
+         table.insert(command, 5, "--depth=1")
+      end
 
-   if not fs.execute(unpack(command)) then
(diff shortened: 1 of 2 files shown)


hishamhm:
Please check that `rockspec.rockspec_format` >= "3.0" when fetching submodules (use `deps.compare_versions`).

mpeterv:
Looks like `rockspec_format` can be nil when using `write_rockspec`, as it does not use fetch functions for loading rockspecs. I'll

ignacio:
Should we close #345 in favor of this one then?

src/luarocks/fetch/git.lua
```diff
@@ -77,12 +105,25 @@ function git.get_sources(rockspec, extract, dest_dir, depth)
...
    if tag_or_branch and not git_can_clone_by_tag() then
-      local checkout_command = {fs.Q(git_cmd), "checkout", tag_or_branch}
-      if not fs.execute(unpack(checkout_command)) then
+      if not fs.execute(fs.Q(git_cmd), "checkout", tag_or_branch) then
          return nil, 'Failed to check out the "' .. tag_or_branch ..'" tag or branch.'
       end
    end
 
+   -- Fetching git submodules is supported only when rockspec format is >= 3.0.
+   if not rockspec.rockspec_format or not deps.compare_versions("3.0", rockspec.rockspec_format) then
```
hishamhm: Do not assume that a missing `rockspec_format` means >= 3.0 (it means 1.0). If `write_rockspec` is not setting it, I guess we should change that instead. I'd prefer to have write_rockspec produce format-1.0 compatible rockspecs by default (even after 3.0 is released, it can be a long time before it shows up in distros), so adding a `--rockspec-format <version>` flag to `write_rockspec` would be the best approach.

Also, if we start putting checks like this in the code, it might be a good idea to have a function `deps.is_format(rockspec, "3.0")` to make the >= check more readable. :)

mpeterv:
Repushed with a fix. Also I think it would make sense to bump `type_check.rockspec_format` to 3.0 on luarocks-3 branch, so luarocks could accept rockspecs using new features.

hishamhm:
> Also I think it would make sense to bump type_check.rockspec_format to 3.0 on luarocks-3 branch, so that luarocks could accept rockspecs using new features.

Good call. Done!

PR #398, opened 2015-06-25
Refactor cfg config file loading

Tieske:
builds on top of #397. After dealing with that PR I wasn't happy with the config file handling code, which was repetitive and hard to grasp.

Adds more structure to loading config files, removing duplicate code and errorhandling

### DIFF
--- a/src/luarocks/cfg.lua
+++ b/src/luarocks/cfg.lua
@@ -35,6 +35,9 @@ end
 cfg.program_version = "scm"
 cfg.program_series = "2.2"
 cfg.major_version = (cfg.program_version:match("([^.]%.[^.])")) or cfg.program_series
+cfg.variables = {}
+cfg.rocks_trees = {}
+cfg.platform = {}
 
 local persist = require("luarocks.persist")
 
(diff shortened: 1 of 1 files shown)


hishamhm:
This is not just a refactor, it changes behavior too, right? With the new code, a user can redefine the platform list and thus change which defaults are picked. If nothing's broken by these changes, I welcome the new behavior.

However, I notice that now there's `cfg.platform` and `cfg.platforms` (set via `defaults`), and they seem redundant. `platforms` is the one used elsewhere by LuaRocks for the platform-overrides in rockspecs. Could you check this? Can we just change `cfg.platform` to `cfg.platforms` directly?

Tieske:
> This is not just a refactor, it changes behavior too, right? With the new code, a user can redefine the platform list and thus change which defaults are picked. If nothing's broken by these changes, I welcome the new behavior.

Not intentional, but I think you're right. 

> However, I notice that now there's  `cfg.platform`  and  `cfg.platforms`  (set via  defaults ), and they seem redundant.  `platforms`  is the one used elsewhere by LuaRocks for the platform-overrides in rockspecs. Could you check this? Can we just change  `cfg.platform`  to  `cfg.platforms`  directly?

No I don't think so. If I dump that table I get;
````
   ["platform"] = {
      ["windows"] = "true";
      ["mingw32"] = "true";
   };
   ["major_version"] = "2.2";
   ["makefile"] = "Makefile.win";
   ["lua_interpreter"] = "lua.exe";
   ["platforms"] = {
      [1] = "win32";
      [2] = "mingw32";
      [3] = "windows";
   };
````
So the tables don't match. I'll try and track their uses and see whether it can be combined.

hishamhm:
Oh, you're right, one is a set, the other an array. One should be generated from the other, but you'll want users to write in the config files in array form. Will take a bit of care to get it right. Look out, also, that in `cfg.platforms` the order is relevant (it affects the semantics of platform overrides).

Tieske:
my output is now;
````
   ["platforms"] = {
      [1] = "windows";
      [2] = "mingw32";
      ["win32"] = "true";
      ["mingw32"] = "true";
      [3] = "win32";
      ["windows"] = "true";
   };
````
users can write in their config files `platforms.windows = true` now, so not in array form.
But because the array part is generated from the keyed part, the order (by use of `pairs`) of the array part cannot be guaranteed.

It works for me, but where in  the code is the order relevant for the platform overrides, as you mentioned?

hishamhm:
> It works for me, but where in the code is the order relevant for the platform overrides, as you mentioned?

Platform overrides go through the list via ipairs. The order is relevant because you can have a `platforms` section in a rockspec with general "unix" stuff and then you override defines using more specific entries such as "linux" and "macosx".

I just checked and this is not a theoretical concern: if the order is not deterministic, the LuaSocket rockspec will break on Windows (it has both "mingw32" and "win32" sections which redefine variables).

Tieske:
Just added a fix for that. Please give it a try.

hishamhm:
Giving this a closer look, I realize that "windows" is higher up in priority than "win32" and "mingw32", a design mistake on my part. Technically, fixing that ordering would require handling that different orderings based on the value of rockspec_format when reading a rockspec. Still, on who-knows-how-many-years, nobody has complained about that (not many rockspecs with Windows-specifc code, and they probably just use "win32" instead), so it might be something we could fix on the bump to 3.0, make a note about it in the release notes and fix rockspecs if some... or just do nothing.

Tieske:
For now leave it. I think we should in a future update fix the "win32" references all around and change them to "windows"

Tieske:
any concerns left? or is this good to merge?

ignacio:
Maybe merge this and open a new issue in branch luarocks 3 to not loose track of the win32 / windows / mingw issue?

hishamhm:
@Tieske, yes, it's good to merge.
@ignacio, good idea.

PR #352, opened 2015-06-25
luarocks list does not respect the --local flag

ignacio:
~~~~
d:\Desarrollo\luarocks>luarocks list oauth --local

Installed rocks:
----------------

oauth
   0.0.6-1 (installed) - c:/luarocks/systree/lib/luarocks/rocks
~~~~

I couldn't find where the --local flag is dealt with. It is mapped to --tree=/path/to/home/tree somewhere?


ignacio:
I just saw that I never sent a PR for this. @hishamhm could you take a look at it when you can?

hishamhm:
I will commit an alternative

ignacio:
No issues. It's working.

PR #399, opened 2015-06-26
write_rockspec: allow selecting rockspec format

mpeterv:
Add `--rockspec-format` option for `write_rockspec` command, to be used as `--rockspec-format=1.1`.


hishamhm:
Thank you!

PR #407, opened 2015-07-02
Prefer versioned config files

Tieske:
As discussed here; https://github.com/keplerproject/luarocks/issues/403#issuecomment-117988369

### DIFF
--- a/install.bat
+++ b/install.bat
@@ -794,10 +794,11 @@ print()
 print("Configuring LuaRocks...")
 
 -- Create a site-config file
-if exists(S[[$LUADIR\luarocks\site_config.lua]]) then
-	exec(S[[RENAME "$LUADIR\luarocks\site_config.lua" site_config.lua.bak]])
+local site_config = S("site_config_$LUA_VERSION"):gsub("%.","_")
+if exists(S([[$LUADIR\luarocks\]]..site_config..[[.lua]])) then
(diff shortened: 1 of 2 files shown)


install.bat
```diff
@@ -820,24 +822,24 @@ site_config.LUAROCKS_MD5CHECKER=[[md5sum]]
...
 
 -- create config file
-vars.CONFIG_FILE = vars.SYSCONFDIR.."\\config.lua"
+vars.CONFIG_FILE = vars.SYSCONFDIR.."\\"..vars.SYSCONFFILENAME
 if not exists(vars.SYSCONFDIR) then
 	mkdir(vars.SYSCONFDIR)
 end
 if exists(vars.CONFIG_FILE) then
-	local nname = backup(vars.CONFIG_FILE, "config.bak")
+	local nname = backup(vars.CONFIG_FILE, S"config-$LUA_VERSION.bak")
```
ignacio: Wouldn't it be better to just use
```lua
local nname = backup(vars.CONFIG_FILE, vars.SYSCONFFILENAME..".bak")
````
so you don't have to keep both names format in sync by hand?

Tieske: I think it could. I didn't do it because current behaviour doesn't include the `.lua` extension. If you change it the backup file would be `config-5.1.lua.bak` because `vars.CONFIG_FILE` already includes the `.lua` extension.

PR #413, opened 2015-07-18
Win installer: Unversion LR target path

### LINKED ISSUE #151: upgrade breaks installed batch files
The batchfiles LR generates (for example when installing LDoc), contains PATH references to the LR installation so it can load the `luarocks.loader` module.
Because this path contains the LR version; `2.0`, the batch file breaks after installling `2.1` (and removing `2.0`)

````
C:\Temp>ldoc
c:/users/public/lua/5.1/bin/lua.exe: module 'luarocks.loader' not found:
        no field package.preload['luarocks.loader']
        no file 'c:\users\public\lua\5.1/share/lua/5.1/luarocks\loader.lua'
        no file 'c:\users\public\lua\5.1/share/lua/5.1/luarocks\loader/init.lua'
        no file 'c:\users\public\lua\LuaRocks\2.0\lua\luarocks\loader.lua'
        no file 'c:\users\public\lua\LuaRocks\2.0\lua\luarocks\loader\init.lua'
        no file '.\luarocks\loader.lua'
<SNIP>
        no file 'c:\users\public\lua\5.1\lib\lua\5.1\luarocks.dll'
stack traceback:
        [C]: ?
        [C]: ?

C:\Temp>
````

Currently this requires all rocks with generated batch files to be reinstalled using the new LR version.

Does this problem also exist on other platforms?

Could this be circumvented by installing the loader as a rock maybe? (guess not as it has dependencies on a lot of other LR stuff)

Maybe not set the path in the rock specific batch file, but create a generic batch file setting the LR loader path. This one can than be called from the rockspecific batch files. Upon installing LR, the batch file gets updated and all existing batch files then automatically follow the new path.

please comment.

### PR DESCRIPTION
**Pull request by Tieske**
This PR fixes the installer to use a non-versioned path for the LR installation (fixes #151). As discussed here https://github.com/keplerproject/luarocks/issues/403#issuecomment-118026700 . It also fixes the retainment of the existing config files.

One problem though; because the LR path is moving one level up, the system rock tree is now in the same path. e.g. default used to be `c:\program files (x86)\luarocks\2.2` and now is `c:\program files (x86)\luarocks`. This means that when deleting that existing directory upon installation, the default location of the system tree; `c:\program files (x86)\luarocks\systree`, is now also deleted.

iirc correctly the system tree used to be installed in the Lua directories, not the LuaRocks directories. Some 2 years ago this was changed, but with hind sight that seems a bad change, so maybe that should be reverted.

For now; this is for discussion only, do not merge yet, until the systree location issue is resolved.

### DIFF
--- a/.appveyor/build.bat
+++ b/.appveyor/build.bat
@@ -18,9 +18,8 @@ if not defined LUAROCKS_VER set LUAROCKS_VER=2.2.1
 set LUAROCKS_SHORTV=%LUAROCKS_VER:~0,3%
 
 if not defined LR_EXTERNAL set LR_EXTERNAL=c:\external
-if not defined LUAROCKS_INSTALL set LUAROCKS_INSTALL=%LUA_DIR%\LuaRocks
-if not defined LR_ROOT set LR_ROOT=%LUAROCKS_INSTALL%\%LUAROCKS_SHORTV%
-if not defined LR_SYSTREE set LR_SYSTREE=%LUAROCKS_INSTALL%\systree
+if not defined LR_ROOT set LR_ROOT=%LUA_DIR%\LuaRocks
+if not defined LR_SYSTREE set LR_SYSTREE=%LR_ROOT%\systree
 
 ::
 :: =========================================================
@@ -35,7 +34,7 @@ if not exist %LUA_DIR%\bin\%LUA%.exe call :die "Missing Lua interpreter at %LUA_
 :: =========================================================
 
 cd %APPVEYOR_BUILD_FOLDER%
-call install.bat /LUA %LUA_DIR% /Q /LV %LUA_SHORTV% /P "%LUAROCKS_INSTALL%" /TREE "%LR_SYSTREE%"
+call install.bat /LUA %LUA_DIR% /Q /LV %LUA_SHORTV% /P "%LR_ROOT%" /TREE "%LR_SYSTREE%"
 
 if not exist "%LR_ROOT%" call :die "LuaRocks not found at %LR_ROOT%"
 
--- a/.appveyor/build.bat
+++ b/.appveyor/build.bat
@@ -47,6 +47,7 @@ set LUA_PATH=%LUA_PATH%;%LR_SYSTREE%\share\lua\%LUA_SHORTV%\?\init.lua
 set LUA_CPATH=%LR_SYSTREE%\lib\lua\%LUA_SHORTV%\?.dll
 
 call luarocks --version || call :die "Error with LuaRocks installation"
+cd && dir
 call luarocks list
 
 
--- a/.appveyor/build.bat
+++ b/.appveyor/build.bat
@@ -47,7 +47,6 @@ set LUA_PATH=%LUA_PATH%;%LR_SYSTREE%\share\lua\%LUA_SHORTV%\?\init.lua
 set LUA_CPATH=%LR_SYSTREE%\lib\lua\%LUA_SHORTV%\?.dll
 
 call luarocks --version || call :die "Error with LuaRocks installation"
-cd && dir
 call luarocks list
 
 
--- a/install.bat
+++ b/install.bat
@@ -487,18 +487,17 @@ local function backup_config_files()
     temppath = os.getenv("temp").."\\LR-config-backup-"..tostring(math.random(10000))
     if exists(temppath) then temppath = nil end
   end
-  print("'"..temppath.."'")
   vars.CONFBACKUPDIR = temppath
   mkdir(vars.CONFBACKUPDIR)
   exec(S[[COPY "$PREFIX\config*.*" "$CONFBACKUPDIR" >NUL]])
-  exec(S[[COPY "$FULL_PREFIX\lua\luarocks\site_config*.*" "$CONFBACKUPDIR" >NUL]])
+  exec(S[[COPY "$PREFIX\lua\luarocks\site_config*.*" "$CONFBACKUPDIR" >NUL]])
 end
 
 -- restore previously backed up config files
 local function restore_config_files()
   if not vars.CONFBACKUPDIR then return end -- there is no backup to restore
   exec(S[[COPY "$CONFBACKUPDIR\config*.*" "$PREFIX" >NUL]])
-  exec(S[[COPY "$CONFBACKUPDIR\site_config*.*" "$FULL_PREFIX\lua\luarocks" >NUL]])
+  exec(S[[COPY "$CONFBACKUPDIR\site_config*.*" "$PREFIX\lua\luarocks" >NUL]])
   -- cleanup
   exec(S[[RD /S /Q "$CONFBACKUPDIR"]])
   vars.CONFBACKUPDIR = nil
@@ -610,11 +609,10 @@ else
 end
 
 vars.PREFIX = vars.PREFIX or os.getenv("PROGRAMFILES")..[[\LuaRocks]]
-vars.FULL_PREFIX = S"$PREFIX\\$VERSION"
-vars.BINDIR = vars.FULL_PREFIX
-vars.LIBDIR = vars.FULL_PREFIX
-vars.LUADIR = S"$FULL_PREFIX\\lua"
-vars.INCDIR = S"$FULL_PREFIX\\include"
+vars.BINDIR = vars.PREFIX
+vars.LIBDIR = vars.PREFIX
+vars.LUADIR = S"$PREFIX\\lua"
+vars.INCDIR = S"$PREFIX\\include"
 vars.LUA_SHORTV = vars.LUA_VERSION:gsub("%.", "")
 
 if INSTALL_LUA then
@@ -659,7 +657,7 @@ print(S[[
 ==========================
 
 Will configure LuaRocks with the following paths:
-LuaRocks        : $FULL_PREFIX
+LuaRocks        : $PREFIX
 Config file     : $CONFIG_FILE
 Rocktree        : $TREE_ROOT
 
@@ -689,18 +687,18 @@ print([[
 -- Install LuaRocks files
 -- ***********************************************************
 
-if exists(vars.FULL_PREFIX) then
+if exists(vars.PREFIX) then
   if not FORCE then
-    die(S"$FULL_PREFIX exists. Use /F to force removal and reinstallation.")
+    die(S"$PREFIX exists. Use /F to force removal and reinstallation.")
   else
     backup_config_files()
-    print(S"Removing $FULL_PREFIX...")
-    exec(S[[RD /S /Q "$FULL_PREFIX"]])
+    print(S"Removing $PREFIX...")
+    exec(S[[RD /S /Q "$PREFIX"]])
     print()
   end
 end
 
-print(S"Installing LuaRocks in $FULL_PREFIX...")
+print(S"Installing LuaRocks in $PREFIX...")
 if not exists(vars.BINDIR) then
 	if not mkdir(vars.BINDIR) then
 		die()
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by Tieske** [comment]
@ignacio any idea why LR refuses to run, complaining that the current directory doesn't exist? The added debug line "cd && dir" gives proper output...

I had a quick look at the source in `fs\win32\tools.lua`, but I don't understand this 'directory stack' concept that seems to be used? what is it for, and how does it work?

**Comment by ignacio** [comment]
Hi @Tieske . Will take a look at this later today and get back to you.

**Comment by Tieske** [comment]
It's already fixed. I thought it to be an appveyor issue, but it turned out to be an issue with `cfg.Lua`.

So back to the original issue; location of the system tree... Any ideas?

**Comment by hishamhm** [comment]
Perhaps bring the issue to the LR mailing list to see if any other Windows users chime in? I trust you guys to make the final call, but it would be nice to at least hear what users have to say on this matter.

**Comment by ignacio** [comment]
I think of two options.
1) Even when the versioned directory is removed (e.g. 2.2), still leave a fixed one. So instead of `c:\program files (x86)\luarocks\2.2` we get `c:\program files (x86)\luarocks\install` or something like that.

2) Move the system tree to a different place. [Local AppData](http://windows.microsoft.com/en-us/windows-8/what-appdata-folder) seems like a reasonable place. I'm pretty sure this has been discussed already. I can't recall the pros and cons of each location.

**Comment by siffiejoe** [comment]
If I understand correctly, the current issue is that a `rm -rf $LUAROCKS_PREFIX` also removes the system rocks tree, because it is now inside of `$LUAROCKS_PREFIX`. If so, another obvious option would be to *not* do the `rm -rf $LUAROCKS_PREFIX`. Either remove the necessary files/directories explicitly, or leave them be. If the installer is used to install additional versions of LuaRocks in the same directory, you shouldn't remove the old installation anyway.

**Comment by Tieske** [comment]
> Move the system tree to a different place. Local AppData seems like a reasonable place. I'm pretty sure this has been discussed already. I can't recall the pros and cons of each location.

iirc it was for security reasons, LR data is not real "data" but application code, and as such needs admin priviledges to write to, hence the Program Files location, as AppData is not protected in that way.

> If I understand correctly

@siffiejoe  you did

> If so, another obvious option would be to not do the rm -rf $LUAROCKS_PREFIX. Either remove the necessary files/directories explicitly

Not the cleanest solution imo, also very vulnerable to upgrade mistakes if files are moved/added/renamed. Due to the versioned directory they were separated before, my preference would be to keep them separated

>  If the installer is used to install additional versions of LuaRocks in the same directory, you shouldn't remove the old installation anyway.

I don't see how this could work? The other discussion was regarding one LR installation serving rocks for multiple Lua installations, maybe that is what you meant? This pull request takes care of retaining the `config[-x.x].lua` and `site_config[_x_x.lua]` files in that case (copied to a temp location before removing the full directory, and copied back after the main files have been installed)

**Comment by hishamhm** [comment]
> 1) Even when the versioned directory is removed (e.g. 2.2), still leave a fixed one. So instead of c:\program files (x86)\luarocks\2.2 we get c:\program files (x86)\luarocks\install or something like that.

I thought of that option too.

Alternatively, how about this? 

* "c:\Program Files (x86)\LuaRocks.org\LuaRocks"
* "c:\Program Files (x86)\LuaRocks.org\Installed Rocks"

Looks very Windows-y to me ;)

**Comment by Tieske** [comment]
I like the LuaRocks.org namespacing. :+1: 

Alternatively I was considering using the Lua tree;
````
{root}
  +-- bin
  +-- include
  +-- lib
  |    +-- lua
  |         +-- <LuaVersion>
  +-- man
  |    +-- man1
  +-- share
       +-- lua
            +-- <LuaVersion>
````
This is the default structure of the makefile for MinGW. And Lua 5.3 includes the defaults paths for these as well (LuaWinMake also creates this structure when installing).
So if one would use the `{root}` above as the systemtree, then no Lua path environment variables need to be set at all. Everything would work out of the box, on defaults.

as a bonus it would be versioned (except for `bin`).

**Comment by Tieske** [comment]
we could even place a small `luarocks.bat` wrapper batch file in the `bin` directory, so only adding the path to `lua.exe` to the system path would make it all work, including the LuaRocks commandline.

**Comment by hishamhm** [comment]
> Alternatively I was considering using the Lua tree;

Yeah, that would be the Unix-y solution. I think that's what I wanted to go with in the earliest days (due to sheer laziness/simplicity of having everything look/work the same everywhere) but for some reason we didn't go with that (Kepler already had different conventions, maybe? Or maybe Windows people were offended by the Unix-like paths, I don't remember :) )

**Comment by Tieske** [comment]
I think things have changed, the tree structure has been standardized a little bit more by Lua 5.3 including the defaults paths for this structure. And the earlier mentioned security, previously everyone on Windows had admin rights, and that has also changed (for the better).

If LR adopts this structure, I think the 'standard' would become a little stronger again...

**Comment by hishamhm** [comment]
Well, you certainly have my support for the unified bin+lib+share structure in all OSes :)

**Comment by ignacio** [comment]
So, what would `{root}` be? I didn't get that part. Could you provide some examples with full paths?

**Comment by Tieske** [comment]
`{root}` would be the root of the Lua installation. On my system that would be `C:\Program Files (x86)\Lua`.
So the full tree of my Lua installation is;
````
C:\Program Files (x86)
    +-- Lua   
        +-- bin
        +-- include
        +-- lib
        |    +-- lua
        |         +-- 5.1
        +-- man
        |    +-- man1
        +-- share
             +-- lua
                  +-- 5.1
````
At the same time my system tree is;
````
C:\Program Files (x86)
    +-- LuaRocks
        +-- systree
            +-- bin
            +-- lib
            |    +-- lua
            |    |    +-- 5.1
            |    +-- luarocks
            |         +-- rocks
            +-- share
                 +-- lua
                      +-- 5.1
````

So if I would point my system tree to `C:\Program Files (x86)\Lua` they would exactly match on top of each other. And in case of Lua 5.3 I wouldn't even have to change the Lua environment variables to find the installed rocks, as they are already in the defaults.

**Comment by ignacio** [comment]
So you'd be install the system tree in `C:\Program Files (x86)\Lua`. Great.
And LuaRocks own files would live in `c:\Program Files (x86)\LuaRocks` and not in `C:\Program Files (x86)\Lua`, right?

**Comment by Tieske** [comment]
Yes, just the system rocktree would be integrated into the Lua directory tree, LuaRocks would still have its own, non-versioned, installation directory.

**Comment by ignacio** [comment]
Great. I'm good with this.

**Comment by Tieske** [comment]
Just updated this PR with the change as discussed. Please give this a try and let me know whether it is good to go.

**Comment by hishamhm** [comment]
Sorry about chiming in late, but... changing the tree to a Unix-like structure is good, but I'm unsure about our installer installing Lua at `C:\Program Files (x86)\Lua`. Looking at this path, it looks like something straight from lua.org. Shouldn't we either use something namespaced like `C:\Program Files (x86)\LuaRocks.org\LuaRocks` and `C:\Program Files (x86)\LuaRocks.org\Lua`, to avoid any possible clashes with other Lua installations? (I understand that the latter is also misleading as it looks like Lua comes from LuaRocks.org, but at least that's true in the sense that this is the _build_ we're shipping).

OR...

While we're at it... why two trees? If the LR system tree will store stuff in Lua's tree anyway, why not install LuaRocks itself there too? In other words...

```
c:\Program Files (x86)\
   LuaRocks
      bin
         lua-5.1
         luarocks-5.1
         luarocks
      lib
         lua
            5.1
      share
         lua
            5.1
```

Removing a LuaRocks version when upgrading equates to deleting `bin\luarocks*.*` and `share\lua\5.1\luarocks`, assuming the configuration stays the same... it shouldn't be messy, right?

hishamhm:
Also, this will be a major usability change for users... maybe we should make this move in the `luarocks-3` branch?

Tieske:
> Sorry about chiming in late, but... changing the tree to a Unix-like structure is good, but I'm unsure about our installer installing Lua at `C:\Program Files (x86)\Lua`. Looking at this path, it looks like something straight from lua.org.

The bundled Lua version is installed as `C:\Program Files (x86)\LuaRocks\lua5.1.exe`, so inside the LR directory. That example was just my own Lua installation, so it wasn't installed by the LR installer. Sorry for the confusion.

> While we're at it... why two trees? If the LR system tree will store stuff in Lua's tree anyway, why not install LuaRocks itself there too?

I'm not sure about this one. I see your point, but I'm not comfortable with it (yet). Have to give it some thought. Maybe take it one step at a time. And make that a next step.
Maybe @ignacio or @siffiejoe have some thoughts on this?

> Also, this will be a major usability change for users... maybe we should make this move in the luarocks-3 branch?

No, I don't think so. Last time we moved the tree, it also was a point release. A user just needs the /TREE argument to move the tree elsewhere, or he doesn't move it and just reinstalls the modules he wants/needs. Imho adding a note to the release should suffice.

siffiejoe:
> Maybe @ignacio or @siffiejoe have some thoughts on this?

To clarify: We are talking about four directory trees. Two are also rocks trees (one requires admin rights, the other doesn't). Then there is the Lua installation directory, and the LuaRocks installation directory.

The current plan is to put one of the rocks trees into the Lua installation directory, because then Lua 5.3, when installed using a certain directory layout, doesn't need any `LUA_PATH*` variables set to access the modules. Correct? I'm a bit uncomfortable installing LuaRocks into a directory owned by another program, but ok. But then the LuaRocks batch files should be installed side-by-side with `lua.exe` so that you also don't have to modify the PATH variable (if you have already done so for Lua). (The various tools shouldn't be in PATH, so I'd put them in `.../lib/luarocks/tools` or something.)

Another idea was to move the LuaRocks modules from the LuaRocks installation directory into the system rocks tree. This is already done on Unix, and it will happen anyway as soon as we support updating LuaRocks via rockspec on Windows, so I'm all for it. (Btw., what is holding us back there?)

If all of this is realized, we only have two directory trees left: One Lua installation directory which includes the system rocks tree and the LuaRocks installation, and one separate user rocks tree. This is exactly like on Unix but with some additional subdirectories in `.../lib/luarocks` (tools, icons, etc.).

ignacio:
> (The various tools shouldn't be in PATH, so I'd put them in `.../lib/luarocks/tools` or something.)

Yes, they must not be in the path. Care is taken to address them using absolute paths so a LR install won't mess with my libeay32.dll and so on.

> This is already done on Unix, and it will happen anyway as soon as we support updating LuaRocks via rockspec on Windows, so I'm all for it. (Btw., what is holding us back there?)

IIRC, the trouble is replacing files while in use. [This issue](https://github.com/keplerproject/luarocks/issues/265) mentions it.

siffiejoe:
> IIRC, the trouble is replacing files while in use. This issue mentions it.

Yes, but that should not apply to Lua modules since they are opened, parsed, compiled, and then *closed*. So when LuaRocks code is executed there shouldn't be any Lua files in use. Is it the batch files that are the problem? I'll try to dig up the original source of that info ...

ignacio:
I think the problem was related to C modules which LR can use if available (LuaSocket, LuaSec, etc). Maybe there is no actual problem upgrading LuaRocks itself.

hishamhm:
> I think the problem was related to C modules which LR can use if available (LuaSocket, LuaSec, etc).

Yes, that's why we have `use_fs_modules = false` on Windows now.

> Maybe there is no actual problem upgrading LuaRocks itself.

Yes, maybe there isn't, and to be honest I just never tested it.

hishamhm:
Note however that for self-upgrade to work the first install has to be different, so that the LuaRocks files show up in the manifest. On Unix that required a special rockspec and makefile changes, for which I never implemented equivalents on Windows.

Tieske:
> I'm a bit uncomfortable installing LuaRocks into a directory owned by another program, but ok.

Conceptually this is ok I think. LuaRocks is just a tool that helps extend the Lua environment, hence it has to modify it. So a rockstree is part of a Lua installation, but managed by LuaRocks.

> Another idea was to move the LuaRocks modules from the LuaRocks installation directory into the system rocks tree. This is already done on Unix, and it will happen anyway as soon as we support updating LuaRocks via rockspec on Windows, so I'm all for it. (Btw., what is holding us back there?)

Here it is different imo. LuaRocks is a different application, one could consider it just a coincidence that it is written in Lua itself. But still it is a different application. Hence my hesitation. But reducing platform differences might be worth it.

> But then the LuaRocks batch files should be installed side-by-side with lua.exe so that you also don't have to modify the PATH variable (if you have already done so for Lua). (The various tools shouldn't be in PATH, so I'd put them in .../lib/luarocks/tools or something.)

As an alternative; see [this commit](https://github.com/Tieske/luarocks/commit/99ac529810458db50e2fdb9a471b11ae23c0a4d2), it adds a "shortcut" batch file to the bin directory to access LR itself. And hence the LR installation remains separate from the Lua installation. Just a thought...

Tieske:
as for updating LuaRocks by itself; there are also the included binaries, and batchfiles. I'm not worried about the binaries, but the batchfiles might be an issue.

The way I understand they work is by executing a line, if the batch file is changed/deleted, the next line is read, but might be different because of a change. Probably if multiple statements are within () they are considered one, even if on multipe lines, so that makes it even less predictable.

I think workarounds can be devised, but it does require some care.

Tieske:
for now; should we merge this PR and open a new issue for making LR update itself on windows?

Tieske:
updated to warn if an environment variable is given, but the file referred to isn't available.

Tieske:
@hishamhm Can this be merged or is there still an issue left?

PS. I think we need a different branching model if developing in parallel for LR 3 and LR 2. Something like a development branch, for which all enhancements will be merged in both the master branch (LR 2.x) and in the LR 3 branch. New LR 3 features are only to be merged in the LR 3 branch.

hishamhm:
@Tieske Yes, I believe this can be merged.

As for branches, I've been remiss and committed stuff into `luarocks-3` that should have gone into master instead... Maybe we should think a bit more about the branching model, yes.

PR #418, opened 2015-08-18
ci: updates appveyor config to use VS2015

jasonwilliams200OK:
This PR sets the CI build against Visual Studio 2015.

AppVeyor uses the full-featured free-for-OSS edition of VS2015: Community Edition (http://visualstudio.com/community/). VS2015 packs version of `cl.exe` which compiles C99 (finally), C++1[1/4/7] features + the XP support: http://blogs.msdn.com/b/vcblog/archive/2015/06/19/c-11-14-17-features-in-vs-2015-rtm.aspx. So many `_MSC_VER` directives can be purged away.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,8 +1,5 @@
 version: 2.2.1.{build}-test
 
-os:
-- Windows Server 2012 R2
-
 shallow_clone: true
 
 environment:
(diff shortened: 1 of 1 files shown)


ignacio:
Wouldn't it be better leaving out the "os" section entirely?
VS 2015 is currently available on the default image and using that came in handy in the past (i.e. migration to the new and faster infrastructure: http://www.appveyor.com/blog/2015/06/23/new-oss-build-environment-and-xamarin-support)

jasonwilliams200OK:
@Ignacio, yup that makes sense. I have removed the `os` section. :)
Separately, if we have a solution file, we can build it with MSBuild as `%ProgramFiles(x86)%\MSBuild\14.0\Bin\MSBuild luarocks.sln /p:Configuration=Release` (and for `Debug`, drop `/p:Configuration=Release`) against x86 and x64.

ignacio:
Thanks! I don't think that will be necessary. LuaRocks itself does not need a compiler. It is only used in the tests, when installing other rocks that do need one (like luasocket, for instance).

Thanks for the PR.

Issue #447, opened 2015-11-07
Configure fails under Debian: "Could not find a downloader helper program"


zoresvit:
When I try to compile Luarocks, it fails on configuration stage as it can't find a downloader and md5 checker though both are installed and available in `PATH`:

```
$ ./configure  --with-lua=/home/zoresvit/devel/src/neovim/.deps/usr --with-lua-include=/home/zoresvit/devel/src/neovim/.deps/usr/include/luajit-2.0 --lua-suffix=jit 
Checking Lua interpreter... luajit found in /home/zoresvit/devel/src/neovim/.deps/usr/bin
Checking Lua includes... lua.h found in /home/zoresvit/devel/src/neovim/.deps/usr/include/luajit-2.0/lua.h
Could not find a downloader helper program. Tried: curl wget fetch.
Make sure one of them is installed and available in your PATH.

configure failed.
```

However if I specify them explicitly as arguments, configuration works fine:

```
$ ./configure  --with-lua=/home/zoresvit/devel/src/neovim/.deps/usr --with-lua-include=/home/zoresvit/devel/src/neovim/.deps/usr/include/luajit-2.0 --lua-suffix=jit --with-downloader=wget --with-md5-checker=md5sum
Checking Lua interpreter... luajit found in /home/zoresvit/devel/src/neovim/.deps/usr/bin
Checking Lua includes... lua.h found in /home/zoresvit/devel/src/neovim/.deps/usr/include/luajit-2.0/lua.h
Configuring for system... Linux
Configuring for architecture... x86_64
Using unversioned rocks dir: /usr/local/lib/luarocks/rocks
Writing configuration...

Installation prefix: /usr/local
LuaRocks configuration directory: /usr/local/etc/luarocks
Using Lua from: /home/zoresvit/devel/src/neovim/.deps/usr

Done configuring.
- Type 'make build' and 'make install':
  to install to /usr/local as usual.
- Type 'make bootstrap':
  to install LuaRocks in /usr/local as a rock.
```

The reason of failure is `find_program` function which is fragile:

```
find_program() {
   path="$PATH"
   item="`echo "$path" | sed 's/\([^:]*\):.*/\1/'`"
   path="`echo "$path" | sed -n 's/[^:]*::*\(.*\)/\1/p'`"
   found="no"
   while [ -n "$item" ]
   do
      if [ -f "$item/$1" ]
      then
         found="yes"
         break
      fi
[...]

PR #448, opened 2015-11-07
Fix `find_program` function to be less fragile

zoresvit:
Look for executable path using system utility `which` instead of fragile regular expressions. See #447  for details. Also see discussion in [neovim issues](https://github.com/neovim/neovim/issues/3620#issuecomment-154757577).

### DIFF
--- a/configure
+++ b/configure
@@ -71,26 +71,7 @@ EOF
 # Helper functions
 
 find_program() {
-   path="$PATH"
-   item="`echo "$path" | sed 's/\([^:]*\):.*/\1/'`"
-   path="`echo "$path" | sed -n 's/[^:]*::*\(.*\)/\1/p'`"
-   found="no"
-   while [ -n "$item" ]
-   do
-      if [ -f "$item/$1" ]
-      then
-         found="yes"
-         break
-      fi
-      item="`echo "$path" | sed 's/\([^:]*\):.*/\1/'`"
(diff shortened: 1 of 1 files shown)


siffiejoe:
`which` might not be available everywhere. It is in Debian-based Linuxes (in the package `debianutils`) but it is not required by POSIX.

daurnimator:
Use `command -v` instead of `which`

hishamhm:
@daurnimator does it work with the Debian/Ubuntu /bin/sh?

I don't remember exactly why we went with that function instead of `which` originally, but I think it was because it really wasn't available out of the box in some systems.

daurnimator:
IIRC `command -v` is POSIX specified.
With bash its a builtin.

siffiejoe:
`command -v` is specified by [POSIX](http://pubs.opengroup.org/onlinepubs/9699919799/utilities/command.html).

hishamhm:
I checked the `git blame` history for `find_program` and wow, that code dates back to the CVS days. I think it's safe to replace it with `command -v`, indeed.

mpeterv:
Unlike previous implementation, `command -v` returns path to the program, not the directory containing it.

zoresvit:
@mpeterv There was no comments regarding this in the original implementation and the name of the function `find_program` suggested the full path including the binary. In addition `command -v` worked unlike the original code.

But if the containing directory is required, there's still no need for regular expressions. The single line of code just needs to be replaced with:

```
dirname $(command -v "$1") 2>/dev/null
```

PR #450, opened 2015-11-10
Improve webpages

hugomg:
* Add doctypes (quirksmode is evil)
* Reverse the order of release tarballs (newer releases now on top)


hishamhm:
Thank you! Cheers from JFK airport :)

PR #453, opened 2015-11-14
write config directory only if explictly given, use defaults otherwise

Tieske:
implements nr 2 listed [here](https://github.com/keplerproject/luarocks/issues/403#issuecomment-117988369)

### DIFF
--- a/install.bat
+++ b/install.bat
@@ -8,6 +8,7 @@ local vars = {}
 vars.PREFIX = nil
 vars.VERSION = "2.2"
 vars.SYSCONFDIR = nil
+vars.SYSCONFFORCE = nil
 vars.CONFBACKUPDIR = nil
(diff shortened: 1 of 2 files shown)


hishamhm:
I'm wondering if LR3 would be a good point to clean up some of the compatibitily stuff such as all those checks for unversioned trees and filenames... I noticed that Unix uses SYSCONFDIR and guesses the filename, while Windows uses SYSCONFIG.

Tieske:
Is something expected or can this be merged? would be nice for the next release

ignacio:
Should we merge this and open an issue to unify Unix and Windows behaviour in LR3 ?

hishamhm:
@ignacio I think it's a good idea. Let's get the improvement out now, and if the behavior changes somewhat, LR3 is a new major release so any future behavior changes are justified.

PR #455, opened 2015-11-24
fix: bad errormessage and missing quotes

Tieske:


ignacio:
Looks good.

Tieske:
thx for checking

PR #466, opened 2015-12-14
Use lua-http for http operations if installed

daurnimator:
This pull-request will make luarocks use [lua-http](https://github.com/daurnimator/lua-http) for HTTP operations if it is installed.

----------------------

# REVIEW ONLY

Please don't merge yet; I need to create a lua-http release first.

### DIFF
--- a/src/luarocks/fs/lua.lua
+++ b/src/luarocks/fs/lua.lua
@@ -550,10 +550,35 @@ end
 end
 
 ---------------------------------------------------------------------
+-- Download function
+---------------------------------------------------------------------
+
+-- Default to native downloader
+local download_handlers = setmetatable({}, {__index = function()
+      return fs.use_downloader
+end;})
+
+--- Download a remote file.
+-- @param url string: URL to be fetched.
+-- @param filename string or nil: this function attempts to detect the
+-- resulting local filename of the remote file as the basename of the URL;
+-- if that is not correct (due to a redirection, for example), the local
+-- filename can be given explicitly as this second argument.
+-- @return (boolean, string): true and the filename on success,
+-- false and the error message on failure.
+function fs_lua.download(url, filename, cache)
+   assert(type(url) == "string")
+   assert(type(filename) == "string" or not filename)
+   filename = fs.absolute_name(filename or dir.base_name(url))
+   local scheme = dir.split_url(url)
+   return download_handlers[scheme](url, filename, cache)
+end
+
+---------------------------------------------------------------------
 -- LuaSocket functions
 ---------------------------------------------------------------------
 
-if socket_ok then
+if socket_ok and cfg.no_proxy then -- delegate to the configured downloader so we don't have to deal with whitelists
 
 local ltn12 = require("ltn12")
 local luasec_ok, https = pcall(require, "ssl.https")
@@ -565,14 +590,14 @@ local redirect_protocols = {
 
 local function request(url, method, http, loop_control)
    local result = {}
-   
+
    local proxy = cfg.http_proxy
    if type(proxy) ~= "string" then proxy = nil end
    -- LuaSocket's http.request crashes when given URLs missing the scheme part.
    if proxy and not proxy:find("://") then
       proxy = "http://" .. proxy
    end
-   
+
    if cfg.show_downloads then
       io.write(method.." "..url.." ...\n")
    end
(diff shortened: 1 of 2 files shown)


daurnimator:
Anyone able to review this?

src/luarocks/fs/lua.lua
```diff
@@ -550,10 +551,110 @@ end
...
+         return false, err2
+      end
+   end
+   if cache and headers:has("last-modified") then
+      local tsfd = io.open(filename..".timestamp", "w")
+      if tsfd then
+         tsfd:write(headers:get("last-modified"))
+         tsfd:close()
+      end
+   end
```
mpeterv: file:close() should be added here, or else the file can be used before all contents are saved.

src/luarocks/fs/lua.lua
```diff
@@ -715,13 +776,36 @@ function fs_lua.download(url, filename, cache)
    return true, filename
 end
 
-else --...if socket_ok == false then
+if not rawget(download_handlers, "http") then
+   function download_handlers.http(url, filename, cache)
+      local content, err, https_err = http_request(url, http, cache and filename)
+      if https_err then
+         if not downloader_warning then
+            util.printerr("Warning: falling back to "..cfg.downloader.." - install luasec to get native HTTPS support")
```
mpeterv: Perhaps lua-http should be mentioned here too if it supports https?

mpeterv:
@daurnimator the new code is not covered because lua-http isn't installed on travis-ci =(

src/luarocks/fs/lua.lua
```diff
@@ -550,10 +551,110 @@ end
...
+function fs_lua.download(url, filename, cache)
+   assert(type(url) == "string")
+   assert(type(filename) == "string" or not filename)
+   filename = fs.absolute_name(filename or dir.base_name(url))
+   local scheme = dir.split_url(url)
+   return download_handlers[scheme](url, filename, cache)
+end
+
+-- Try lua-http first
+if http_request_ok then
```
mpeterv: Is lua-http interface going to be as stable as luasocket or maybe a version check should be added here?

hishamhm:
This is a lot of code and each module-backend option we add multiplies the complexity of our test matrix.

Supporting two implementations of the APIs in the codebase (one based on external commands, one based on Lua modules) is already a constant source of headaches, but we currently feel we have to do this because of bootstrapping.

So, the pressing question is: _why_ add yet another code path? Adding this means supporting this in the long-term, having to track new dependencies, and making sure every feature we have or add now would need to work properly on _three_ different scenarios.

src/luarocks/fs/lua.lua
daurnimator: resolved.

src/luarocks/fs/lua.lua
daurnimator: Resolved.

daurnimator:
> @daurnimator the new code is not covered because lua-http isn't installed on travis-ci =(

Indeed. I wasn't sure how best to add it to the testing matrix.
Normally I would add to the travis matrix (e.g. add a new variable "LUA_HTTP=true"); but luarocks doesn't seem to be doing that for anything else?
How is the curl fallback tested vs luasocket?

> Is lua-http interface going to be as stable as luasocket or maybe a version check should be added here?
The lua-http interfaces I've used I consider stable, minus some of the error indications.
I do need to create a versioning scheme for lua-http though.

daurnimator:
> So, the pressing question is: why add yet another code path? Adding this means supporting this in the long-term, having to track new dependencies, and making sure every feature we have or add now would need to work properly on all these new scenarios.

I find the luasocket http module a mess; I created lua-http as a breath of fresh air.
Why prefer lua-http?:
  - Redirects are supported well (luasocket doesn't redirect from http => https; no strange interactions between retries)
  - SSL supported in the one interface
  - HTTP2 is supported
  - Able to be async: in future this means we could download packages in parallel

hishamhm:
You only answered why you like lua-http better than luasocket (which is obvious, since it's your module). 

Unfortunately, you didn't provide a convincing argument as for what what real, current problems does this fix to the point that it would justify the additional complexities I enumerated above.

One alternative way to go about this would be to think how could we make the structure of LuaRocks more modular to accomodate alternative implementations of things. This, I think, would spark a wider architectural discussion.

Tieske:
I think @hishamhm has a good point; we need really, really convincing arguments to add another set of complexities

hishamhm:
@daurnimator Revisiting this, there is one killer feature which I think would be worth adding a lua-http backend to `luarocks.fs`: HTTP persistent connections. Running a command like `luarocks install busted` on a blank system opens a _lot_ of connections to luarocks.org and it seems a noticeable amount of time is spent on that. I did change a bunch of wget calls in the test suite into one multi-file wget call and got nice speedups. Would lua-http be able to that kind of speed up for LuaRocks as well?

daurnimator:
@hishamhm if I get time to finish off the connection pooling stuff it could. All the foundational stuff is there for it, just need to add the actual pool.

PR #467, opened 2015-12-19
Improve builtin build backend errors

mpeterv:
This improves errors in two cases:

* When a file name for a module doesn't have an extension: currently `.o` is shown in the error message instead of actual name.
* When a file for a Lua module doesn't exist: currently failure cause isn't displayed.

To be merged after 2.3.0 is released, I suppose.


daurnimator:
:+1: good to merge I think?

PR #469, opened 2015-12-22
Add git+ssh support to the fetchers.

geoffleyland:
This adds support for git+ssh, which is used for authenticated repo access by at least Github and bitbucket.  I'm not sure master is the right place to put this, but since it adds functionality rather than modifies it, I figure it's non-breaking?

### DIFF
--- a/src/luarocks/fetch/git_ssh.lua
+++ b/src/luarocks/fetch/git_ssh.lua
@@ -2,9 +2,12 @@
 --- Fetch back-end for retrieving sources from Git repositories
 -- that use ssh:// transport. For example, for fetching a repository
 -- that requires the following command line:
--- `git clone git@example.com:username/foo.git
+-- `git clone ssh://git@example.com/path/foo.git
 -- you can use this in the rockspec:
--- source = { url = "git+ssh://git@example.com:username/foo.git" }
+-- source = { url = "git+ssh://git@example.com/path/foo.git" }
+-- It also handles scp-style ssh urls: git@example.com:path/foo.git,
+-- but you have to prepend the "git+ssh://" and why not use the "newer"
+-- style anyway?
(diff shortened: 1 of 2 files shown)


geoffleyland:
This now supports both git+ssh syntaxes (sort of, so long as a git+ssh:// is prepended), thanks to siffiejoe.

siffiejoe:
Works fine with my private gitolite server (both URL styles). Regarding breaking/non-breaking change: It's a breaking change, because older LuaRocks releases won't know how to handle rockspecs with `git+ssh://` in their `source.url`. But new fetchers have been accepted before, because they are supposed to be optional plugins, and the older LuaRocks releases give a clear error message why the source couldn't be fetched.

src/luarocks/fetch/git_ssh.lua
```diff
@@ -0,0 +1,33 @@
...
+-- @param extract boolean: Unused in this module (required for API purposes.)
+-- @param dest_dir string or nil: If set, will extract to the given directory.
+-- @return (string, string) or (nil, string): The absolute pathname of
+-- the fetched source tarball and the temporary directory created to
+-- store it; or nil and an error message.
+function git_ssh.get_sources(rockspec, extract, dest_dir)
+   rockspec.source.url = rockspec.source.url:gsub("^git.", "")
+
+   -- Handle old-style scp-like git ssh urls
+   if rockspec.source.url:match("ssh://[^/]+:[^%d]") then
```
siffiejoe: Maybe you should anchor (`^`) the pattern here as well ...

geoffleyland:
I'm closing this one because I think I screwed things up for my end, so I'll try again.  Sorry about that, because we'll lose the history, but I'll put the anchor in.

PR #470, opened 2015-12-22
Fix util.popen_read

geoffleyland:
util.popen_read wasn't exported (and is used in cfg.lua) and called io.open, not io.popen.  I'm guessing that it should be exported, and should call io.popen?

### DIFF
--- a/src/luarocks/fetch.lua
+++ b/src/luarocks/fetch.lua
@@ -355,7 +355,27 @@ function fetch.get_sources(rockspec, extract, dest_dir)
       if not ok then return nil, err end
       fs.unpack_archive(rockspec.source.file)
       if not fs.exists(rockspec.source.dir) then
-         return nil, "Directory "..rockspec.source.dir.." not found inside archive "..rockspec.source.file, "source.dir", source_file, store_dir
+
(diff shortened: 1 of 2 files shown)


geoffleyland:
Whoops, I don't understand how pull requests work.  I intended to have just the fix for utils.popen_read on this pull request, but when I committed the support for fetch.get_source making wild assumptions about what it downloaded (which is more controversial), it got dragged in to this pull request.  Have a look and comment, and I'll try again later.

hishamhm:
> Whoops, I don't understand how pull requests work.

I made this exact same mistake before. I think the way to avoid this is to create a new branch for each PR.

I fetched your tree and cherry-picked the commit for the fix. I _think_ the cleanest way for us to proceed from here is that we should close this PR and then you `git pull` from luarocks-3 to synchronize and then submit a new PR. Sometimes git can be confusing!

A comment on the source.dir feature: please use `deps.format_is_at_least(rockspec, "3.0")` to version-gate the feature (grep for that in the `luarocks-3` branch for examples).

geoffleyland:
Thanks for committing the popen fix.  I'll try again for the source.dir searching

PR #472, opened 2015-12-22
If an archive contains one top level directory, assume it's the one we're looking for.

geoffleyland:
Sorry, I've totally screwed this up again - I ended up just deleting my repo and starting again, with the consequence that two old pull requests are dead.  I guess I'm learning from my mistakes.

Anyway, this version checks deps.format_is_at_least, and works locally for me.

### DIFF
--- a/src/luarocks/fetch.lua
+++ b/src/luarocks/fetch.lua
@@ -355,7 +355,32 @@ function fetch.get_sources(rockspec, extract, dest_dir)
       if not ok then return nil, err end
       fs.unpack_archive(rockspec.source.file)
       if not fs.exists(rockspec.source.dir) then
-         return nil, "Directory "..rockspec.source.dir.." not found inside archive "..rockspec.source.file, "source.dir", source_file, store_dir
+
+         -- if rockspec.source.dir can't be found, see if we only have one
+         -- directory in store_dir.  If that's the case, assume it's what
+         -- we're looking for.
+         -- Ideally, we'd only do this if rockspec.source.dir was not
+         -- defined in the rockspec, but load_local_repository already
+         -- set it to an inferred value if it wasn't set there, so we
+         -- can't tell its status in the rockspec.
+         -- In any case, we only do this with rockspecs newer than 3.0.
+         local dir_count, found_dir = 0
+
+         if deps.format_is_at_least(rockspec, "3.0") then
+            local files = fs.list_dir()
+            for _, f in ipairs(files) do
+               if fs.is_dir(f) then
+                  dir_count = dir_count + 1
+                  found_dir = f
+               end
+            end
+         end
+
+         if dir_count == 1 then
+            rockspec.source.dir = found_dir
+         else
+            return nil, "Directory "..rockspec.source.dir.." not found inside archive "..rockspec.source.file, "source.dir", source_file, store_dir
+         end
       end
       fs.pop_dir()
    end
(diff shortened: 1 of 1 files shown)


src/luarocks/fetch.lua
```diff
@@ -355,7 +355,32 @@ function fetch.get_sources(rockspec, extract, dest_dir)
       if not ok then return nil, err end
       fs.unpack_archive(rockspec.source.file)
       if not fs.exists(rockspec.source.dir) then
-         return nil, "Directory "..rockspec.source.dir.." not found inside archive "..rockspec.source.file, "source.dir", source_file, store_dir
+
+         -- if rockspec.source.dir can't be found, see if we only have one
+         -- directory in store_dir.  If that's the case, assume it's what
+         -- we're looking for.
+         -- Ideally, we'd only do this if rockspec.source.dir was not
+         -- defined in the rockspec, but load_local_repository already
```
hishamhm: `load_local_rockspec`, you mean.

Wouldn't it be better to set a boolean there (such as `rockspec.source.dir_set`) so we can know here if the value was inferred or explicit?

geoffleyland:
It would be much better.  As a newbie to the code I was trying to tread lightly.  Too lightly.  I've changed it so that `load_local_rockspec` sets `rockspec.source.dir_set` and that's checked in `get_sources`.

hishamhm:
I don't see `dir_set` being used in the other function...

geoffleyland:
Well this is embarrassing.  I thought I'd tested that, but I must have mucked up the rockspec_format = "3.0".  Here's another try.  Thanks for your patience.

src/luarocks/fetch.lua
```diff
@@ -355,7 +356,29 @@ function fetch.get_sources(rockspec, extract, dest_dir)
...
+         -- If rockspec.source.dir can't be found, see if we only have one
+         -- directory in store_dir.  If that's the case, assume it's what
+         -- we're looking for.
+         -- We only do this if the rockspec source.dir was not set, and only
+         -- with rockspecs newer than 3.0.
+         local dir_count, found_dir = 0
+
+         if not rockspec.source.dir_set and deps.format_is_at_least(rockspec, "3.0") then
+            local files = fs.list_dir()
+            for _, f in ipairs(files) do
```
hishamhm: Rethinking this, I believe it would be best to do something like

```lua
   if #files == 1 and fs.is_dir(files[1]) then
      found_dir = f
   end
```

In other words, assume the single directory is the `source.dir` only if it is the only file in the archive. If the archive has multiple top-level files, then `.` would probably be the best `source.dir` (archive does not contain a top-level directory and it's just a bunch of files in the root, typical .zip)

geoffleyland: I just tried this and the trouble is that at this point store_dir contains both the zipfile and whatever we got from unzipping it, so there's (at least) two entries in the directory.  We could delete the zip file (where's the appropriate place to do that?  Just after it's unzipped?) or maybe try to ignore it, but that would get a bit too complicated.

hishamhm: Oh, I see. Deleting it would probably be the best option, but a feature gets in the way: `source.url` can be a single *.lua or *.c, and `fs.unpack_archive` conveniently accepts *.lua and *.c files as a no-op, so again some logic to determine "this is really an archive file that can be deleted" would be needed, which brings us back to square one. :\

I think auto-detecting `"."` will have to stay out for now, then.

PR #474, opened 2015-12-23
Fetch LuaJIT from GitHub mirror on Travis

mpeterv:


mpeterv:
I've merged master with luabitop installation fix into luarocks-3 branch so the build should pass now.

PR #476, opened 2015-12-26
fix missing variable expansion when using a self contained configuration

Tieske:


hishamhm:
I just tagged and packaged 2.3.0-rc2, including this. Ready to release 2.3.0 as soon as I get the OK :)

ignacio:
Just installed this. Tested the /SELFCONTAINED thing and it is fixed.
Also updated my main installation with no problems.

Tieske:
:+1:

PR #478, opened 2016-01-04
Upgrade to LuaCov 0.9.1, make coverage viewable

mpeterv:


mpeterv:
So now you can go to coveralls.io, click on filenames and see which lines are covered and which are missed.

hishamhm:
Always nice to see a shorter configuration! Thanks!

PR #481, opened 2016-01-06
[WIP] Internal support for upgrade and more

mpeterv:
This PR implements a new internal option that specifies how a dependency is resolved:

* "install": always install the latest possible version;
* "upgrade": install the latest possible version unless it's already installed;
* "satisfy": install the latest possible version unless some (matching) version is already installed.

If we interpret `luarocks install foo` as "satisfy dependency `foo`" and `luarocks install foo 2.0` as "satisfy dependency `foo == 2.0`", then various combinations of upgrading/installation logic are expressed as pairs of "install", "upgrade" or "satisfy": the first one for the "root" rock (foo in this example) and the other for all its dependencies, recursively.

So, currently LuaRocks uses ("install", "satisfy"). I think it would be useful to support all the combinations. E.g. ("upgrade", "satisfy") means "upgrade a rock"; ("satisfy", "satisfy") is a lazy install that doesn't hit the network if the rock is already installed. ("upgrade", "upgrade") is deep upgrade, etc.

In this PR I implement this as options for new `install.install_by_name` function used by `install` command and for `deps.satisfy_dependencies`. There are no significant visible behavior changes. On [upgrade-install](https://github.com/mpeterv/luarocks/tree/upgrade-install) branch of my fork I've implemented command-line flags `--upgrade`, `--lazy`, `--upgrade-deps` and `reinstall-deps` for testing.

This is work in progress but any feedback and reviews are welcome. I hope that even a half-broken (well, it sorta works) half-implementation (but see the other branch above) will bring `luarocks upgrade` or something equivalent closer to reality =D

One major issue is that when recursively upgrading or reinstalling dependencies shared dependencies are processed several times. It can be fixed by adding a blacklist argument to `deps.satisfy_dependencies`, but it may be worth it to rewrite it to build closure of all dependencies before installing anything at all.

### DIFF
--- a/src/luarocks/build.lua
+++ b/src/luarocks/build.lua
@@ -160,9 +160,11 @@ end
 -- "all" for all trees, "order" for all trees with priority >= the current default,
 -- "none" for no trees.
 -- @param build_only_deps boolean: true to build the listed dependencies only.
+-- @param deps_install_mode string: Dependency install mode: "install" to always reinstall,
+-- "upgrade" to upgrade to latest version, "satisfy" to only install if missing.
 -- @return (string, string) or (nil, string, [string]): Name and version of
 -- installed rock if succeeded or nil and an error message followed by an error code.
-function build.build_rockspec(rockspec_file, need_to_fetch, minimal_mode, deps_mode, build_only_deps)
+function build.build_rockspec(rockspec_file, need_to_fetch, minimal_mode, deps_mode, build_only_deps, deps_install_mode)
    assert(type(rockspec_file) == "string")
    assert(type(need_to_fetch) == "boolean")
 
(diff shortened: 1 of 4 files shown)


mpeterv:
Even when this PR is finished it's just some dead code unless the interface is somehow exposed. So, can we decide what interface we'll have for upgrading/etc? It would be great if the interface changes could be split into breaking and non-breaking, and then the latter part could be implemented in LuaRocks 2.4.0 instead of waiting for 3.0.0.

For example, if we want to make `luarocks install` lazy by default (no network access if the rock is already installed), we still need a way to upgrade the rock and/or its dependencies. So we add `--upgrade` and `--deep-upgrade` options, for instance. In that case, these options can be implemented in LR 2 without breaking anything.

PR #483, opened 2016-01-11
provide a better error message for missing json libraries

Tieske:


src/luarocks/upload/api.lua
```diff
@@ -126,7 +135,7 @@ if not ltn12_ok then -- If not using LuaSocket and/or LuaSec...
 function Api:request(url, params, post_params)
    local vars = cfg.variables
    local json_ok, json = require_json()
-   if not json_ok then return nil, "A JSON library is required for this command." end
+   if not json_ok then return nil, "A JSON library is required for this command. "..json end
```
daurnimator: this same log message is in multiple places.

I'll also note that this is a minor merge conflict with #466

daurnimator:
Btw, these json libraries have slightly different apis e.g. on invalid json:
  - cjson throws an error 
  - dkjson returns `nil, position, error_message`

Is this something we need to handle?

src/luarocks/upload/api.lua
Tieske: Thx. good catch. I fixed the 2nd occurence.

Tieske:
> Btw, these json libraries have slightly different apis e.g. on invalid json:

Not until it proves a problem imo

mpeterv:
LGTM

PR #491, opened 2016-01-30
add new platform: MXE

starius:
MXE (M cross environment) is a Makefile that compiles a cross compiler and cross compiles many free libraries such as SDL and Qt. Target OS is MinGW.

MXE has [a plugin called "luarocks"](https://github.com/mxe/mxe/blob/master/plugins/luarocks/README.md) which adds a luarocks itself and dozens of packages. It applies [a patch](https://github.com/mxe/mxe/blob/master/plugins/luarocks/luarocks-1-fixes.patch) to luarocks to make it usable in MXE. I have done this work for luarocks 2.2.2 and then have to [rebase](https://github.com/mxe/mxe/commit/4f2f9bf8064a0ad8733a2297d5a0cbcad7f35d3f) this patch for luarocks 2.3.0. This pull request is equal to that patch.

See also https://github.com/mxe/mxe/pull/1017 and https://github.com/mxe/mxe/pull/1182

### DIFF
--- a/src/luarocks/cfg.lua
+++ b/src/luarocks/cfg.lua
@@ -120,6 +120,9 @@ elseif system and system:match("^MINGW") then
    cfg.platforms.windows = true
    cfg.platforms.mingw32 = true
    cfg.platforms.win32 = true
+elseif system and system:match("^MXE") then
+   cfg.platforms.unix = true
+   cfg.platforms.mxe = true
 else
    cfg.platforms.unix = true
    -- Fall back to Unix in unknown systems.
@@ -137,6 +140,7 @@ local platform_order = {
    linux = 7,
    macosx = 8,
    cygwin = 9,
+   mxe = 13,
    -- Windows
    win32 = 10,
    mingw32 = 11,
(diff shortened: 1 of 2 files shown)


src/luarocks/cfg.lua
```diff
@@ -137,6 +140,7 @@ local platform_order = {
    linux = 7,
    macosx = 8,
    cygwin = 9,
+   mxe = 13,
```
hishamhm: this ordering is weird... please put the numbers in order (you can change the other numbers if needed)

hishamhm:
This sounds really specific. The code assumes that LuaRocks was installed in the same prefix as your MXE tools (in other words, that it is being installed by your MXE script/makefile). If the only way to use this is in conjunction with the MXE environment anyway, it seems to me that this custom config belongs there rather than here.

Another important observation is that virtually all changes are in configuration. Instead of patching LuaRocks, you could obtain the same results writing a config.lua file for your environment that sets these values. The only thing that would break is if you want to add "mxe" platform-overrides to your rockspecs, but if you are patching them, that's not the case, I assume.

Apart from that, the only real change in the could would be 

>    if cfg.is_platform("cygwin") or cfg.is_platform("mxe") then

which we could change to make non-cygwin-specific, by changing it to a setting like `cfg.link_lua_explicitly` (set by default to true on cygwin only). With this one-line change to LuaRocks and a custom config.lua, I think you could use it on MXE without patching LuaRocks.

starius:
You are right: most of my changes can be moved to config.lua.
But [two changes](https://github.com/LuaAndC/mxe/blob/6ec7dff3ad613564eb27bbffdb7e094d05d76684/plugins/luarocks/luarocks-1-fixes.patch) can not be moved to config.

Can you add "mxe" to the list of platforms to remove one of the remaining changes, please?

>> if cfg.is_platform("cygwin") or cfg.is_platform("mxe") then

> which we could change to make non-cygwin-specific, by changing it to a setting like cfg.link_lua_explicitly (set by default to true on cygwin only).

+1 for that change. Can you do it, please?

hishamhm:
@starius I just made a significant change to make it possible for you to set `platforms = { "unix", "mxe" }` in the config file, and I also added the setting `link_lua_explicitly = true`. With these changes, I think everything can be done via config file. Please try the latest code in Git!

PR #492, opened 2016-02-01
Adds 'shell' command.

ignacio:
Adds the 'luarocks shell' command, that launches a new shell with the environment variables LUA_PATH and LUA_CPATH set with the values configured when installing LuaRocks.

This was suggested [here](https://twitter.com/hisham_hm/status/682227274876338176).

### DIFF
--- a/src/luarocks/shell.lua
+++ b/src/luarocks/shell.lua
@@ -1,0 +1,38 @@
+
+--- @module luarocks.shell
+-- Driver for the `luarocks shell` command.
+local shell_cmd = {}
+
+local util = require("luarocks.util")
+local cfg = require("luarocks.cfg")
+local fs = require("luarocks.fs")
+
+shell_cmd.help_summary = "Starts a new shell with the variables from `luarocks path` correctly set."
+shell_cmd.help_arguments = ""
+shell_cmd.help = [[
+Starts a new shell with the variables LUA_PATH and LUA_CPATH set as 
+configured by the LuaRocks installation.
+]]
+
+--- Driver function for "shell" command.
+-- @return boolean This function always succeeds.
+function shell_cmd.run(...)
+   local lr_path, lr_cpath = cfg.package_paths()
+   local path_sep = cfg.export_path_separator
+   
+   lr_path =  lr_path.. ";" .. package.path
+   lr_cpath = lr_cpath .. ";" .. package.cpath
+
+   lr_path = cfg.export_lua_path:format(util.remove_path_dupes(lr_path, ';'))
+   lr_cpath = cfg.export_lua_cpath:format(util.remove_path_dupes(lr_cpath, ';'))
+
+   local shell, cmd = os.getenv("SHELL")
+   if shell then
+      cmd = ([[LUA_PATH="%s" LUA_CPATH="%s" "%s"]]):format(lr_path, lr_cpath, shell)
+   else
+      cmd = ([[cmd /k "%s && %s"]]):format(lr_path, lr_cpath)
+   end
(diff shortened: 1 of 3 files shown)


src/luarocks/shell.lua
```diff
@@ -0,0 +1,38 @@
...
+shell_cmd.help = [[
+Starts a new shell with the variables LUA_PATH and LUA_CPATH set as 
+configured by the LuaRocks installation.
+]]
+
+--- Driver function for "shell" command.
+-- @return boolean This function always succeeds.
+function shell_cmd.run(...)
+   local lr_path, lr_cpath = cfg.package_paths()
+   local path_sep = cfg.export_path_separator
```
mpeterv: Unused

src/luarocks/shell.lua
```diff
@@ -0,0 +1,38 @@
...
+   lr_path = cfg.export_lua_path:format(util.remove_path_dupes(lr_path, ';'))
+   lr_cpath = cfg.export_lua_cpath:format(util.remove_path_dupes(lr_cpath, ';'))
+
+   local shell, cmd = os.getenv("SHELL")
+   if shell then
+      cmd = ([[LUA_PATH="%s" LUA_CPATH="%s" "%s"]]):format(lr_path, lr_cpath, shell)
+   else
+      cmd = ([[cmd /k "%s && %s"]]):format(lr_path, lr_cpath)
+   end
+   return fs.execute(cmd)
```
mpeterv: This causes a crash if the shell exits with non-zero (e.g. run `luarocks shell` and then `exit 1`)

hishamhm:
I've been experimenting locally with this idea too. @Tieske suggested this could be implemented as a command, `luarocks-shell`. The command, if implemented by writing it during build, may be less fragile and less dynamic (pros and cons there); I haven't figured out which one is really best.

I think it also needs to set the PATH, and add not only the path of bin/ entries for the active rock trees in it, but also the path of the LuaRocks binaries itself. (i.e. it should be guaranteed that a plain `luarocks` invocation works inside a launched shell).

In any case, having something like `luarocks-shell` in my machine for a few days, I noticed it's something I ended up using a lot. It will be a very useful feature.

ignacio:
Ok, thanks both @hishamhm and @mpeterv for the pointers. I'll update this PR, but the possibility of generating a script at install time is not something I would oppose.

ignacio:
@hishamhm could you share the script you're using? I'm taking a look at what would be needed to generate it and while the Windows side of things is clear to me, I'm a bit lost as to what to do on the "other" side :)

hishamhm:
This ended up falling on the wayside, sorry. My initial attempt at a stand-alone `luarocks-shell` for Unix was simply this:

```bash
#!/bin/sh

export PATH=$(dirname $(readlink -f "$0"))":$PATH"
eval $(luarocks path --bin)
export LUA_PATH LUA_CPATH PATH
exec "$SHELL" "$@"
```

Tieske:
I like this approach (just ran into this PR for the first time, dunno how I missed it before)

If the script is part of the LR installation, then based on its own location it would be able to find everything it needs right? No need to do hardcoded install time stuff...

hishamhm:
@Tieske Would it be possible to launch a CMD instance that auto-runs the Visual Studio setup batch file on startup then gives you an interactive shell? This would be necessary to give us a "shell that works" on Windows.

(We could also check that this Visual Studio batch file can be found during INSTALL.BAT)

Tieske:
That is already the case (if it could be found) see https://github.com/luarocks/luarocks/blob/master/install.bat#L960

PR #495, opened 2016-02-07
Add compatibility with lhf's lmd5 module.

siffiejoe:
Monkey-patch the module table of lhf's lmd5 module to add the `sumhexa` function, the only md5-related function LuaRocks currently uses.

### DIFF
--- a/src/luarocks/fs/lua.lua
+++ b/src/luarocks/fs/lua.lua
@@ -728,6 +728,14 @@ end
 
 if md5_ok then
 
+-- Support the interface of lmd5 by lhf in addition to md5 by Roberto
+-- and the keplerproject.
+if not md5.sumhexa and md5.digest then
+   md5.sumhexa = function(msg)
+      return md5.digest(msg)
+   end
+end
+
 --- Get the MD5 checksum for a file.
 -- @param file string: The file to be computed.
 -- @return string: The MD5 checksum or nil + error


ignacio:
Thank you

src/luarocks/fs/lua.lua
```diff
@@ -728,6 +728,14 @@ end
 
 if md5_ok then
 
+-- Support the interface of lmd5 by lhf in addition to md5 by Roberto
+-- and the keplerproject.
+if not md5.sumhexa and md5.digest then
+   md5.sumhexa = function(msg)
+      return md5.digest(msg)
+   end
+end
```
hishamhm: why not just `md5.sumhexa = md5.sumhexa or md5.digest` ?

siffiejoe: `md5.digest` has additional optional arguments. I could have written

```lua
md5.sumhexa = md5.sumhexa or function(msg)
   return md5.digest(msg)
end
```

but I started out with

```lua
if md5.sumhexa then
   print("md5 by roberto")
elseif md5.digest then
   print("lmd5 by lhf")
   md5.sumhexa = function(msg)
      return md5.digest(msg)
   end
else
   error("unsupported md5 module")
end
```

and after removing the debugging output that's what was left.

PR #497, opened 2016-02-11
Fix loud archive unpacking

mpeterv:
In luarocks.fs.unix.tools.unpack_archive:

* Execute commands in quiet mode, like in fs.unzip
* Quote archive name
* Use cfg.vars.TAR instead of literal "tar" when unpacking .tar.bz2

### DIFF
--- a/src/luarocks/fs/unix/tools.lua
+++ b/src/luarocks/fs/unix/tools.lua
@@ -293,13 +293,19 @@ end
 function tools.unpack_archive(archive)
    assert(type(archive) == "string")
 
+   local pipe_to_tar = " | "..vars.TAR.." -xf -"
+
+   if not cfg.verbose then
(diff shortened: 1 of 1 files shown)


hishamhm:
Not sure how I feel about this... I occasionally find it useful to see the code being unpacked (gives a glimp if there's anything else in the ), kind of like seeing the compiler lines as the modules are built. More opinions are welcome!

mpeterv:
@hishamhm well, the Windows implementation is quiet. And unpacking normal source rocks doesn't produce noise, so I think it's consistent to make source file unpacking quiet as well. Perhaps it should be loud if `--verbose` is used though.

mpeterv:
Update:

* Separated quoting fixes, pushed them to master.
* Commands for unpacking .tar.gz and .tar.bz2 are only silenced when --verbose is used, same way fs.execute_quiet does it.

hishamhm:
Merged, thank you!!

PR #498, opened 2016-02-11
Make new_version work with scm rockspecs

mpeterv:
I'd like to have a single scm rockspec in my repo and make rockspecs for releases by adding source.tag field. `luarocks new_version`, given an scm rockspec and a version, currently prints

```
Warning: invalid URL - Error fetching file: Unsupported protocol git+https
Warning: URL is the same, but MD5 has changed. Old rockspec is broken.
```

and produces a rockspec without source.tag.

This PR fixes some behaviour issues in MD5 checking (it is currently checked even if the old rockspec doesn't have source.md5 field or if source archive fetching failed). For the scm rockspec usecase, it adds --tag flag that sets source.tag field in the new rockspec. It's also used to infer new version if it's not given. Common convention `tag == 'v' .. version` is supported. So, now I can run `luarocks new-rockspec myrock-scm-1.rockspec --tag 0.2.0` to create a rockspec for 0.2.0 release.

### DIFF
--- a/src/luarocks/new_version.lua
+++ b/src/luarocks/new_version.lua
@@ -48,31 +48,47 @@ local function try_replace(tbl, field, old, new)
    return false
 end
 
-local function check_url_and_update_md5(out_rs, out_name)
-   local old_md5 = out_rs.source.md5
-   out_rs.source.md5 = nil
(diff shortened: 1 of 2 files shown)


hishamhm:
This sounds really useful! Thanks!

PR #504, opened 2016-02-19
Use mkstemp()/mkdtemp() where available

Alloyed:
This PR makes luarocks

* Use mkdtemp() to create temporary directories on unix platforms
* Respect $TMPDIR in environments where it has been set

as suggested by <http://0pointer.net/blog/projects/tmp.html>

Relevant snippet:

> Usually it is recommended to use $TMPDIR if it is set before falling back to /tmp directly. As mentioned, this is a tmpfs on many Linuxes/Unixes (and most likely will be for most soon), and hence should be used only for small files. It's generally a shared namespace, hence the only APIs for using it should be mkstemp(), mkdtemp() (and friends) to be entirely safe

When luaposix is installed, they will be called directly, otherwise the shell [mktemp](http://linux.die.net/man/1/mktemp) will be used instead.

Not tested on MacOS/BSD.

### DIFF
--- a/src/luarocks/fs/lua.lua
+++ b/src/luarocks/fs/lua.lua
@@ -28,8 +28,6 @@ local patch = require("luarocks.tools.patch")
 
 local dir_stack = {}
 
-math.randomseed(os.time())
-
 local dir_separator = "/"
 
 --- Quote argument for shell processing.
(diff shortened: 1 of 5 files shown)


src/luarocks/fs/lua.lua
```diff
@@ -785,6 +770,40 @@ function fs_lua.get_permissions(file)
...
+   if temp_dir then
+      return temp_dir
+   else
+      return nil, err
+   end
+end
+
+--- Create and return the name of a temporary file.
+-- @return string or (nil, string): name of temporary file or (nil, error message) on failure.
+function fs_lua.tmpname()
```
hishamhm: Hi! Thanks for contributing! 
I'm confused about this one: in which way is this better than just using `os.tmpname()`? Just the fact that we add `luarocks_` to the name?

Alloyed: When I wrote this I had assumed from the name that os.tmpname() just used tmpnam(), but actually looking at it the PUC lua source it looks like it will also use mkstemp where it's available.
This is probably not necessary then, the only difference would be that it would respect $TMPDIR where the os.tmpname() just hardcodes "/tmp"

hishamhm: Assuming `/tmp` is not too bad (we've never had problems with that). Could you restore the `fs.unix` version then?

hishamhm:
Merged, thanks! Sorry about the delay!

PR #516, opened 2016-03-11
Add HaikuOS as platform

xspager:
Hi! Is it possible to add [Haiku](https://www.haiku-os.org/) as a supported platform? It is Posix compliant but need some platform specific options like linking with lnetwork in the case of luasocket.

### DIFF
--- a/src/luarocks/cfg.lua
+++ b/src/luarocks/cfg.lua
@@ -120,6 +120,9 @@ elseif system and system:match("^MINGW") then
    cfg.platforms.windows = true
    cfg.platforms.mingw32 = true
    cfg.platforms.win32 = true
+elseif system == "Haiku" then
+   cfg.platforms.unix = true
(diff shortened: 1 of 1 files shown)


src/luarocks/cfg.lua
```diff
@@ -140,8 +143,8 @@ local platform_order = {
    -- Windows
    win32 = 10,
    mingw32 = 11,
-   windows = 12 }
-
+   windows = 12, 
+   haiku = 13 }
```
hishamhm: Please put haiku after cygwin as 10 and increment the ones below it. These numbers don't need to be the same (only their order matters) but I guess it looks nicer if we keep all the Unixes and Windows together. Thanks!

xspager:
Done!

hishamhm:
Thanks!

xspager:
Awesome! Thanks!

Issue #520, opened 2016-03-17
Improve error messages for rocks not found


Tieske:
See #519

Basically, installing `lapis` on Lua 5.2, simply says it cannot be found. It might be better that the error message displayed the constraints within it was not found. eg.

Lapis not found within constraints; Lua 5.2

Maybe someone else has a better idea, anyway a distinction in errormessage for 'not found' versus 'constraints not met'

siffiejoe:
I think the `lua` dependency is special in that there are separate manifests for each Lua version. And in the Lua 5.2 manifest there *isn't* a `lapis`. To get the necessary information we'd have to include packages (at least the latest version) that don't have a 5.2 compatible rock.

hishamhm:
What @siffiejoe said is correct. With the existing manifest, there's no way to know the difference between "the rock doesn't exist" and "the rock doesn't exist for your Lua version"...

ignacio:
It is not an ideal solution, and we know that people don't read, but, could we see if we can get away with adding a message, directing users to LuaRocks site when a search has been made and no results were found? Maybe a special page or some docs, stating why does some rock that is known to exist does not appear as a result when searching.

Tieske:
We know we're looking for 5.2. So can we add a message listing the constraints we looked for? And additionally add a message like; 

"Please check the package name and dependencies, or look it up on luarocks.org."

mpeterv:
The constraints don't mention 'lua' explicitly. Constraints, when searching a version of some rock, include only restrictions for the rock itself, but not for its dependencies or other installed or provided rocks. So just printing constraints doesn't help.

mpeterv:
But I think it should be possible to look through other manifests when the appropriate one doesn't contain the rock, and print supported Lua versions.

A similar, more rare case is when the rock exists for installed Lua version but no versions match constraints. The message should say that existing versions don't match constraints instead of "no results".

Tieske:
don't know the code, my thought simply was; if we're searching the 5.2 manifest, then we know we're doing 5.2. No?

ignacio:
I would go with @Tieske and just add a message pointing users to LuaRocks. For instance, my LR is 2.3.0 + Lua 5.1. `base2base` is a module for 5.3 only. I'd show:

```
luarocks search base2base

No compatible rocks found, or the rock does not exist.
Please check the rock's name or make sure the rock exists and is compatible with Lua 5.1 by visiting:
https://luarocks.org/search?q=base2base
```

[...]

PR #524, opened 2016-03-19
Show supported Lua versions on fail in search.find_suitable_rock

mpeterv:
This PR essentially does two things: first, it ensures that error message returned by `search.find_suitable_rock` is always shown, often together with some context from its caller (the function is used in `luarocks <command> <rock> [<version>]` where`<command`> is `install`, `build`, `unpack` or `download`, and also to satisfy dependencies).

Then, it uses new `lua_version` argument for some manifest loading functions to check if a query that wasn't satisfied using manifest for installed Lua version can be satisfied for some other ones. If there are matching rock versions for other Lua versions, they (Lua versions) are listed in the error message. It looks like this:

```
$ luarocks install lapis

Error: lapis supports only Lua 5.1 but not Lua 5.3.
$ luarocks install sailor 0.5

Error: sailor 0.5 supports only Lua 5.1 and Lua 5.2 but not Lua 5.3.
$ luarocks install rock_depending_on_sailor_0.5_or_newer

Missing dependencies for rock_depending_on_sailor_0.5_or_newer:
sailor >= 0.5

Error: Could not satisfy dependency sailor >= 0.5: Matching sailor versions support only Lua 5.1 and Lua 5.2 but not Lua 5.3.
```

`luarocks search <rock>` isn't affected because the query can match several rocks for it, so another `search.*` function is used. It should use the same logic if nothing was found and simply flip `exact_name` flag when checking for supported Lua versions. There will be a follow-up PR for that if this approach is accepted.

ref #520

### DIFF
--- a/src/luarocks/search.lua
+++ b/src/luarocks/search.lua
@@ -369,7 +369,7 @@ function search.act_on_src_or_rockspec(action, name, version, ...)
    query.arch = "src|rockspec"
    local url, err = search.find_suitable_rock(query)
    if not url then
-      return nil, "Could not find a result named "..name..(version and " "..version or "").."."
+      return nil, "Could not find a result named "..name..(version and " "..version or "")..": "..err
    end
    return action(url, ...)
 end
--- a/src/luarocks/search.lua
+++ b/src/luarocks/search.lua
@@ -170,14 +170,15 @@ end
 -- If the arch field is omitted, the local architecture (cfg.arch)
 -- is used. The special value "any" is also recognized, returning all
 -- matches regardless of architecture.
+-- @param lua_version string: Lua version in "5.x" format, defaults to installed version.
 -- @return true or, in case of errors, nil, an error message and an optional error code.
-function search.manifest_search(results, repo, query)
+function search.manifest_search(results, repo, query, lua_version)
    assert(type(results) == "table")
    assert(type(repo) == "string")
    assert(type(query) == "table")
    
    query_arch_as_table(query)
-   local manifest, err, errcode = manif.load_manifest(repo)
+   local manifest, err, errcode = manif.load_manifest(repo, lua_version)
    if not manifest then
       return nil, err, errcode
    end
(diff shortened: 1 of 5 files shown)


hishamhm:
Wow, don't get me wrong, I'm all for good UI, but that's a lot of code for just a better error message!

I'm also worried about how LuaJIT comes into play, especially since the idea is to support `luajit >= 2.1` style dependencies in the near future. I think @ignacio's suggestion of the error message pointing to the website might be more accurate/useful in the end.

I see this PR includes several commits, some of them are more general cleanups. Maybe some of them could be picked up regardless?

mpeterv:
> I see this PR includes several commits, some of them are more general cleanups. Maybe some of them could be picked up regardless?

Right, I'll cherry-pick some parts of the first three commits.

> Wow, don't get me wrong, I'm all for good UI, but that's a lot of code for just a better error message!

Yes, but it's a pretty good message IMO :smile: Directing users to website to look at information that is available to LuaRocks already is a bit strange: LuaRocks stops and says "no, not going to help you with that, look it up yourself". It would be OK if at least 'luarocks search' searched through all manifests, so that user didn't have to switch between terminal and browser.

As for luajit, this depends on how we support it, if there will be another manifest for luajit rocks then this PR will just work (almost).

mpeterv:
Split, cherry-picked, rebased and repushed.

src/luarocks/search.lua
```diff
@@ -278,6 +280,23 @@ local function pick_latest_version(name, versions)
    return nil
 end
 
+-- Find out which other Lua versions provide rock versions matching a query,
+-- @param query table: A dependency query matching a single rock.
+-- @return table: array of Lua versions supported, in "5.x" format.
+local function supported_lua_vesions(query)
```
hishamhm: Typo: vesions → versions

mpeterv:
Fixed the typo and the merge conflict.

hishamhm:
Ok, merging this! If we improve LuaJIT support, I'm counting on you to update this feature ;)

mpeterv:
> If we improve LuaJIT support, I'm counting on you to update this feature ;)

No problem! Thanks for the merge!

PR #543, opened 2016-04-17
Add support for the --tree argument to the path command

ryanplusplus:
I was trying to cobble together a solution for project-local dependency installations using `--tree` and wanted to add these to all of the appropriate paths, but found that `luarocks path` does not respect the `--tree` argument.

This adds support for the `--tree` argument in a way that is hopefully not too stupid. Any feedback is very much appreciated.

### DIFF
--- a/src/luarocks/cfg.lua
+++ b/src/luarocks/cfg.lua
@@ -666,17 +666,23 @@ function cfg.make_paths_from_tree(tree)
    return lua_path, lib_path, bin_path
 end
 
-function cfg.package_paths()
+function cfg.package_paths(current)
    local new_path, new_cpath, new_bin = {}, {}, {}
-   for _,tree in ipairs(cfg.rocks_trees) do
+   local function add_tree_to_paths(tree)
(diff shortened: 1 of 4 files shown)


hishamhm:
Hi! Thank you for contributing!

Could you revert all those whitespace-only changes in your patch? It makes it hard to read the diff, and adds many unrelated entries to the history of files.

ryanplusplus:
@hishamhm Yes, I can revert them, but are you sure you want them reverted? It's probably a good idea to get them cleaned up since many editors remove trailing whitespace automatically. As far as commit history, all of the whitespace changes are in files that include actual changes so the git log will still make sense. Finally, you can ignore whitespace-only changes by appending `?w=1` to the pull request URL to make it easier to review.

ryanplusplus:
Well, you used to be able to add `?w=1` to ignore whitespace changes, but apparently that no longer works. Thanks GitHub.

ryanplusplus:
All whitespace changes removed except one where the tabbing was incorrect and inconsistent within a modified function.

hishamhm:
Merged! Thank you!

PR #545, opened 2016-04-18
Add a more descriptive error message

yawnt:
I also had to google and find https://github.com/keplerproject/luarocks/issues/238 to solve the issue.
This should prevent further confusion in the future :)

### DIFF
--- a/src/luarocks/fetch.lua
+++ b/src/luarocks/fetch.lua
@@ -151,7 +151,7 @@ function fetch.fetch_and_unpack_rock(rock_file, dest)
       unpack_dir = dest
       local ok, err = fs.make_dir(unpack_dir)
       if not ok then
-         return nil, "Failed unpacking rock file: " .. err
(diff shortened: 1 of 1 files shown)


Tieske:
This is platform specific, so I think the error message should be constructed based on the unpack utility configured.

hishamhm:
On Unix it is possible to figure out based on the error code whether the command failed because the program was not found (error code 127: on Lua 5.1,  32512 (0x7f00); on Lua 5.2+, third result is 127 (0x7f)), not sure if possible on Windows.

In the latest code, `configure` now checks that `unzip` is present at build time, so I think these mystery errors won't happen in the future. (On Windows the tool is bundled, to the errors are even more unlikely.) Not sure if we should tweak the error message, but I'm still open for opinions!

yawnt:
mmm, i installed luarocks from the latest distributed package.. if master does have a check for unzip in `configure`, i'd say this is redundant :)

hishamhm:
@yawnt ah, ok :) closing this for now, then!

Issue #534, opened 2016-04-03
Incorrect MSVC setup command for VS 2012 x64


mpeterv:
As you can see here: https://ci.appveyor.com/project/ignacio/lua-appveyor-example/build/0.0.1.203-test/job/snkhgm22rgwksmar#L527 vcvarsall.bat for VS 2012 fails when argument is `amd64`.

The build still passes because a script executed same vcvarsall.bat passing `x86_amd64` instead beforehand. That appears to be working for other VS versions, too. I think that `amd64` should be replaced by `x86_amd64` in install.bat, or the latter should be used as fallback if amd64 fails.

ignacio:
@xpol did a lot of research on this matter so he could comment on this, but iirc, `x86_amd64` is used for cross compiling (see #443)

The VS 2012 installation on appveyor does not have what is needed. I think that they omitted to install x64 support. I don't have a working VS2012 install so I cannot test (I uninstalled it) but it left a lot of files on my drive :) So, part of the contents of `vcvarsall.bat` are:

```
@echo off
if "%1" == "" goto x86
if not "%2" == "" goto usage

if /i %1 == x86       goto x86
if /i %1 == amd64     goto amd64
if /i %1 == x64       goto amd64
if /i %1 == arm       goto arm
if /i %1 == x86_arm   goto x86_arm
if /i %1 == x86_amd64 goto x86_amd64
goto usage

:x86
if not exist "%~dp0bin\vcvars32.bat" goto missing
call "%~dp0bin\vcvars32.bat"
goto :SetVisualStudioVersion

:amd64
if not exist "%~dp0bin\amd64\vcvars64.bat" goto missing
call "%~dp0bin\amd64\vcvars64.bat"
goto :SetVisualStudioVersion
````

I should fix lua-appveyor-example, but I don't think this is an issue with LuaRocks.

mpeterv:
Well, x64 support is there, or the build would fail, it's just implemented by 32bit compiler. It's kind of cross-compiling because compiler and compiled programs have different bitness, but not really because compiled program have same architecture as host (and as lua binary). I think that LuaRocks should support this:

* Need to build x86 rocks because Lua is x86 -> pass "x86" or ""
* Need to build x64 rocks because Lua is x64 ->
  - If there is x64 compiler targeting x64 pass "amd64"
  - If there is x86 compiler targeting x64 pass "x86_amd64"

ignacio:
[...]

PR #548, opened 2016-05-03
Add cross compiling support for MSVC.

xpol:
This is needed, when native compiler is not installed but cross compiler does.

See #534.

### DIFF
--- a/install.bat
+++ b/install.bat
@@ -525,6 +525,14 @@ local function get_msvc_env_setup_cmd()
 				return ('call "%s"'):format(full_path)
 			end
 		end
+
+		-- try vcvarsall.bat in case MS changes the undocumented bat files above.
+		-- but this way we don't konw if specified compiler is installed...
+		local vcvarsall = vcdir .. 'vcvarsall.bat'
+		if exists(vcvarsall) then
+			local vcvarsall_args = { x86 = "", x86_64 = " amd64" }
+			return ('call "%s"%s'):format(vcvarsall, vcvarsall_args[vars.UNAME_M])
+		end
 	end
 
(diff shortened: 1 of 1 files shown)


mpeterv:
Thank you @xpol! Haven't tested this yet but it looks good. By the way, is there a reason why `install.bat` is both a batch file and a Lua script? It messes up syntax highlighting a lot. Would it make more sense to move the Lua code into `install.lua` and call that from `install.bat`?

mpeterv:
Tested, it works.

xpol:
I just choose syntax manually.
I think it is good to separate bat and Lua. But I don't known how it works, I just update it.

xpol:
i tested with vs2013.

Tieske:
> By the way, is there a reason why install.bat is both a batch file and a Lua script?

The installer used to be a batch file, but it was very cumbersome and hard to maintain, so it was rewritten as a Lua file (by @ignacio). I for one would like to keep it this way, the repo is already dragging so much files along, keeping this in one file is a bigger benefit than the syntax highlighting imo.

mpeterv:
@xpol I think we also need to ensure that these scripts that we now use directly exist for all VS versions, because unlike vcvarsall.bat their paths are not documented. Or, better, fall back to vcvarsall.bat if arch-specific scripts are not found.

hishamhm:
I want to make a new minor release. Windowsfolk, is this good for merging?

xpol:
@mpeterv 
commit 11b65a6 has fallback to documented vcvarsall.bat when the undocumented bats not found.

mpeterv:
Tested this locally with some VS versions and on appveyor with all combinations of VS version/arch.

PR #496, opened 2016-02-10
Implement --upgrade and other flags

mpeterv:
This implements flags `--upgrade`, `--lazy`, `--upgrade-deps` and `--reinstall-deps` flags for `install`, `make` and `build` commands on top of modified version of #481.

`install --upgrade rock` does nothing if the latest version of rock is installed. `install --lazy rock` does nothing if any version of rock is installed. `install --upgrade-deps rock` also recursively upgrades dependencies of rock. `install --reinstall-deps rock` also recursively reinstalls them.

This changes output format for dependencies: instead of printing a summary of all missing dependencies before installing all of them, each installation of a dep is announced separately immediately before installation. This allows handling each dependency separately, in a single function (called deps.fulfil_requirement, it's also used to install rock in `luarocks install rock`), without checking status of each one beforehand, which is not cheap for upgrades.

Is it possible to get this, possibly with different interface, in LR2?

TODO:
- Possibly replace `--upgrade-deps` with `--deep-upgrade`, which would imply `--upgrade`.
- Possibly rename `--reinstall-deps` to `--ignore-installed`, like in pip - there it also works for the main rock, not only for deps, because `--lazy` is default.
- Document the new options.
- Add tests.

PR #551, opened 2016-05-10
Change output and order of deps installation

mpeterv:
This PR makes output for `luarocks install` and other commands that satisfy dependencies closer to what #496 does:

* Process dependencies in the order of `rockspec.dependencies` array.
* Announce each missing dependency just before installing it, instead of listing them all beforehand (makes it easier to understand why each rock is installed in a long list).
* Show what versions are already installed for missing dependencies.
* Remove not very useful (IMO) lines like `switching to build mode` and `updating manifest`.

See comparison of outputs of `luarocks install sailor`: https://gist.github.com/mpeterv/a43fbcdeadaec40dd6fa3284ea7f07dd

In a nutshell:

```
Using https://luarocks.org/valua-0.3-1.src.rock... switching to 'build' mode
Updating manifest for /home/mpeterv/luarocks/here/lib/luarocks/rocks
valua 0.3-1 is now built and installed in /home/mpeterv/luarocks/here (license: MIT)
```

changes to

```
sailor 0.5-4 depends on valua >= 0.2.2 (missing)
Installing https://luarocks.org/valua-0.3-1.src.rock
valua 0.3-1 is now installed in /home/mpeterv/luarocks/here (license: MIT)
```

### DIFF
--- a/src/luarocks/build.lua
+++ b/src/luarocks/build.lua
@@ -332,15 +332,7 @@ function build.build_rockspec(rockspec_file, need_to_fetch, minimal_mode, deps_m
    ok, err = manif.update_manifest(name, version, nil, deps_mode)
    if err then return nil, err end
 
-   local license = ""
-   if rockspec.description and rockspec.description.license then
-      license = ("(license: "..rockspec.description.license..")")
-   end
-
-   local root_dir = path.root_dir(cfg.rocks_dir)
-   util.printout(name.." "..version.." is now built and installed in "..root_dir.." "..license)
-   util.printout()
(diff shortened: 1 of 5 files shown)


hishamhm:
The new output looks good, but can we still have the list of missing dependencies displayed at the top? I think it gives a good preliminary indication when "there's a lot to do".

mpeterv:
Readded the list. Also replaced `missing` label with `not installed` to avoid repetition in the list header.

hishamhm:
Thank you!

PR #552, opened 2016-05-10
Allow passing --force to `luarocks install` in all cases

mpeterv:
When installing an old version of a rock that breaks existing
dependencies, removing previously installed version fails,
and --force is suggested. However, when installing using a rockspec
or a source rock --force is not forwarded to build command.

This really needs a test, plus a test to verify failure without `--force`, but that would require pulling in some new rocks: those currently fetched don't use sufficiently strict dependency constraints.


mpeterv:
Updated to also forward `--force-fast`.

mpeterv:
Considering this a bug fix because of existing `luarocks make` and `luarocks build` behavior (which is untested as well...)

Issue #446, opened 2015-10-29
Custom installation directory?


adnelson:
I'm trying to build packages with luarocks using the nix package manager, which means that I need to be able to have luarocks install a package into a prefix path that will be generated automatically. I'd love a situation in which I could both:

  * Specify any number of paths in which luarocks can find existing dependencies (rather than just $HOME/.luarocks and global paths), and
  * Specify a custom output directory (again, rather than just $HOME/.luarocks and global paths).

Peeking around through the documentation, I haven't seen any way to accomplish this. Although rockspec files can refer to a `PREFIX` variable, but I'm not clear on how this is set.

```
luarocks make my-package.rockspec --prefix=/path/to/custom/prefix
```
And have all of the generated binaries and .lua files be output to the given prefix. Of course, using an environment variable rather than a command-line flag would also be fine.

daurnimator:
You want `--tree`

adnelson:
@daurnimator `--tree` works to set the output directory, thanks! Unfortunately it doesn't seem to work on the input directories. I can't seem to get luarocks to pick up the packages by writing a configfile either. For example, I have the package `luafilesystem` installed in the folder `/nix/store/9gmxg8pfkf103rmzya54w83i52v0n436-lua5.1-filesystem-1.6.3-1`. Here are the files installed in that folder:

```bash
$ (cd /nix/store/9gmxg8pfkf103rmzya54w83i52v0n436-lua5.1-filesystem-1.6.3-1; find .)
.
./lib
./lib/luarocks
./lib/luarocks/rocks
./lib/luarocks/rocks/luafilesystem
./lib/luarocks/rocks/luafilesystem/1.6.3-1
./lib/luarocks/rocks/luafilesystem/1.6.3-1/doc
./lib/luarocks/rocks/luafilesystem/1.6.3-1/doc/us
./lib/luarocks/rocks/luafilesystem/1.6.3-1/doc/us/doc.css
./lib/luarocks/rocks/luafilesystem/1.6.3-1/doc/us/examples.html
./lib/luarocks/rocks/luafilesystem/1.6.3-1/doc/us/index.html
./lib/luarocks/rocks/luafilesystem/1.6.3-1/doc/us/license.html
./lib/luarocks/rocks/luafilesystem/1.6.3-1/doc/us/luafilesystem.png
./lib/luarocks/rocks/luafilesystem/1.6.3-1/doc/us/manual.html
[...]

PR #554, opened 2016-05-16
Implement progress line for downloads using LuaSocket

mpeterv:
It looks like this:

```
Downloading https://luarocks.org/luacheck-0.15.0-1.src.rock
  Got 1.03 MiB (100%) | 20.77 KiB/s
```

Current downloaded size, percentage and download speed are updated on each luasocket step. Redirections are handled. Progress line is shown only for sufficiently large files or slow downloads, thresholds may need some tweaking.

The last commit turns this on by default and removes any mentions of `cfg.show_downloads` from tests.

### DIFF
--- a/src/luarocks/fs/lua.lua
+++ b/src/luarocks/fs/lua.lua
@@ -12,10 +12,10 @@ local util = require("luarocks.util")
 local path = require("luarocks.path")
 
 local socket_ok, zip_ok, unzip_ok, lfs_ok, md5_ok, posix_ok, _
-local http, ftp, lrzip, luazip, lfs, md5, posix
+local socket, ftp, lrzip, luazip, lfs, md5, posix
 
 if cfg.fs_use_modules then
-   socket_ok, http = pcall(require, "socket.http")
+   socket_ok, socket = pcall(require, "socket")
    _, ftp = pcall(require, "socket.ftp")
    zip_ok, lrzip = pcall(require, "luarocks.tools.zip")
    unzip_ok, luazip = pcall(require, "zip"); _G.zip = nil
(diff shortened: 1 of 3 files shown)


src/luarocks/fs/lua.lua
```diff
@@ -544,86 +544,183 @@ end
...
 
 local ltn12 = require("ltn12")
+local http = require("socket.http")
 local luasec_ok, https = pcall(require, "ssl.https")
 
-local redirect_protocols = {
-   http = http,
-   https = luasec_ok and https,
-}
+-- Patch ltn12 source used by LuaSocket to get fixed length
```
hishamhm: My only concern is: is this safe/stable across LuaSocket versions?...

I also noticed I couldn't get the length when I wrote the `spinner` patch (or else I would have made a progress bar). I also noticed that, at least in my tests, the ltn12 `step` wasn't that responsive (that's why I ended up with just a generic "something's happening" spinner).

mpeterv: It is very stable (underlying code not changed since 2004). Tested it with the oldest version of Luasocket available on Luarocks (2.0.1-2), it works.

What do you mean by being not responsive? It updates too rarely?

daurnimator:
I feel like this is overly hacky for luasocket. 

This sort of thing would be much nicer with #446

mpeterv:
Updated: simplified thresholds: instead of using speed and size simply show the progress line if download isn't complete in some time windows/

src/luarocks/fs/lua.lua
hishamhm: > What do you mean by being not responsive? It updates too rarely?

Yes, in my experience, with the dots in step(), most of the delay is caused by the initial setup of the connection, the actual downloads of the payload run fast/are short, so there's only one or two `step` updates after the initial delay for typical small rocks. My copas-based spinner was more meant as a psychological "hang in there" than as a way to measure progress.

But well, since I committed the dots at one point I suppose there were situations where they were useful.

mpeterv: There are a few large rocks (1MiB or more), like sailor and luacheck (the fancy documentation fonts are probably worth it), for them this progress line is really nice. But I agree that the connection setup is typically the longest part. Perhaps combining progress line with a spinner is optimal, and this PR should be a work in progress for now.

PR #558, opened 2016-05-22
Don't put array items on the same line when persisting

mpeterv:
Rockspec fields with array values (e.g. dependencies) are typically written with each item on its own line. This change makes `luarocks new-version` follow this style, too. This also affects they way manifests are saved.

### DIFF
--- a/src/luarocks/persist.lua
+++ b/src/luarocks/persist.lua
@@ -147,7 +147,7 @@ write_table = function(out, tbl, level, field_order)
       end
 
       write_value(out, v, level, sub_order)
-      if type(k) == "number" then
+      if type(v) == "number" then
          sep = ", "
          indent = false
       else


hishamhm:
IIRC, this style was used because of manifest files. Since manifest files are zipped now, I _suppose_ it makes no difference...

mpeterv:
@hishamhm why? This doesn't change manifest size, both separators have length 2

hishamhm:
deeply indented subtables produce lots of spaces

mpeterv:
@hishamhm would you like an option to retain old behaviour when saving manifests or is this OK?

src/luarocks/persist.lua
```diff
@@ -147,17 +144,11 @@ write_table = function(out, tbl, level, field_order)
       end
 
       write_value(out, v, level, sub_order)
-      if type(k) == "number" then
```
hishamhm: I was looking at the manifest tables. I suggest this compromise solution: `if type(v) == "number"`. This will catch the version tables correctly.

mpeterv: Is that for parsed version tables in `dependencies` table? That could work. Although isn't that table redundant? `repository` table contains dependencies in string format already...

mpeterv: Updated the PR.

hishamhm: it is redundant, but I wanted to store the pre-parsed info for speeding things up for `luarocks.loader` (though I don't remember if I measured it, and if I did, it was back in 2007 or so when I first wrote this code). The idea was that `luarocks.loader` should work as little as possible to have minimal impact on `require`.

PR #561, opened 2016-05-22
Move common `luarocks.fs.{unix,win32}.tools` functions into a new module

mpeterv:
Some functions in `luarocks.fs.{unix,win32}.tools` have same implementations. Some of them use different implementations of local platform specific helper functions. This PR creates exposes these functions and moves common functions into a new `luarocks.fs.tools` module.


hishamhm:
Wondering if I can trust the AppVeyor build as an indicator that nothing broke on Windows... :)
Should I merge this right away or do I make a new release first?

mpeterv:
@hishamhm I've tested this on Windows, a single install commands gave pretty good coverage for the changed commands, so it should be good.

PR #562, opened 2016-05-23
Write rockspec infer scm url

mpeterv:
When using write-rockspec with a local path, check if it's a git or hg repo and infer URL from origin location. Also update homepage detection to use this and support more protocols and domains. Now `luarocks write-rockspec` produces a working scm rockspec out-of-the-box for rocks using git or hg :tada:


hishamhm:
Awesome! Thanks!

PR #563, opened 2016-05-25
luarocks new-version: make rockspec argument optional

mpeterv:
Like with `make`, it makes sense for `new-version` to default to latest available rockspec, picked same way `make` does it.


hishamhm:
Nice! Thank you!

PR #567, opened 2016-06-03
Refactor cli to avoid double args parsing

mpeterv:
Instead of 'run' function command modules now have 'command' function that accepts already parsed flags. Code mutating original args in 'luarocks.command_line' and 'util.forward_flags' are no longer needed.

There was a comment saying that the old 'run' interface, although undocumented, is used by some people. It may be irrelevant now but I've added an utility function wrapping the new interface to provide 'run'. It's called on all command modules.

### DIFF
--- a/src/luarocks/add.lua
+++ b/src/luarocks/add.lua
@@ -12,6 +12,7 @@ local index = require("luarocks.index")
 local fs = require("luarocks.fs")
 local cache = require("luarocks.cache")
 
+util.add_run_function(add)
 add.help_summary = "Add a rock or rockspec to a rocks server."
 add.help_arguments = "[--server=<server>] [--no-refresh] {<rockspec>|<rock>...}"
 add.help = [[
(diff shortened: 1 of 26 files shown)


mpeterv:
@hishamhm the comment about 'run' interface being used by other programs is four years old, do you know if it's still true?

Alloyed:
afaict it is a reference to
https://github.com/stevedonovan/LuaRocksTools/tree/master/api
which at least on github was last updated 5 years ago, and had major bugs anyways when I tried to use it verbatim.

I use a patched variant to automate luarocks commands in loverocks ([example][2]), but I am perfectly okay with my abuse of internal APIs breaking. The proposed patch seems easy enough to feature-detect.

[2]: https://github.com/Alloyed/loverocks/blob/master/loverocks/commands/purge.lua#L18

hishamhm:
@mpeterv @Alloyed Sounds reasonable! I think we could get rid of the run() function in the LuaRocks 3 branch.

mpeterv:
@hishamhm done!

hishamhm:
awesome, thank you!

PR #574, opened 2016-06-22
Improve error on archive unpack error

mpeterv:
Currently fetch ignores archive unpack errors and simply checks if expected source directory exists. This may be a workaround to support an undocumented feature that allows to use an url pointing directly to a `.lua` or `.c` file, but `fs.unpack_archive` already ignores such files.

With this PR unpack errors are detected earlier, and use a different message for unrecognized archive extension, which is a common error that happens when wrongly using `https://` in place of `git+https://`:

`Directory <name> not found inside archive <name>.git`

becomes

`Couldn't extract archive <name>.git: unrecognized extension`.

### DIFF
--- a/src/luarocks/fs/unix/tools.lua
+++ b/src/luarocks/fs/unix/tools.lua
@@ -192,8 +192,7 @@ function tools.unpack_archive(archive)
       -- Ignore .lua and .c files; they don't need to be extracted.
       return true
    else
-      local ext = archive:match(".*(%..*)")
-      return false, "Unrecognized filename extension "..(ext or "")
+      return false, "Couldn't extract archive "..archive..": unrecognized filename extension"
    end
    if not ok then
       return false, "Failed extracting "..archive
--- a/src/luarocks/fetch.lua
+++ b/src/luarocks/fetch.lua
@@ -344,7 +344,8 @@ function fetch.get_sources(rockspec, extract, dest_dir)
    if extract then
       local ok, err = fs.change_dir(store_dir)
       if not ok then return nil, err end
-      fs.unpack_archive(rockspec.source.file)
+      ok, err = fs.unpack_archive(rockspec.source.file)
+      if not ok then return nil, err end
       if not fs.exists(rockspec.source.dir) then
          return nil, "Directory "..rockspec.source.dir.." not found inside archive "..rockspec.source.file, "source.dir", source_file, store_dir
       end
--- a/src/luarocks/fs/win32/tools.lua
+++ b/src/luarocks/fs/win32/tools.lua
@@ -204,8 +204,7 @@ function tools.unpack_archive(archive)
       -- Ignore .lua and .c files; they don't need to be extracted.
       return true
    else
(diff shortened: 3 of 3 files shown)


mpeterv:
Would be nice if luarocks could also say something like `The rockspec seems be using https:// protocol instead of git+https:// or hg+https://` in this common case, not sure where to put that. Or maybe we could just improve the rockspec creation guide focusing more on scm-based ones.

mpeterv:
The tests fail because find_base_dir can be called on urls using non-basic protocols, and unpack fails on them.

src/luarocks/fs/unix/tools.lua
```diff
@@ -192,8 +192,7 @@ function tools.unpack_archive(archive)
       -- Ignore .lua and .c files; they don't need to be extracted.
       return true
    else
-      local ext = archive:match(".*(%..*)")
-      return false, "Unrecognized filename extension "..(ext or "")
+      return false, "Couldn't extract archive "..archive..": unrecognized extension"
```
hishamhm: I think it should keep saying "filename extension"; "extension" is vague.

hishamhm:
Also, which tests fail?

mpeterv:
>>Or maybe we could just improve the rockspec creation guide focusing more on scm-based ones.
>
>But the guide already points to git:// first, no?
https://github.com/keplerproject/luarocks/wiki/Creating-a-rock

Right, I looked at rockspec format page. It's really just nitpicking though.

>The right thing to do, in luarocks-3, would be to detect .git URLs. Do any other SCMs use filename extension like that to identify an SCM URL or is this just a Git thing? The fetch system could be adjusted to try detection by filename extension first, then protocol.

But `.git` extension is optional, so it shouldn't be possible to detect all git urls.

> Also, which tests fail?

new-version and write-rockspec tests that run corresponding commands on scm rockspecs, I've fixed it.

src/luarocks/fs/unix/tools.lua
mpeterv: Fixed.

hishamhm:
> But .git extension is optional, so it shouldn't be possible to detect all git urls.

Sure, a plain "https://github.com/user/project" can't be detected, but I notice that users assume that the .git extension is a way of telling LuaRocks that the URL is a Git repo.

mpeterv:
@hishamhm that could work, but what would the standard way of doing things be, 'git+https://foo.com/bar' or 'https://foo.com/bar.git'? Or do you want to remove the first one?

hishamhm:
I'd keep them both. I guess Git users would naturally grativate towards the latter (it's the default URL Github gives you for copy-pasting), but we'd still have to handle foo:// protocols for the other SCMs.

mpeterv:
Ok, that could be helpful. We'll have two ways for doing one thing, but that's fine as one of them is more generic than the other. Sorry for derailing this with my suggestions, can this PR be merged?

hishamhm:
Sure!

PR #577, opened 2016-06-30
Core modules reorganization.

hishamhm:
Isolate all modules used by luarocks.loader in a luarocks.core.* namespace. Core modules can only require other core modules, and they only use require() at the top-level chunk. In other words, after luarocks.loader is setup, package.path can be altered at will and it should not affect the ability of luarocks.loader to run, and the luarocks.loader modules should not interfere with modules loaded later.

This was motivated by @robooo's GSoC work on porting the LuaRocks test suite to Lua using Busted. Busted itself runs using luarocks.loader, and we need to ensure that the modules loaded by the tests do not conflict with the ones used by luarocks.loader.

A circular dependency between luarocks.manif and luarocks.search was removed by creating a separate luarocks.manif.writer module.

Also, luarocks-admin commands were moved to a luarocks.admin.* namespace.


test/testing.sh
```diff
@@ -577,7 +577,6 @@ fail_config_system_config() { rm -f "$testing_lrprefix/etc/luarocks/config.lua";
 test_config_system_config() { mkdir -p "$testing_lrprefix/etc/luarocks"; touch "$testing_lrprefix/etc/luarocks/config.lua"; $luarocks config --system-config; err=$?; rm -f "$testing_lrprefix/etc/luarocks/config.lua"; return $err; }
 fail_config_system_config_invalid() { mkdir -p "$testing_lrprefix/etc/luarocks"; echo "if if if" > "$testing_lrprefix/etc/luarocks/config.lua"; $luarocks config --system-config; err=$?; rm -f "$testing_lrprefix/etc/luarocks/config.lua"; return $err; }
 test_config_user_config() { $luarocks config --user-config; }
-fail_config_user_config() { LUAROCKS_CONFIG="/missing_file.lua" $luarocks config --user-config; }
```
mpeterv: Why is this test removed? For some reason can't see it in the travis log on master either.

mpeterv:
Dependencies of `luarocks.loader` before and after this PR: http://imgur.com/a/Krewe

mpeterv:
I'll look through all the changes more thoroughly but so far it looks good. One thing I don't like is the unused `require` locals. I understand the purpose but IMO it's better to add a readme to the core subdirectory explaining why `require` shouldn't be called after the top part.

test/testing.sh
hishamhm: I think there was a discussion about this, but I don't remember where. But the situation right now is that this test is not working because the behavior of config loading in luarocks.cfg was changed a while ago (when @Tieske made it smarter). 

The end result is that now, a missing file in $LUAROCKS_CONFIG only causes a warning, and then luarocks.cfg proceeds trying to load the defaults. On my machine, where I do have an existing ~/.luarocks, this ends up returning success, and the test fails. 

I assume that on Travis, where ~/.luarocks doesn't exist, this fails with a file-not-found status, so it gets the expected fail status, but for the wrong reason. In any case, the test, written as is, is misleading.

I deleted it here because I stumbled upon this (again) when running the test suite locally to check these changes. (My approach in this patch was "fix things until the testsuite and luacheck-in-dit seem happy.)

hishamhm:
The unused `require` locals annoy luacheck (and therefore look ugly in Dit), but I wanted an enforced runtime check to verify that this rule isn't being violated.

And thank you for the graphs! It looks reassuring. Also, since the luarocks.core.* modules are in general smaller, we get the added benefit that luarocks.loader is now a bit lighter-weight. :)

mpeterv:
> The unused require locals annoy luacheck (and therefore look ugly in Dit), but I wanted an enforced runtime check to verify that this rule isn't being violated.

All right, maybe it's clear enough that `something_unused = nil` means "don't use `something_unused`. BTW luacheck can be silenced with `-- luacheck: ignore` on the line with assignment. (another BTW, maybe we could add `.luacheckrc` to the root so that `luacheck src` produces no warnings? Unfortunately dit can't use the config but plugins for other editors that use luacheck from command line can. Sorry for derail.)

hishamhm:
> BTW, maybe we could add .luacheckrc to the root so that luacheck src produces no warnings? 

Sure!

> Unfortunately dit can't use the config but plugins for other editors that use luacheck from command line can. 

I'll look into supporting .luacheckrc in Dit during my next round of productive procrastination. :)

PR #583, opened 2016-07-08
Minor changes and fixes for new test suite

mpeterv:
This is just a heap of commits adjusting the new test suite a little. Not sure if this can be merged completely, will probably refine and cherry-pick commits one by one.

So far I like the new test suite a lot. Implementing it using busted allows us to later improve it further:

* Make it Windows-compatible
* Try to reduce output noise for successful tests. Typically in busted specs tests don't output anything at all if they pass. On failure, assertion that failed should provide enough information for debugging. This is definitely not true for `assert.is_true` and `assert.is_false` assertions that are used when checking command success and failure (it just prints: `expected true, got false`). If we'll go this way we'll need custom assertions that execute command and print it and its output on unexpected result. A collections of such assertions would be valuable for testing other command-line programs, might want to make that a library.


hishamhm:
Wow, thanks Peter! That's a thorough review! I'm glad you like the new test suite, @robooo has been putting a lot of work on it (and it's awesome that we already got it to a mergeable state and he's still basically halfway through his GSoC!)

If both you and @robooo are fine with all these changes, I think they are good to merge.

PR #591, opened 2016-07-15
update 7zip from the official page and mkdir and libiconv2.dll from GOW*

Stanzilla:
* https://github.com/bmatzelle/gow

Looks like the binaries were not that old after all, copying them over only the two files had changes.


hishamhm:
Thanks for the PR! But it looks like these broke the build on Windows, according to the Appveyor run. :(

Issue #602, opened 2016-08-16
Feature request: install.bat could autodetect /LV


hishamhm:
While looking at [this log](https://ci.appveyor.com/project/robooo/luarocks/build/2.2.1.304-test/job/9a7dx0uoywxt9ytf) it seems that install.bat succeeded finding a Lua executable (`/LUA` was given), but then it proceeded to try to find a Lua 5.1 library, when the Lua executable found was actually Lua 5.2. The installer could run the Lua executable after finding it (something like `lua -e 'print((_VERSION:gsub("Lua ","")))'`) to extract its version, and use it to auto-detect the version when `/LV` is not given (and possibly, to check that it is consistent, when `/LV` is given).

mpeterv:
For other systems `./configure` already does that, right? Should be easy to replicate.

PR #605, opened 2016-08-21
Detect Lua version in install.bat

mpeterv:
Implements #602. The first four commits are almost pure refactoring (visible changes should be that there will be no trailing backslashes in autodetected LUA_BINDIR, LUA_INCDIR and LUA_LIBDIR), but I'd like to have them reviewed since the code may not be very well covered with tests right now. It may be useful to test @robooo's Windows testsuite with this before merging.

### DIFF
--- a/install.bat
+++ b/install.bat
@@ -287,56 +287,32 @@ end
 -- ***********************************************************
 -- Detect Lua
 -- ***********************************************************
-local function look_for_interpreter (directory)
+local function look_for_interpreter(directory)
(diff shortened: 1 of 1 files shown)


mpeterv:
I've pulled this on top of #606 and removed `/LV` from LuaRocks installation command, it works: https://ci.appveyor.com/project/mpeterv/luarocks/build/2.2.1.1-test

robooo:
great, good job

PR #608, opened 2016-08-22
Run Windows tests with MinGW

hishamhm:
Adds MinGW to the test matrix in Appveyor.
Note commit af41c31 — it's a change in the LuaRocks defaults, but I think it makes sense because Makefile.win is a leftover from Kepler days, and those are usually nmake makefiles for MSVC; any rockspecs depending on that default value were most likely broken with MinGW.

This PR also removes "/LV" from the testsuite, thanks to #605.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,23 +7,23 @@ environment:
 
   matrix:
   - LUA: "lua 5.1"
-    COMPILER: "msvc"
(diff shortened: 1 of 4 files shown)


mpeterv:
Looks good to me, although probably need to revert verbose call in a test if it now passes?

PR #617, opened 2016-09-29
Update deployment target for MAC OS X 10.12

o-lim:
In of MAC OSX v10.12, support for `clock_gettime` was added. This causes builds with luarocks to fail on MAC OS X 10.12 for C modules that use `clock_gettime` since the `MAC_OSX_VERSION_MIN_REQUIRED` macro does not reflect version 10.12, but 10.8 instead. This is due to luarocks setting the deployment target to 10.8 for MAC OS X version 10.10 and later. Hence, for MAC OS X versions 10.12 and later, the deployment target should be at least 10.12.


hishamhm:
This patch is not needed because LuaRocks 3 will allow setting the deployment target version explicitly in rocks that require a particular version. Support is already committed in the `luarocks-3` branch.

PR #623, opened 2016-10-06
[WIP] Add source.revision

### LINKED ISSUE #569: Feature request: "revision" field in rockspecs
example:
```
source = {
  url = "git://my/repo",
  revision = "86d59e66d0c12656ed34d4f4a9bab06ac7775f77"
}
```
In this context revision means an atomic reference to a repo's state, e.g. commit hashes in git and changeset IDs in Mercurial.

This can be useful for when you don't directly control a source repo to add tags to it, or if you want more specific rockspecs than you want tags. It can also be used as verification for tags, the same way source.md5 is used.

Implementation-wise, it doesn't look like git lets you clone a specific commit, so instead you need to pick the closest branch/tag and then `git checkout` from there. Mercurial's `--rev` flag also does changesets in addition to branches/tags so it just be a matter of also passing the changeset id in.

Alloyed:
Resolves #569.

In hg, this is implemented by passing revision into the clone command (`hg clone myrepo --rev myrev`).

In git, this is implemented by performing a full clone (`git clone myrepo`) and then doing `git checkout myrev`.

Originally I planned to have a special case that compares `source.tag` and `source.revision` when both are defined. This would double check the tag's commit-id, but from a security perspective that just moves the trust from the repository host to the rockspec host. Git already supports tag/commit signing with GPG, which are better ways to verify trust.

todo:
- [x] input validation (revisions should _only_ be full commit/changeset IDs)
- [x] git optimizations

tests:
- [x] error on revision in rockspec < 3.0
- [ ] error on revision in unsupported URL type (other schemes don't do this it looks like)
- [x] simple revision in git
- [x] simple revision in hg
- [x] revision + tag
- [x] revision + branch (with `--single-branch` this can be special-cased to avoid full clone)
- [x] revision + tag + branch

### DIFF
--- a/src/luarocks/fetch.lua
+++ b/src/luarocks/fetch.lua
@@ -237,6 +237,10 @@ function fetch.load_local_rockspec(filename, quick)
    if rockspec.source.cvs_module then rockspec.source.module = rockspec.source.cvs_module end
    if rockspec.source.cvs_tag then rockspec.source.tag = rockspec.source.cvs_tag end
 
+   if rockspec.source.revision and not deps.format_is_at_least("3.0") then
+      return nil, "source.revision is only supported in rockspec format 3.0 or higher"
+   end
+
    local name_version = rockspec.package:lower() .. "-" .. rockspec.version
    if basename ~= "rockspec" and basename ~= name_version .. ".rockspec" then
       return nil, "Inconsistency between rockspec filename ("..basename..") and its contents ("..name_version..".rockspec)."
--- a/src/luarocks/fetch.lua
+++ b/src/luarocks/fetch.lua
@@ -237,10 +237,6 @@ function fetch.load_local_rockspec(filename, quick)
    if rockspec.source.cvs_module then rockspec.source.module = rockspec.source.cvs_module end
    if rockspec.source.cvs_tag then rockspec.source.tag = rockspec.source.cvs_tag end
 
-   if rockspec.source.revision and not deps.format_is_at_least("3.0") then
-      return nil, "source.revision is only supported in rockspec format 3.0 or higher"
-   end
-
    local name_version = rockspec.package:lower() .. "-" .. rockspec.version
    if basename ~= "rockspec" and basename ~= name_version .. ".rockspec" then
       return nil, "Inconsistency between rockspec filename ("..basename..") and its contents ("..name_version..".rockspec)."
--- a/src/luarocks/fetch/git.lua
+++ b/src/luarocks/fetch/git.lua
@@ -57,7 +57,7 @@ end
 -- @return (string, string) or (nil, string): The absolute pathname of
 -- the fetched source tarball and the temporary directory created to
 -- store it; or nil and an error message.
-function git.get_sources(rockspec, extract, dest_dir, depth)
+function git.get_sources(rockspec, extract, dest_dir)
    assert(type(rockspec) == "table")
    assert(type(dest_dir) == "string" or not dest_dir)
 
@@ -85,13 +85,21 @@ function git.get_sources(rockspec, extract, dest_dir, depth)
    store_dir = fs.absolute_name(store_dir)
    local ok, err = fs.change_dir(store_dir)
    if not ok then return nil, err end
-   
-   -- there's no way to get depth from a commit id, unfortunately
-   if rockspec.source.revision then 
-      depth = "--"
+
+   -- shallow clone by default
+   local depth = "--depth=1"
+   if rockspec.source.revision then
+      if rockspec.source.branch then
+         -- If the rockspec specifies a branch, then clone only that branch
+         depth = "--single-branch"
+      else
+         -- otherwise, we need a full clone to make sure we fetch all available
+         -- revisions
+         depth = "--"
+      end
    end
 
-   local command = {fs.Q(git_cmd), "clone", depth or "--depth=1", rockspec.source.url, module}
+   local command = {fs.Q(git_cmd), "clone", depth, rockspec.source.url, module}
    local tag_or_branch = rockspec.source.tag or rockspec.source.branch
    -- If the tag or branch is explicitly set to "master" in the rockspec, then
    -- we can avoid passing it to Git since it's the default.
--- a/src/luarocks/fetch/git.lua
+++ b/src/luarocks/fetch/git.lua
@@ -32,6 +32,7 @@ local function git_is_at_least(git_cmd, version)
    return git_version(git_cmd) >= vers.parse_version(version)
 end
 
+
 --- Git >= 1.7.10 can clone a branch **or tag**, < 1.7.10 by branch only. We
 -- need to know this in order to build the appropriate command; if we can't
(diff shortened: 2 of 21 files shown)


src/luarocks/type_check.lua
```diff
@@ -58,6 +58,7 @@ local rockspec_types = {
       dir = string_1,
       tag = string_1,
       branch = string_1,
+      revision = string_1,
```
hishamhm: You should use `revision = { _type = "string", _version = "3.0" }` here (or you can even create a `string_3` variable)

src/luarocks/fetch.lua
```diff
@@ -235,6 +235,10 @@ function fetch.load_local_rockspec(filename, quick)
    if rockspec.source.cvs_module then rockspec.source.module = rockspec.source.cvs_module end
    if rockspec.source.cvs_tag then rockspec.source.tag = rockspec.source.cvs_tag end
 
+   if rockspec.source.revision and not deps.format_is_at_least("3.0") then
```
hishamhm: With the version checking for the `revision` field in the type checker, you don't need to version-check again here.

hishamhm:
Yay, thanks for taking on this feature! I added some comments.

Alloyed:
Thanks for the review, I'm working on a commit that resolves it and adds a few of the simpler tests right now.

Something else: I noticed that luarocks already calls the "-1" part of a rockspec version a "rockspec revision", so I'm worried about the two getting confused for each other. Wikipedia suggests [changeset ](https://en.wikipedia.org/wiki/Changeset) as a possible alternate name I'd be okay with.

hishamhm:
* I think `source.revision` won't be ambiguous (that's what the table namespacing is for, after all) and is more obvious than "changeset" — every time I saw people asking for this feature they said "I'd like my rockspec to point to a particular revision", not "changeset". (I'm trying to avoid here a situation where we pick a technically correct name that ends up being more cumbersome in regular use, as it happened when we picked "scm" to refer to in-development rocks).
* Do we need to force all 40 characters in the revision field? A lot more often I see people using much shorter substrings of revisions and that works fine in practice. Again, I'm thinking about convenience here. Those who want the safety of the full hash can still have it. But it's customary for Git-based tools to accept substrings of the hash.
* As for the error message, how about adding an _expecting field to be used for composing the error message? As in `_expecting = "a revision identifier"`

Alloyed:
1. Understood on the name/typechecking

2. The CLI tools will allow up to the shortest unique substring, but the issue is that this leads to a potential time-bomb: what if another commit made after the fact shares that substring? If the original commit is still available a rockspec that used to work will break, or if (say, a bad actor) came along and replaced the original git history, you could get a completely different commit.
If a rockspec is temporary this isn't a big deal, so maybe this is an argument to be more lenient here, and more strict in luarocks.org.

hishamhm:
> If a rockspec is temporary this isn't a big deal, so maybe this is an argument to be more lenient here, and more strict in luarocks.org.

This is not a big practical problem at luarocks.org because `luarocks upload` also stores a .src.rock file which contains a copy of the sources. `luarocks build` uses the .src.rock by default instead of fetching from git.

mpeterv:
>luarocks build uses the .src.rock by default instead of fetching from git.

So does luarocks install

Alloyed:
Something I noticed: right now, if you use luarocks with an http/https/ssh url, luarocks is forced to do a full clone. Do you remember the reasoning behind it? At least on my machine doing a shallow clone works fine for all three types.

Anyways, an in-words description of how cloning works in this PR:
1. If you don't need a specific revision, you get a shallow clone (git clone --depth=1). this includes normal branches and tags.
2. If you do need a specific revision, and you specify a branch, you get only the commits on that branch (git clone --single-branch)
3. If you specify a revision by itself you get the entire repo. (git clone --)

src/luarocks/fetch/git.lua
```diff
@@ -86,7 +92,20 @@ function git.get_sources(rockspec, extract, dest_dir, depth)
...
+   -- shallow clone by default
+   local depth = "--depth=1"
+   if rockspec.source.revision then
+      if rockspec.source.branch and git_supports_single_branch(git_cmd) then
+         -- If the rockspec specifies a branch, then clone only that branch
+         depth = "--single-branch"
+      else
+         -- otherwise, we need a full clone to make sure we fetch all available
+         -- revisions
+         depth = "--"
```
daurnimator: This seems prone to later issues: `--` signifies "things after this point are arguments not options".

src/luarocks/fetch/git.lua
```diff
@@ -86,7 +92,20 @@ function git.get_sources(rockspec, extract, dest_dir, depth)
...
+      end
+   end
+
+   local command = {fs.Q(git_cmd), "clone", depth, rockspec.source.url, module}
```
daurnimator: Do url and module need to be quoted?

src/luarocks/fetch/git.lua
```diff
@@ -96,16 +115,20 @@ function git.get_sources(rockspec, extract, dest_dir, depth)
...
    end
    ok, err = fs.change_dir(module)
    if not ok then return nil, err end
-   if tag_or_branch and not git_can_clone_by_tag() then
-      if not fs.execute(fs.Q(git_cmd), "checkout", tag_or_branch) then
-         return nil, 'Failed to check out the "' .. tag_or_branch ..'" tag or branch.'
+
+   local committish = rockspec.source.revision or tag_or_branch
+   if committish then
+      if not fs.execute(fs.Q(git_cmd), "checkout", committish) then
```
daurnimator: Quote commitish (and/or use `--` before it)

src/luarocks/fetch/hg.lua
```diff
@@ -32,9 +32,15 @@ function hg.get_sources(rockspec, extract, dest_dir)
...
 
    local command = {hg_cmd, "clone", url, module}
-   local tag_or_branch = rockspec.source.tag or rockspec.source.branch
-   if tag_or_branch then
-      command = {hg_cmd, "clone", "--rev", tag_or_branch, url, module}
+
+   -- TODO: if revision is specified and so are tag/branch, then validate to
+   -- make sure tag/branch matches revision
+   local rev = rockspec.source.revision
+   rev = rev or rockspec.source.tag
```
daurnimator: No need to keep assigning, just start the line from `or`

src/luarocks/fetch/hg.lua
```diff
@@ -32,9 +32,15 @@ function hg.get_sources(rockspec, extract, dest_dir)
...
+   rev = rev or rockspec.source.branch
+
+   if rev then
+      command = {hg_cmd, "clone", "--rev", rev, url, module}
```
daurnimator: Why did git_cmd need `fs.Q` but hg_cmd doesn't?

hishamhm:
This was auto-closed by accident when the (now merged) `luarocks-3` branch was deleted. Sorry!

PR #624, opened 2016-10-06
Don't use coroutines in util.sortedpairs

mpeterv:
util.sortedpairs is used in luarocks loader since @6dc745a.
Openresty does not like coroutines being used from inside `require`,
resulting in "attempt to yield across C-call boundary" error.
New version of util.sortedpairs uses a prepared array of ordered keys
instead of coroutines.

Also added some tests for it.


hishamhm:
Thank you! (And thanks for adding tests!! :D )

PR #626, opened 2016-10-10
Fix of openssl location on TravisCI OSX build

robooo:
Travis on OS X platform couldn't locate openssl headers, now the path needs to be specified in LR command.


daurnimator:
Does travis come with openssl already installed via brew?

hishamhm:
@robooo Thank you! Merging.

PR #628, opened 2016-10-12
Fix manif.find_{current,next}_provider returning "untracked" incorrectly

mpeterv:
`find_providers` function used by `manif.find_current_provider`
and `manif.find_next_provider` needs relative path from a directory
an installed file is deployed in (bin dir, lua dir, or lib dir) to the file.
It then uses that path as key in manifest, converting it to
module name beforehand for lua files and lib files.

It happened to leave a leading slash in this relative path for
lua and lib files. `path.path_to_module` has a workaround stripping
leading dots caused by leading slashes. However, if the file
doesn't have `.lua` extension, slashes are not converted to dots
and the workaround doesn't trigger.

The issue results in files falsely considered "untracked" and backed-up
when reinstalling a different version of a rock,
see sailorproject/sailor#138.

The fix is to use correct relative paths without leading slashes.

### DIFF
--- a/src/luarocks/manif.lua
+++ b/src/luarocks/manif.lua
@@ -464,6 +464,11 @@ function manif.zip_manifests()
    end
 end
 
+local function relative_path(from_dir, to_file)
+   -- It is assumed that `from_dir` is prefix of `to_file`.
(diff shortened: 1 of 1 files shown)


mpeterv:
A partial fix that only worked for commands was committed in 2009! b92f5aa

PR #629, opened 2016-10-14
Relax Lua version detection to support Ravi

mpeterv:
Ravi has `"Ravi 5.3"` as `_VERSION`. Don't use `_VERSION:sub(5)` to get Lua
version, match `" (5%.[123])$"` instead.

(with this patch I got sailor working under ravi very easily using experimental hererocks support: https://github.com/mpeterv/hererocks/tree/ravi :smile: )

### DIFF
--- a/configure
+++ b/configure
@@ -226,8 +226,8 @@ then
 fi
 
 detect_lua_version() {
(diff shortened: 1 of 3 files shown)


hishamhm:
Hopefully the author of Ravi will keep such a good level of backward compatibility :)

Issue #302, opened 2015-01-20
Upgrading a rock with bin scripts deletes them (Windows only)


ignacio:
For instance, luacheck installs a bin script (luacheck.bat). I had version 0.7.3 installed. When I upgraded to version 0.8.0, the script got deleted.

The script was located in `C:\luarocks\systree\bin\luacheck.bat`

~~~~
D:\>luarocks install luacheck
Installing https://rocks.moonscript.org/luacheck-0.8.0-1.src.rock...
Using https://rocks.moonscript.org/luacheck-0.8.0-1.src.rock... switching to 'build' mode
Updating manifest for c:\luarocks\systree/lib/luarocks/rocks
luacheck 0.8.0-1 is now built and installed in c:\luarocks\systree (license: MIT/X11)

Checking stability of dependencies on the absence of
luacheck 0.7.3-1...

Removing luacheck 0.7.3-1...
c:\luarocks\2.2\tools\cp: cannot stat `c:\\luarocks\\systree\\bin\\luacheck_0_7_3_1-luacheck.bat': No such file or directory
c:\luarocks\2.2\tools\cp: cannot stat `c:\\luarocks\\systree\\share\\lua\\5.1\\luacheck_0_7_3_1-luacheck.lua': No such file or directory
Removal successful.
~~~~

I uploaded the result of `luarocks install luacheck --verbose` here:
https://www.dropbox.com/s/rpyl02lbwk3wqw3/luarocks_issue.zip?dl=0

mpeterv:
luacheck 0.7.3 doesn't contain luacheck.bat script. luacheck 0.8.0 does, but it's not in the rockspec.

On Linux I only got the second error message when upgrading from 0.7.3 to 0.10.0 (`cannot stat /usr/local/share/lua/5.3/luacheck_0_7_3_1-luacheck.lua`). Conversely, downgrading from 0.10.0 to 0.7.3 produces `cannot stat /usr/local/share/lua/5.3/luacheck_0_7_3_1-luacheck/init.lua`.

[...]

PR #631, opened 2016-10-19
Fix conflict resolution

mpeterv:
When deploying or deleting files, resolve conflicts purely
based on module names and command names, not file names.
Also, don't assume that in case of a conflict both packages have the
same file providing the module or command; it can be false due to binary
wrappers and `path_to_module("mod/init.lua")` == `path_to_module("mod.lua"). This fixes
issues like in #302.

Also added a test for this, it fails on master.


mpeterv:
Added suffix checks everywhere but not in file backing up... (E.g. on Windows installing a new version of a wrapped script will move previous one out of the way if it's tracked (unlike on master), but won't back up an untracked version)

PR #638, opened 2016-10-28
Remove packages without manifest rebuild

mpeterv:
Implements a function that removes a version from manifest without rebuilding it. Also some minor refactoring and an optimization: exit from manifest updating functions early if it was missing and rebuilt.


mpeterv:
@hishamhm should I split out and apply commits with refactoring or is this fine as it is?

hishamhm:
@mpeterv Thank you! When I checked yestered the link the CI was still running, so I didn't comment. Let's merge this as it is.

PR #640, opened 2016-10-29
luarocks.cmd.* hierarchy

hishamhm:
Move command drivers to `luarocks.cmd.*` and `luarocks.admin.cmd.*`

Each module in `cmd` corresponds to a command and it only export the `command` driver function. Functions that are shared by multiple commands remain in their top-level modules.


src/luarocks/manif/writer.lua
```diff
@@ -286,73 +313,105 @@ function writer.make_manifest(repo, deps_mode, remote)
...
-         return nil, err
-      end
-      manifest, err = manif.load_manifest(repo)
-      if not manifest then
-         return nil, err
-      end
+      -- Manifest built by `manif.make_manifest` should already
+      -- include information about given name and version,
+      -- no need to update it.
+      return manif.make_manifest(rocks_dir, deps_mode)
```
mpeterv: This should refer to `writer.make_manifest` now, same for `remove_from_manifest`

hishamhm: Thanks!!

PR #642, opened 2016-11-01
Add luaenv support in configure script.

xpol:


hishamhm:
This is something that the environment manager should handle, not LuaRocks. Both [hererocks](http://github.com/mpeterv/hererocks) and [luaver](http://github.com/DhavalKapil/luaver) handle `configure` flags by themselves, `luaenv` should do the same. We shouldn't add explicit support for a specific environment manager in LuaRocks.

PR #644, opened 2016-11-01
Fix conflict resolution on deploy/remove w.r.t. C/Lua modules with same name

### LINKED ISSUE #424: luarocks show prints wrong directory
I installed a bunch of files whose name didn't end in `.lua` as Lua modules with `install.lua`. They are installed in the relevant `share` directory. But `luarocks show` prints out their paths under the corresponding `lib` directory, even though there are no files there.

For example: `lua = { ["foo.bar"] = "bar" }` in rockspec, file is installed under `share/5.3/package/bar`, but `luarocks show` says it's under `lib/5.3/package/bar`.

mpeterv:
See second commit message. This doesn't fix #424 since that would require moving some functions into manif_core (rock manifests are required for precise `path.which`), will be done in LR3. Also added a test, failing on master (on v2.4.1, too): https://ci.appveyor.com/project/mpeterv/luarocks/build/2.2.1.33-test/job/04t0s5ss4gnyf0ne

(A lot of changes but I hope that this makes the way things work clearer not only for me. Also highlights issues with manifest format to be fixed in LR3)

### DIFF
--- a/spec/make_spec.lua
+++ b/spec/make_spec.lua
@@ -2,11 +2,11 @@ local test_env = require("test/test_environment")
 local lfs = require("lfs")
 local run = test_env.run
 local testing_paths = test_env.testing_paths
+local env_variables = test_env.env_variables
 
 test_env.unload_luarocks()
 
 local extra_rocks = {
-   "lpeg-1.0.0-1.rockspec",
    "/luasocket-3.0rc1-2.src.rock",
    "/luasocket-3.0rc1-2.rockspec",
    "/lxsh-0.8.6-2.src.rock",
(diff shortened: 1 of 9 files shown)


hishamhm:
Wow, a comprehensive commit! Kudos for the tests and comments. I admit it's not easy to follow the diff in detail for such a big change, but I got the general idea from the comments and I'll trust the tests.

Feel free to commit the port to `luarocks-3` directly. Thank you!

mpeterv:
@hishamhm thanks for merging! I'll port this to `luarocks-3` soon, In the meantime 2,4,2 can be tagged.

(Please reopen #424, I think the phrase "this doesn't fix #424" tripped github's autoclose feature)

PR #652, opened 2016-11-11
`luarocks show `improvements: fix module paths, add command listing

### LINKED ISSUE #424: luarocks show prints wrong directory

mpeterv:
First commit fixes #424 by replacing `path.which` with `repos.which` that uses recently introduced more precise functions. Second commit adds command listing to the output.

### DIFF
--- a/src/luarocks/cmd/show.lua
+++ b/src/luarocks/cmd/show.lua
@@ -67,6 +67,12 @@ local function installed_rock_label(name, tree)
    return installed and "(using "..version..")" or "(missing)"
 end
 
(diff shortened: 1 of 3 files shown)


hishamhm:
Great, thank you!

PR #654, opened 2016-11-13
Split out luarocks.vers from luarocks.deps

mpeterv:
luarocks.deps module used to contain version and dependency parsing
functions as well as some high-level dependency handling:
matching and resolving them. A lot of modules only need the parsing
half. Split it into its own module, luarocks.vers. Also remove a workaround for a cyclic dependency
between fetch, deps and manif.

### DIFF
--- a/src/luarocks/deps.lua
+++ b/src/luarocks/deps.lua
@@ -1,147 +1,13 @@
 
---- Dependency handling functions.
--- Dependencies are represented in LuaRocks through strings with
--- a package name followed by a comma-separated list of constraints.
--- Each constraint consists of an operator and a version number.
--- In this string format, version numbers are represented as
--- naturally as possible, like they are used by upstream projects
--- (e.g. "2.0beta3"). Internally, LuaRocks converts them to a purely
--- numeric representation, allowing comparison following some
--- "common sense" heuristics. The precise specification of the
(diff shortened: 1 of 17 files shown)


src/luarocks/deps.lua
```diff
@@ -1,147 +1,13 @@
...
--- comparison criteria is the source code of this module, but the
--- test/test_deps.lua file included with LuaRocks provides some
--- insights on what these criteria are.
+--- High-level dependency related functions.
 local deps = {}
-setmetatable(deps, { __index = require("luarocks.core.deps") })
 
 local cfg = require("luarocks.core.cfg")
-local manif = require("luarocks.core.manif")
+local manif = require("luarocks.manif")
```
hishamhm: are you going to move get_versions from luarocks.core.manif to luarocks.manif?

mpeterv: Oh yes, I forgot about the thing that was the initial reason for all this :smile: Yes, I'll move it. Inside or outside this PR, I don't think it matters.

hishamhm: please do it in this PR so we document in the history why the change was made :)

hishamhm: nevermind, I can move it and just reference this PR in the commit!

PR #656, opened 2016-11-27
Add `bin_dir` to luarocks config if `--versioned_rocks_dir` is specified

AitorATuin:
This solves the following problem:

 - luarocks installed for 2 lua versions (5.1, 5.3) with
`--versioned_rocks_dir` enabled.

 - install the same rock in both versions (a rock providing a binary
   command like `ldoc`)

     - luarocks-5.1 install ldoc
     - luarocks-5.3 install ldoc

 - remove the package from both versions

     - luarocks-5.1 remove ldoc
     - luarocks-5.3 remove ldoc

 - the last command returns an error about not being able to found the
   ldoc binary (since it was removed by the previous command).

```
Removing ldoc 1.4.6-2...

Error: Could not remove ../luarocks/trees/bin/ldoc: ../luarocks/trees/bin/ldoc not found
```

The problem is solved fixing the `bin_dir` in the luarocks so I added it to the default configuration when `--versioned_rocks_dir` is enabled.


AitorATuin:
The `bin_dir` is far from being ideal: `/usr/local/bin/lua-version/` and `~/.luarocks/bin/lua-version`. Any suggestion?
Here locally I'm using `/usr/lib/luarocks/bin/lua-version/` and `~/.luarocks/lib/luarocks/bin/lua-version` and after that I use `luarocks path --bin` to set the PATH

hishamhm:
Closing this since it is out of date, and I don't think using versioned subdirectories under bin/ is standard practice in Unix (this would break compatibility for existing installations).

Issue #662, opened 2016-12-31
Site config for defaults.external_deps_dirs ?


xpol:
Currently the defaults.external_deps_dirs hard coded in `src/luarocks/cfg.lua`:

For `unix` it is `defaults.external_deps_dirs = { "/usr/local", "/usr" }` that almost works. But for macOS, there is brew installed OpenSSL library not link to `/usr/local`, this case we needs `/usr/local/opt` too.

For `windows` it is `defaults.external_deps_dirs = { "c:/external/" }` that maybe not flexible at all. Maybe user already have a better place, eg `C:\Program Files\GnuWin`.

In both above case, user may need to have a site config item like `site_config.LUAROCKS_USER_EXTERNAL_DEPS_DIRS` which is inserted **in front of** above `defaults.external_deps_dirs`.

hishamhm:
The default is just, as the name says, a default. This configuration can be overridden using the config file, either per-user (`~/.luarocks/config.lua`) or system-wide (`/etc/luarocks/config-5.x.lua`). There is no need to put this into site_config because this is not a "bootstrap" configuration (in fact, I am considering deprecating site-config altogether).

PR #663, opened 2016-12-31
basical site config for defaults.external_deps_dirs.

xpol:
see #662

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,6 @@
+coverage:
+  notify:
+    gitter:
+      default:
+        url: "https://webhooks.gitter.im/e/c8ee7e8a380f711aea39"
+        threshold: 1%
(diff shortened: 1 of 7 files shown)


mpeterv:
@xpol if I understand it correctly the main difference between this and #668 are that this PR adds an option to add several external_dep_dirs, as opposed to just one. If so, I think I like this one better, it's more flexible.

I'm not sure if this new config option should append custom dirs to default lists or overwrite them completely. The latter is more flexible, and it's consistent with `LUAROCKS_[RUNTIME_]EXTERNAL_DEPS_SUBDIRS` options.

In any case, these PRs should probably target `luarocks-3` branch.

xpol:
Yes,  overwrite them completely would make code simpler.
And It seems I need to make an other PR based on luarocks-3.

PR #664, opened 2017-01-06
Check cache owner

hishamhm:
Check ownership of cache directory and emit a warning. This prevents `sudo luarocks` to take over ownership of the user's ~/.cache/luarocks directory.

This is a blatant copy of pip's equivalent feature, down to the error message. :)

### DIFF
--- a/src/luarocks/command_line.lua
+++ b/src/luarocks/command_line.lua
@@ -187,7 +187,9 @@ function command_line.run_command(...)
       util.warning("The directory '" .. cfg.local_cache .. "' or its parent directory "..
                    "is not owned by the current user and the cache has been disabled. "..
                    "Please check the permissions and owner of that directory. "..
-                   "If executing pip with sudo, you may want sudo's -H flag.")
+                   (cfg.is_platform("unix")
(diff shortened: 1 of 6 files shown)


mpeterv:
It looks like this doesn't handle the case when local cache directory does not exist yet. Not sure what the logic should be in this case. If cache directory doesn't exist, should the ownership of its parent be checked? What if both cache dir and its parent are missing?

hishamhm:
> It looks like this doesn't handle the case when local cache directory does not exist yet. Not sure what the logic should be in this case. Not sure what the logic should be in this case. If cache directory doesn't exist, should the ownership of its parent be checked?

Ah, that's what line 186 was for. I'll double-check.

> What if both cache dir and its parent are missing?

Based on pip's error message I assumed going down one level would be enough.

hishamhm:
@mpeterv The error check was indeed incorrect. Thank you for the review.

> Based on pip's error message I assumed going down one level would be enough.

I decided to check up to two levels, to the grand-parent directory, because on Unix that will usually be `$HOME`, and on Windows,  that's `%LOCALAPPDATA%` (even though we're not doing proper ownership checking on Windows yet).

PR #709, opened 2017-06-03
Don't fail LuaRocks installation when site_config.lua doesn't exist

jirutka:
It causes me a lot of trouble when embedding LuaRocks into another Lua program. Also the expected site_config.lua path is incorrect when installing on Windows.


Stanzilla:
@jirutka can you fix the conflicts, please?

hishamhm:
Closing this PR since LR3 no longer uses site_config. Thank you for bringing the issue to our attention, this was another needle in site_config's coffin! :)

PR #712, opened 2017-06-28
Static build

robooo:
Preparation & refactoring to support the build of static libraries


src/luarocks/build.lua
```diff
@@ -389,17 +390,18 @@ end
 -- @return boolean or (nil, string, exitcode): True if build was successful; nil and an
 -- error message otherwise. exitcode is optionally returned.
 function build.command(flags, name, version)
+   local deps_mode = deps.get_deps_mode(flags)
```
daurnimator: With this line you can also pass `deps_mode` directly to `manif.check_dependencies` below.

src/luarocks/build/builtin.lua
```diff
@@ -167,8 +167,28 @@ local function compile_library(variables, flags)
...
+         return nil, "Failed compiling object "..object
+      end
+      table.insert(objects, object)
+   end
+   return objects
+end
+
 --- Driver function for the builtin build back-end.
 -- @param rockspec table: the loaded rockspec.
+-- @param flags table: the flags table passed to run() drivers.
```
daurnimator: This annotation seems to be in the wrong commit.

src/luarocks/make.lua
```diff
@@ -49,7 +49,8 @@ To install rocks, you'll normally want to use the "install" and
 ]]
 
 --- Driver function for "make" command.
--- @param name string: A local rockspec.
+-- @param flags table: the flags table passed to run() drivers.
```
daurnimator: Annotation change in wrong commit?

src/luarocks/build/builtin.lua
```diff
@@ -143,12 +143,20 @@ local function compile_library(variables, flags)
...
       end
    else
-      compile.object = function(object, source, defines, incdirs)
+      compile.dynamic_object = function(object, source, defines, incdirs)
          local extras = {}
          add_flags(extras, "-D%s", defines, variables)
          add_flags(extras, "-I%s", incdirs, variables)
          return execute(variables.CC.." "..variables.CFLAGS, "-I"..variables.LUA_INCDIR, "-c", source, "-o", object, unpack(extras))
       end
+      if flags["static"] then
```
daurnimator: You can create this function unconditionally. Only the calling of it needs to be conditional on `flags`.

hishamhm:
Also, any idea on why the Travis build failed?

mpeterv:
Could be a bug in the relatively recently released LuaJIT 2.1.0-beta3.

daurnimator:
> Also, any idea on why the Travis build failed?

Yes. there is a bug in lua-cjson that makes it incompatible with the latest luajit 2.1. It's a known issues that can be ignored. https://github.com/LuaJIT/LuaJIT/issues/325

src/luarocks/build/builtin.lua
```diff
@@ -151,27 +143,68 @@ function builtin.run(rockspec)
...
+   if info[1] then sources = info end
+   if type(sources) == "string" then sources = {sources} end
+   for _, source in ipairs(sources) do
+      local object = source:gsub("%.[^.]*$", "."..cfg.obj_extension)
+      if not object then
+         object = source.."."..cfg.obj_extension
+      end
+      local ok = compile_object(object, source, info.defines, info.incdirs)
+      if not ok then
+         return nil, "Failed compiling object "..object
```
mpeterv: So, `build_objects` can return `nil` and error message, but callers do not check it, causing an error later when `nil` is used instead of `objects` array, like in the travis failure.

mpeterv:
@daurnimator right. Tweaking `.travis.yml` and `appveyor.yml` to install LuaJIT 2.1.0-beta2 specifically should work for now.

src/luarocks/build/builtin.lua
robooo: should I add control of `nil` objects before calling `build_objects`?

mpeterv: No, it should be after `build_objects`. Just like `build_objects` catches that `compile_object` can return `nil` and propagates the error, `builtin.run` should check if `build_objects` returned `nil` and handle it.

src/luarocks/static_flags.lua
```diff
@@ -0,0 +1,41 @@
...
+   local rockspec = fetch.load_rockspec(rockspec_file)
+   local command = ""
+
+   -- Do I need to fulfill deps? like this?
+   -- local ok, err, errcode = deps.fulfill_dependencies(rockspec, "all")
+   -- if err then return nil, err, errcode end
+
+   if #rockspec.dependencies > 0 then
+      for _, dep in ipairs(rockspec.dependencies) do
+         if not dep.name:match("lua$") then
```
daurnimator: This needs to be recursive. i.e. we also need flags of dependencies of dependencies of dependencies

src/luarocks/static_flags.lua
```diff
@@ -0,0 +1,41 @@
...
+
+local util = require("luarocks.util")
+local fetch = require("luarocks.fetch")
+local path = require("luarocks.path")
+local search = require("luarocks.search")
+
+util.add_run_function(static_flags)
+
+local function collect_libs(module_name, libs_list)
+   local name, version = search.pick_installed_rock(module_name)
```
daurnimator: This can return nil and and error message.

daurnimator: You need to pass the tree to `pick_installed_rock`

src/luarocks/static_flags.lua
```diff
@@ -0,0 +1,41 @@
...
+         if not dep.name:match("lua$") and not libs_list:match("luarocks-"..dep.name..".a") then
+            libs_list = collect_libs(dep.name, libs_list)
+         end
+      end
+   end
+   return libs_list
+end
+
+function static_flags.command(flags, module_name)
+   local libs_list = ""
```
daurnimator: Makes no sense to decalre empty string as own var here.

src/luarocks/static_flags.lua
```diff
@@ -0,0 +1,41 @@
...
+package.loaded["luarocks.static_flags"] = static_flags
+
+local util = require("luarocks.util")
+local fetch = require("luarocks.fetch")
+local path = require("luarocks.path")
+local search = require("luarocks.search")
+
+util.add_run_function(static_flags)
+
+local function collect_libs(module_name, libs_list)
```
daurnimator: Perhaps collect_libs should collect into a table? (instead of generating many strings via appending)

src/luarocks/static_flags.lua
daurnimator: There should be some way to select a version

.travis.yml
```diff
@@ -40,7 +40,7 @@ matrix:
 
 
 before_install:
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update;fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update && brew install libuv;else sudo -E apt-add-repository -y ppa:linuxjedi/ppa && sudo apt-get -qq update  && sudo apt-get install make automake libtool curl && curl -sSL https://github.com/libuv/libuv/archive/v1.8.0.tar.gz | sudo tar zxfv - -C /usr/local/src && TEMP=`pwd` && cd /usr/local/src/libuv-1.8.0 && sudo sh autogen.sh && sudo ./configure && sudo make && sudo make install && sudo rm -rf /usr/local/src/libuv-1.8.0 && cd $TEMP && sudo ldconfig;fi
```
robooo: I want to have more complex test for this, inner flags etc. but I always find just some dev modules.

src/luarocks/static_flags.lua
```diff
@@ -0,0 +1,41 @@
...
+end
+
+function static_flags.command(flags, module_name)
+   local libs_list = ""
+   local collected_libs = collect_libs(module_name, libs_list)
+   print(collected_libs:sub(1, -2))
+   return collected_libs
+end
+
+return static_flags
```
daurnimator: missing new line at end of file

src/luarocks/static_flags.lua
```diff
@@ -3,7 +3,80 @@ local static_flags = {}
...
+Example:
+$ gcc -o myapp main.c -llua $(luarocks static-flags module_name)
+]]
+
+local function print_static_flags(collected_libs)
+   local full_string = ""
+   for _, value in pairs(collected_libs) do
+      full_string = value .. full_string
+   end
+   print(full_string:sub(1, -2))
```
daurnimator: Use `util.printout` instead of print?

src/luarocks/static_flags.lua
```diff
@@ -3,7 +3,80 @@ local static_flags = {}
...
+This command is useful for building an application with static libraries.
+
+Example:
+$ gcc -o myapp main.c -llua $(luarocks static-flags module_name)
+]]
+
+local function print_static_flags(collected_libs)
+   local full_string = ""
+   for _, value in pairs(collected_libs) do
+      full_string = value .. full_string
```
daurnimator: Don't concatenate in a loop like this: built up a table and then concat it together with `table.concat`

daurnimator: Infact, you already have it in table form, so just use `table.concat`

src/luarocks/static_flags.lua
```diff
@@ -0,0 +1,89 @@
...
+
+   local rockspec = fetch.load_local_rockspec(path.rockspec_file(name, version), false)
+   local lib_name = "luarocks-"..module_name..".a "
+   if not fs.exists(lib_name) then
+      return nil, "Library ".. lib_name .." is missing. Verify location of library or install with 'luarocks make --static <module>'."
+   end
+   if rockspec.external_dependencies then
+      for _, desc in pairs(rockspec.external_dependencies) do
+         if desc.library then
+            lib_name = "-l"..desc.library.." "..lib_name
```
daurnimator: You could insert this as two separate items into the table `"-l"..desc.library` and `lib_name`.

src/luarocks/static_flags.lua
```diff
@@ -0,0 +1,89 @@
...
+         end
+      end
+   end
+
+   local libs_table = libs_table or {}
+   table.insert(libs_table, lib_name)
+   if #rockspec.dependencies > 0 then
```
daurnimator: Can skip this check: `ipairs` will not run the loop if there are no items.

src/luarocks/build/builtin.lua
```diff
@@ -151,27 +143,73 @@ function builtin.run(rockspec)
...
+end
+
+--- Driver function for the builtin build back-end.
+-- @param rockspec table: the loaded rockspec.
+-- @param flags table: the flags table passed to run() drivers.
+-- @return boolean or (nil, string): true if no errors ocurred,
+-- nil and an error message otherwise.
+function builtin.run(rockspec, flags)
+   assert(type(rockspec) == "table")
+   local compile = {} --TODO EXEWRAPPER
```
daurnimator: Why initialise to empty table here just to assign to something else below?

.travis.yml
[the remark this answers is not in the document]
daurnimator: Instead of having this line in `before_install`, perhaps you could move it out to it's own bash script?

src/luarocks/static_flags.lua
```diff
@@ -0,0 +1,89 @@
...
+local fs = require("luarocks.fs")
+
+util.add_run_function(static_flags)
+
+static_flags.help_summary = "Returns all static libraries with flags required for the compiler's Lua flag."
+
+static_flags.help_arguments = "{<name> [<version>]}"
+static_flags.help = [[
+The argument may be the name of locally available module with newest version
+or defined version (optional), which has already built all static libraries.
```
daurnimator: This sentence doesn't flow right.

src/luarocks/static_flags.lua
```diff
@@ -0,0 +1,89 @@
...
+   local name, version = search.pick_installed_rock(module_name:lower(), version, flags["tree"])
+   if not name then
+      util.printout(name..(version and " "..version or "").." is not installed.")
+      return nil, version
+   end
+
+   local rockspec = fetch.load_local_rockspec(path.rockspec_file(name, version), false)
+   local lib_name = "luarocks-"..module_name..".a "
+   if not fs.exists(lib_name) then
+      return nil, "Library ".. lib_name .." is missing. Verify location of library or install with 'luarocks make --static <module>'."
```
daurnimator: I think this error should have `luarocks install` not `luarocks make`?

daurnimator: The trailing space in `lib_name` creates a double space in this error message. See other comments about addressing the trailing space.

daurnimator:
Can show the .a file in `luarocks show` output?

src/luarocks/build/builtin.lua
```diff
@@ -151,27 +143,73 @@ function builtin.run(rockspec)
...
+            add_flags(extras, "-l%s", {"lua"}, variables)
          end
          return execute(variables.LD.." "..variables.LIBFLAG, "-o", library, "-L"..variables.LUA_LIBDIR, unpack(extras))
       end
-      compile_wrapper_binary = function(_, name) return true, name end
+      compile.static_library = function (library, objects, libraries, libdirs, module_name)
+         local static_library_name = "luarocks-"..module_name..".a"
+         return execute(variables.AR, variables.ARFLAGS, static_library_name, unpack(objects))
+      end
+
```
daurnimator: unwanted whitespace

src/luarocks/build/builtin.lua
```diff
@@ -240,10 +267,28 @@ function builtin.run(rockspec)
             if not ok then return nil, err end
          end
          lib_modules[module_name] = dir.path(libdir, module_name)
-         ok = compile_library(module_name, objects, info.libraries, info.libdirs, name)
+         ok = compile.dynamic_library(module_name, objects, info.libraries, info.libdirs, name)
          if not ok then
             return nil, "Failed compiling module "..module_name
          end
+
+         if flags["static"] then
```
daurnimator: Need to fail gracefully on platforms where static libraries are currently unsupported (e.g. win32)

src/luarocks/build/builtin.lua
```diff
@@ -218,20 +256,9 @@ function builtin.run(rockspec)
...
-            local object = source:gsub("%.[^.]*$", "."..cfg.obj_extension)
-            if not object then
-               object = source.."."..cfg.obj_extension
-            end
-            ok = compile_object(object, source, info.defines, info.incdirs)
-            if not ok then
-               return nil, "Failed compiling object "..object
-            end
-            table.insert(objects, object)
+         local objects = build_objects(info, compile.dynamic_object)
```
daurnimator: This section should be conditional on `if flags["dynamic"] then`

src/luarocks/build/builtin.lua
```diff
@@ -240,10 +267,28 @@ function builtin.run(rockspec)
...
          if not ok then
             return nil, "Failed compiling module "..module_name
          end
+
+         if flags["static"] then
+            objects = build_objects(info, compile.static_object)
+            if not objects then
+               return nil, "Failed build of objects for static library"
+            end
+            module_name = name:match("([^.]*)$").."."..util.matchquote(cfg.lib_extension)
```
daurnimator: This is meant to be creating the .a file output name.

src/luarocks/build/builtin.lua
robooo: shouldn't I add `or flags["dynamic"] and flags["static"]` to get dynamic libs always?

src/luarocks/build/builtin.lua
robooo: should I extend this condition of platforms? or just after `if not ok then` condition?

daurnimator: After this line I would put something like: `if compile.static_object == nil then return nil, "static library compilation not available for your platform" end`
You'd then need to fix up which platforms default to static compilation on.

src/luarocks/build/builtin.lua
daurnimator: Do you always want dynamic libs?

If I write `luarocks make --static` I don't really expect a dynamic lib to come out as well.
That's that `luarocks make --static --dynamic` would be for.

src/luarocks/static_flags.lua
```diff
@@ -0,0 +1,87 @@
...
+      return nil, version
+   end
+
+   local rockspec = fetch.load_local_rockspec(path.rockspec_file(name, version), false)
+   local lib_name = "luarocks-"..module_name..".a"
+   local libs_table = libs_table or {}
+   if not fs.exists(lib_name) then
+      return nil, "Library ".. lib_name .." is missing. Verify location of library or create it with 'luarocks make --static <module>'."
+   end
+   table.insert(libs_table, "luarocks-"..module_name..".a")
```
daurnimator: Use `lib_name` here?

PR #716, opened 2017-07-06
Change setup/teardown in make tests to before/after due to wrong use

robooo:
Setup and teardown should be used before `describe` blocks, not `it` blocks (based on busted documentation). This blocked adding and executing new `_spec` files after make tests too.

### DIFF
--- a/spec/make_spec.lua
+++ b/spec/make_spec.lua
@@ -101,11 +101,11 @@ describe("LuaRocks make tests #blackbox #b_make", function()
    end)
 
    describe("LuaRocks make upgrading rockspecs with mixed deploy types", function()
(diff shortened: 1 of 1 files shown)


daurnimator:
Is this fixing a particular issue?

The current code looks to be intentional (the directory only needs to be cleaned up after the `describe` block, not after each test)

robooo:
yes you will get always 
`spec/make_spec.lua:XYZ: attempt to concatenate field 'lib_extension' (a nil value)`
when running tests which are after `make_spec.lua` tests when using teardown. So I could change it to one `it` test before and after OR add if condition `if lib_extension then removing mdt folders`

PR #717, opened 2017-07-07
Forget deprecated options from command line help.

alkino:


hishamhm:
Thank you! Nice catch!

PR #722, opened 2017-08-07
tools.delete does not work with files

### LINKED ISSUE #670: Cannot install any package - Windows 8.1
When attempting to install any package using the following command as an example:

`luarocks install lbase64`

I am getting the following error message and the package is not installed:
```
Setting environment for using Microsoft Visual Studio 2010 x64 tools.
The directory name is invalid.
Installing https://luarocks.org/lbase64-20120807-3.src.rock
cl /nologo /MD /O2 -c -Folbase64.obj -IC:/CS-dev/SVNCommsEngine/Katana/trunk/Toolkit/lua-5.3.2/src lbase64.c
lbase64.c
link -dll -def:base64.def -out:base64.dll C:/tools/lua/lua53.lib lbase64.obj
Microsoft (R) Incremental Linker Version 10.00.40219.01
Copyright (C) Microsoft Corporation.  All rights reserved.

   Creating library base64.lib and object base64.exp
Warning: C:\tools\lua\systree/lib/lua/5.3/base64.dll is not tracked by this installation of LuaRocks. Moving it to C:\tools\lua\systree/lib/lua/5.3/base64.dll~
The directory name is invalid.
The directory name is invalid.

Error: Failed move: could not delete C:\tools\lua\systree/lib/luarocks/rocks/lbase64/20120807-3/lib/base64.dll after copy.
```

The problem seems to be that the function `fs_lua.move ` successfully copies the file; however when attempting to delete the file copied, the file is not deleted. When attempting to delete the file, the following function is called from luarocks/src/luarocks/fs/win32/tools.lua:
```
function tools.delete(arg)
   assert(arg)
   assert(arg:match("^[a-zA-Z]?:?[\\/]"))
   fs.execute_quiet("if exist "..fs.Q(arg.."\\").." ( RMDIR /S /Q "..fs.Q(arg).." ) else ( DEL /Q /F "..fs.Q(arg).." )")
end
```

The script is attempting to delete a file; however it always executes the `RMDIR `command causing the message `The directory name is invalid.`

sebthom:
Function tools.delete performs a test to determine if a path is a file or a directory to select the appropriate delete command `rmdir` vs `del`.
The current test however results in `rmdir` being used on files too, which then results in a build abortion with error `The directory name is invalid.`. 

E.g. `if exist "c:\luarocks\share\lua\5.2\luasocket_3_0rc1_2-socket.lua\" ( echo "I am a folder" ) else ( echo "I am a file" )` wrongly prints `I am a folder` (tested on Windows 10)

Whereas `if exist "c:\luarocks\share\lua\5.2\luasocket_3_0rc1_2-socket.lua\*" ( echo "I am a folder" ) else ( echo "I am a file" )` correctly prints `I am a file` (tested on Windows 10)

This fixes #670

### DIFF
--- a/src/luarocks/fs/win32/tools.lua
+++ b/src/luarocks/fs/win32/tools.lua
@@ -93,7 +93,7 @@ end
 function tools.delete(arg)
    assert(arg)
    assert(arg:match("^[a-zA-Z]?:?[\\/]"))
-   fs.execute_quiet("if exist "..fs.Q(arg.."\\").." ( RMDIR /S /Q "..fs.Q(arg).." ) else ( DEL /Q /F "..fs.Q(arg).." )")
+   fs.execute_quiet("if exist "..fs.Q(arg.."\\*").." ( RMDIR /S /Q "..fs.Q(arg).." ) else ( DEL /Q /F "..fs.Q(arg).." )")
 end
 
 --- Recursively scan the contents of a directory.


hishamhm:
Merged, thank you!

PR #724, opened 2017-08-12
Makefile.setup.inc: do not include trailing slash in LUADIR

blueyed:
LUADIR is used with a following slash everywhere, and it looks confusing
when seeing "//" during the build.


hishamhm:
Merged, thank you!

PR #732, opened 2017-09-26
read just one line when checking md5

mikz:
in some cases reading the whole file fails (on macOS High Sierra, LuaJIT) (by returning nil)
this is timing dependent and waiting between spawning and reading solves the issue as well

### DIFF
--- a/src/luarocks/fs/tools.lua
+++ b/src/luarocks/fs/tools.lua
@@ -144,7 +144,7 @@ function tools.get_md5(file)
    local cmd = md5_cmd[cfg.md5checker]
    if not cmd then return nil, "no MD5 checker command configured" end
    local pipe = io.popen(cmd.." "..fs.Q(fs.absolute_name(file)))
-   local computed = pipe:read("*a")
+   local computed = pipe:read("*l")
    pipe:close()
    if computed then
       computed = computed:match("("..("%x"):rep(32)..")")


hishamhm:
Hi? What LuaJIT version is this on? Sounds like a bug on their (or the OS's?) end.

mikz:
@hishamhm happens to me on [`openresty/1.11.2.5`](https://openresty.org/en/changelog-1011002.html) with https://github.com/openresty/luajit2/releases/tag/v2.1-20170808

mikz:
@hishamhm I know it sounds like interpreter/OS bug, but the chance of fixing it is higher here than on interpreter/OS level.

Even putting some sleep in between spawning the command and reading fixes the issue. 

The code is correct - it needs just one line, right? Then this patch is a good workaround until this gets fixed in different levels.

hishamhm:
Could you try a stable LuaJIT release and confirm that the bug is in their end?

That version you linked isn't even an official LuaJIT release...

hishamhm:
Also:

> The code is correct - it needs just one line, right? Then this patch is a good workaround until this gets fixed in different levels.

Agree, but it also important to get proper diagnostics to know whose fault it is, so we can notify the proper parties.

mikz:
I have a working example: https://github.com/3scale/apicast/commit/e6d5e2703e8bf5f0c5d6f60ec1ce38ff79b1b5eb


Hm. When trying to reproduce it now (after upgrading to macOS HIgh Sierra GM from Beta 3) I get different error:

when executing 
 `cd '/Users/mikz/Developer/apicast/lua_modules/lib/luarocks/rocks/lua-resty-http/0.10-0' && find * 2> /dev/null`

`ERROR: /usr/local/share/lua/5.1/luarocks/fs/unix/tools.lua:115: Interrupted system call`

Any ideas how to debug that?

hishamhm:
The first thing I'd try would be to run the script with a `luajit` interpreter directly and not `resty`. (That's available at `<prefix_of_openresty>/luajit/bin/luajit`).

If that still fails, I'd try a stable/beta LuaJIT release and see if it makes a difference.

mikz:
@hishamhm you are right. It fails with resty but works with plain luajit interpreter. And I get various errors with the resty interpreter. Either the interrupted system call or 
`ERROR: bin/dependencies:23: Failed producing checksum: Failed to compute MD5 hash for file /Users/mikz/Developer/apicast/lua_modules/lib/luarocks/rocks/lua-resty-http/0.10-0/doc/LICENSE
`

Going to report it to resty-cli.

hishamhm:
> Going to report it to resty-cli.

@mikz They're probably going to say you shouldn't be using LuaRocks from within OpenResty, because it does blocking I/O. The `resty` wrapper is actually an NGINX launcher. Doesn't seem to be what you want to use there anyway.

mikz:
@hishamhm I hope not because resty is supposed to handle blocking io just fine (but blocking). Even documentation mentions blocking io in the init phase.

In my case resty is the only available interpreter on the PATH, so thats what I have.

Closing this as this fix would not fix the other error (Interrupted system call) anyway.

mikz:
OpenResty suggests using `*l` as a workaround. https://github.com/openresty/resty-cli/issues/35#issuecomment-332676170

Would you be ok if I'd fix all the occurrences of this?

mikz:
Also this patch fixes the lines issue, but definitely should be used only on demand:

```lua
local lines = getmetatable(io.output()).lines

getmetatable(io.output()).lines = function(self, ...)
    local iter = lines(self, ...)

    return function()
        local ok, ret = pcall(iter)

        if ok then return ret
        else return nil end
    end
end

```

hishamhm:
oh, btw, instead of writing that script you could just write a rockspec will all your dependencies (using `dependencies = { "foo == 1.0", "bar == 2.3" } ` to pin versions, and run `luarocks install --only-deps --tree=lua_modules my-dev-environment-1.0-1.rockspec`

mikz:
@hishamhm I know. But I want transitive dependency locking being automatic like cargo or bundler does. So I'm writing a wrapper around luarocks (and opm in the future) to provide explicit transitive locking via a lockfile.

hishamhm:
Instead of a wrapper, let's add this feature into LuaRocks instead! I'd be happy to merge in this feature and help out getting a PR in. :)

mikz:
@hishamhm That would be nice. But for now I'd prefer prototyping it outside. Also I'm not sure how the OPM support would integrate with LuaRocks. The repository will be soon up on  https://github.com/3scale/lua-rover with first prototype.

hishamhm:
@mikz I see you're using the LuaRocks internal modules directly. This is good because it will make it easy to move the functionality inside LuaRocks once you're done prototyping, but it's also a bit concerning because those are not public APIs guaranteed to be stable (even though in practice they have been mostly stable and a proper stable LuaRocks-as-a-library module is in the works).

PR #733, opened 2017-10-03
Add `luarocks which` command.

hishamhm:


src/luarocks/cmd/which.lua
```diff
@@ -0,0 +1,30 @@
+
+--- @module luarocks.which_cmd
+-- Driver for the `luarocks which` command.
+local which_cmd = {}
+
+local loader = require("luarocks.loader")
+local cfg = require("luarocks.core.cfg")
+local util = require("luarocks.util")
+
+which_cmd.help_summary = "Tell which file is loaded for a given module name."
```
daurnimator: Grammar here doesn't seem right.

hishamhm: How about "Tell which file corresponds to a given module name." ?

PR #734, opened 2017-10-03
Use versioned files only.

hishamhm:
Let's take the opportunity of a new major version to make an
important cleanup: getting rid of the error-prone unversioned
configuration files. This drops support for:

* Unversioned `config.lua` -> use `config-5.x.lua`
* Unversioned `luarocks/site_config.lua` -> it always generates `luarocks/core/site_config_5_x.lua`
* Unversioned `lib/luarocks/rocks` -> it always uses `lib/luarocks/rocks-5.x`

People who want to use a single config.lua file for all versions can always resort to symlinks (I heard that even Windows has them nowadays!)


test/test_environment.lua
```diff
@@ -210,7 +209,7 @@ function test_env.set_args()
    if not test_env.TEST_TARGET_OS then
       title("OS CHECK")
 
-      if execute_bool("sw_vers") then
+      if execute_bool("sw_vers") then
```
daurnimator: whitespace at end of line

hishamhm: I actually put that back in this commit to revert a whitespace-only diff in the full PR, so the spurious change doesn't get in when the PR is squashed.

(I mind the noise of whitespace-only diffs way more than I do the actual presence of trailing whitespace. :) )

PR #741, opened 2017-10-14
Allow <module>.lua and <module>/init.lua in write_rockspec

### LINKED ISSUE #672: `write_rockspec` skips `init.lua` file
`write_rockspec` skips `<module_name> / init.lua` file in case of presence `<module_name> .lua` file

To illustrate, in empty directory:

    echo "" > a.lua
    mkdir a
    echo "" > a/init.lua

    luarocks write_rockspec --lua_version=5.3

Resulted "modules" table in rockspec will contain 

    a = "a.lua"

This is a problem as I often have both "a.lua" and "a/init.lua". (And I await that two records will be generated.):

    a = "a.lua",
    ["a.init"] = "a/init.lua"

My offer is 

in case of two files pretending on one module name, spawn two records.

ryanplusplus:
Catch the case where both <module>.lua and <module>/init.lua are written to build.modules.<module> and ensure that when this happens, <module>/.init.lua is written to build.modules.<module .. '.init'> instead.

Fixes #672

### DIFF
--- a/src/luarocks/cmd/write_rockspec.lua
+++ b/src/luarocks/cmd/write_rockspec.lua
@@ -188,7 +188,17 @@ local function fill_as_builtin(rockspec, libs)
    for _, file in ipairs(fs.find()) do
       local luamod = file:match("(.*)%.lua$")
       if luamod and not luamod_blacklist[luamod] then
-         rockspec.build.modules[path.path_to_module(file)] = prefix..file
+         local path_to_module = path.path_to_module(file)
+         if rockspec.build.modules[path_to_module] then
+            if file:match("init%.lua$") then
+               rockspec.build.modules[path_to_module..".init"] = prefix..file
+            else
+               rockspec.build.modules[path_to_module..".init"] = rockspec.build.modules[path_to_module]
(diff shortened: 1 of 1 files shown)


mpeterv:
Rockspec like this will confuse luarocks greatly. It implicitly relies on that results of `path.path_to_module` is unique for each rockspec. So having `foo.lua` and `foo.so` modules in the same rockspec breaks it, same with `foo.lua` and `foo/init.lua`. (Only one of the modules will be stored in the manifest.)

hishamhm:
@mpeterv That's true. We need tests for these situations, and decide whether we want to support it (even if rockspec 3 only).

The thing is that, as confusing as it is, standard Lua path orders do allow one to have both foo.lua and foo/init.lua at the same time, and to have `require "foo"` matching the former and `require "foo.init"` matching the latter.

As you said the codebase needs more changes to support that, and it's a manifest-breaking change. Should we support that? (If not, we should then at least give an error then?)

hishamhm:
This issue has been solved in LR3 via another refactor, in which LuaRocks no longer gets confused by `foo` vs `foo.init`. But thank you for bringing the issue to our attention!

Issue #756, opened 2017-12-27
luarocks should look for "libfoo.lib" when looking for "foo" on Windows


seishun:
I'm trying to install the `http` module. I'm getting the following error:

```
Error: Failed installing dependency: https://luarocks.org/cqueues-20171014.51-0.src.rock - Could not
 find library file for OPENSSL
  No file ssl.lib in c:/external/lib
  No file ssl.dll in c:/external/lib
  No file libssl.dll in c:/external/lib
```

However, there are no files named like that in my OpenSSL installation. The closest I have is `libssl.lib`. luarocks should look for this pattern too.

daurnimator:
The libraries have different names on windows.
The cqueues rockspec doesn't include the windows-specific names, as it doesn't support windows.

See https://luarocks.org/manifests/daurnimator/luaossl-scm-0.rockspec for what linking to openssl looks like (pay mind to the windows specific sections)

seishun:
It became clear later than the `http` module doesn't work on Windows at all, so this is no longer relevant to me.

PR #757, opened 2017-12-27
Add "lib?.lib" as a library pattern

seishun:
OpenSSL installs found [here](http://slproweb.com/products/Win32OpenSSL.html) have filenames such as `libssl.lib`, `libcrypto.lib`.

Fixes https://github.com/luarocks/luarocks/issues/756.

### DIFF
--- a/src/luarocks/core/cfg.lua
+++ b/src/luarocks/core/cfg.lua
@@ -465,7 +465,7 @@ if cfg.platforms.windows then
 
    defaults.external_deps_patterns = {
       bin = { "?.exe", "?.bat" },
-      lib = { "?.lib", "?.dll", "lib?.dll" },
+      lib = { "?.lib", "?.dll", "lib?.lib", "lib?.dll" },
       include = { "?.h" }
    }
    defaults.runtime_external_deps_patterns = {


daurnimator:
I thought that website's downloads had them as e.g. `ssleay.dll`. I tested it with luaossl.
Have a look at the win32 section of https://luarocks.org/manifests/daurnimator/luaossl-scm-0.rockspec

seishun:
It became clear later than the `http` module doesn't work on Windows at all, so this is no longer relevant to me.

Tieske:
The change makes sense though

daurnimator:
> The change makes sense though

Howso? I don't know of any examples where this would help.

Tieske:
see OP. But also the list provides a way to find matches, and the change just makes it more generic.

daurnimator:
> see OP.

The installer at http://slproweb.com/download/Win32OpenSSL_Light-1_0_2n.exe installs `libeay32.dll` and `ssleay32.dll`. Neither of these will be found with this PR.

> But also the list provides a way to find matches, and the change just makes it more generic.

As far as I know, it's a naming scheme not used on windows: there would be no package fixed by adding these search paths.

Tieske:
then it doesn't make sense indeed :+1:

PR #772, opened 2018-03-05
Tests: add unit tests for some filesystem functions

georgeroman:


spec/fs_spec.lua
```diff
@@ -47,11 +47,158 @@ describe("Luarocks fs test #whitebox #w_fs", function()
...
+           tmpdir = os.tmpname()
+           os.remove(tmpdir)
+           lfs.mkdir(tmpdir)
+           assert.truthy(fs.is_dir(tmpdir))
+       end)
+
+       it("returns false when the argument is a file", function()
+           tmpfile = os.tmpname()
+           local fd = assert(io.open(tmpfile, "w"))
+           fd:write("foo")
```
daurnimator: Needs `assert`

hishamhm:
@georgeroman Tests look good! Minor detail to fix: it's all indented with 4-spaces, LuaRocks codebase uses 3-spaces (to keep the 4-space and 2-space camps equally unhappy :laughing: ).

georgeroman:
@hishamhm I modified and added some more tests 😀

georgeroman:
I'll have a look these days and see how to make them pass on Windows.

spec/fs_spec.lua
```diff
@@ -47,11 +47,665 @@ describe("Luarocks fs test #whitebox #w_fs", function()
...
+         end
+         if tmpdir then
+            lfs.rmdir(tmpdir)
+            tmpdir = nil
+         end
+      end)
+
+      it("returns true when the argument is a directory", function()
+         tmpdir = os.tmpname()
+         os.remove(tmpdir)
```
Tieske: just a reminder that `os.tmpname` usually does not include a path on Windows, but on other platforms it does. 

So no biggy, but using it might leave artefacts in the test directory if those tests fail.

example: https://github.com/stevedonovan/Penlight/blob/2fb9b303f349955597c229f767d421423eaaf2c8/lua/pl/path.lua#L385-L396

PR #773, opened 2018-03-08
Changed fs.pop_dir() a bit

Nair18:
fs.pop_dir() who's purpose actually is to return true if the pop is successful and change current directory to previous directory, isn't doing the same so I have made a small change in the function. I dont know if I misinterpreted the function. Please correct me if I am wrong.


daurnimator:
This is not correct. `popdir` undoes the most recent `pushdir`. It doesn't just go up a directory.

PR #789, opened 2018-04-15
cfg.lua: Expose LUA_VERSION variable for use by rockspec authors

### LINKED ISSUE #719: Target Lua version variable
I can not find way to pass Lua version suffix to makefile. I simply want to pass same value as `luarocks config --lua-ver`. I try `$(LUA_VER)` and `$(LUA_VERSION)` but without success

Update. 
I can not just call `luarocks config --lua-ver` from makefile because I have several luarocks at same time
`luarocks`, `luarocks-5.1`,  `luarocks-5.2` etc.

jiteshpabla:
config.lua: Use LUA_VERSION variable

Fixes #719

### DIFF
--- a/src/luarocks/path.lua
+++ b/src/luarocks/path.lua
@@ -228,6 +228,9 @@ function path.configure_paths(rockspec)
    vars.CONFDIR = path.conf_dir(name, version)
    vars.BINDIR = path.bin_dir(name, version)
    vars.DOCDIR = path.doc_dir(name, version)
+   if rockspec:format_is_at_least("3.0") then
+      vars.LUA_VERSION = _VERSION:match(" (5%.[123])$") or "5.1"
+   end
    rockspec.variables = vars
 end
 


hishamhm:
There is one other tricky bit which I didn't foresee in #719. New rockspecs written with this LUA_VERSION variable available will break when running with older `luarocks` versions. To trap this, we need to make this variable available only if `rockspec_format` is set to 3.0 or above.

This should be possible by setting the variable in `paths.configure_paths`, where one can use `rockspec:format_is_at_least("3.0")`.

jiteshpabla:
I'm not sure if I understand how you came across this "tricky bit", but can you suggest me a way to understand how exactly you got to this conclusion? (like- which lua files should I try to read first to understand this. etc.)

I want to fully understand what I am doing before adding more commits.

hishamhm:
@jiteshpabla that's a great question.

Your code is correct in which it does expose LUA_VERSION to rockspecs.

What triggered my spider-sense is that whenever rockspecs gain a new ability, this can be considered a compatibility breakage. In other words, if the version of the code now in `master` can process a rockspec that the versions of LuaRocks out there can't, we are risking that users who install the new version will create rockspecs that work for them but don't work for other users running older versions.

To solve this problem, the next major LuaRocks release will be called 3.0, and rockspecs that use new features that are available in 3.0+ only need to add `rockspec_format = "3.0"`. Older/current versions of LuaRocks recognize this directive and fail gracefully with a message saying "you need to upgrade LuaRocks to use this rockspec", instead of failing in erratic ways (for example, if `$(LUA_VERSION)` simply expanded to an empty string because it is undefined there).

My next problem then was: how to make sure that users running LuaRocks 3.0 _will_ add `rockspec_format = "3.0"` and not just start using the new features in their rockspecs, causing older versions to fail? The way we do this is that we feature-gate every new feature in the codebase with a version check, using `if rockspec:format_is_at_least("3.0") then ...`. This way, rockspec writers can only access the new features _if_ they add `rockspec_format = "3.0"` to their rockspec. If they don't, the rockspec is assumed to be compatible with LuaRocks 1.x and 2.x (there was no rockspec format break in 2.0).

So there you have it, once I realized this would be new functionality, I noticed that we need to version-gate the feature. And exposing the variable in the `cfg` defaults would do it unconditionally -- doing it where `rockspec.variables` is set [*] we get a chance to version-gate it easily.

Does that make sense?

[*] PS: I'm not entirely happy with the fact that the `rockspec` table is written to after it is loaded, it was a design I'd probably do differently if I started from scratch today. But the current design, for better or worse, is that the rockspec file is loaded into a table and this table is tweaked a bit after the fact, adding `variables` and parsing the `dependencies` tables from string into sub-table entries.

jiteshpabla:
@hishamhm That was very insightful. Thank you.

hishamhm:
@jiteshpabla I retried building, but it's failing on the `luarocks write_rockspec` tests. The reason is because that command builds its own `rockspec` tables, which don't have the `format_is_at_least` function attached to them. You could add a version of that function to those tables in that module, so that your addition to `path` always works. This will make the tests pass.

hishamhm:
@jiteshpabla this branch needs to be rebased locally against `origin/master` and then pushed again with `--force` (it is currently listing 41 commits)

hishamhm:
@jiteshpabla the branch is corrected! But it still needs the fix in `write_rockspec.lua` as outlined in [this comment above](https://github.com/luarocks/luarocks/pull/789#issuecomment-390245463).

hishamhm:
(sorry, hit the wrong button!)

hishamhm:
Don't worry for now: I'm doing a refactor that may fix this indirectly.
Focus on your project, and we'll revisit this after I get the refactor
merged!

PR #790, opened 2018-04-19
Update win32.copy_contents to behave as on Unix

georgeroman:


src/luarocks/fs/win32/tools.lua
```diff
@@ -79,7 +79,8 @@ end
 -- plus an error message.
 function tools.copy_contents(src, dest)
    assert(src and dest)
-   if fs.execute_quiet(fs.Q(vars.CP), "-dR", src.."\\*.*", dest) then
+   assert(fs.is_dir(src))
```
hishamhm: minor change: please use an `if` returning false+message instead of `assert` in this one. The convention is to use `assert()` for data type errors (which are programming mistakes), and error results for all other runtime error conditions.

georgeroman: In this case, the function would behave different on Windows than on Unix, where there is an assertion to check this: https://github.com/luarocks/luarocks/blob/315773caf4ee32c440df27d2339ad0f00460e8c5/src/luarocks/fs/lua.lua#L346. Maybe I should modify it on Unix too, so that it uses an if there too.

hishamhm: Ohh, good catch! That was old code, it would be best to change both, yes.

hishamhm:
Excellent, thanks! Waiting for Appveyor.

PR #794, opened 2018-05-07
Add more tests

georgeroman:


georgeroman:
@hishamhm I think `fs.is_writable` fails on Windows because `fs.chmod` is only implemented on Unix. Manually changing the permissions of the file to nonwritable fixes the tests on Windows.

hishamhm:
Interesting! What did you use to make it nonwritable on Windows?

georgeroman:
I used the `icacls` utility.

hishamhm:
That makes me wonder if instead of fs.chmod we should have in fs a
higher-level function, that we could plausibly implement in both Unix and
Windows. fs.set_writable, fs.set_readable, fs.set_executable? (Without the
Unix distinction of user/group/all) In LuaRocks we use fs.chmod in a very
limited way, after all.

georgeroman:
Some of the previous tests I wrote were run only under Unix because of this permission problem on Windows. Solving this would make them pass on Windows too. I'd be happy to help and do some research on this issue.

georgeroman:
Thank you for this great review! I'll start working on it.

hishamhm:
@georgeroman With the `fs.set_permissions` PR merged, it's a good idea to rebase this one, then I think it'll be good to merge. You can then open a new PR for the next batch of tests.

Issue #546, opened 2016-05-02
Unpacked rocks have creator's permissions


daurnimator:
I installed zash's lua-jwc-0.3-1.src.rock: 
```
$ sudo luarocks-5.1 install lua-jwc
Installing https://luarocks.org/lua-jwc-0.3-1.src.rock...
Using https://luarocks.org/lua-jwc-0.3-1.src.rock... switching to 'build' mode
Updating manifest for /usr/lib/luarocks/rocks-5.1
lua-jwc 0.3-1 is now built and installed in /usr (license: MIT/X11)
```

However, it didn't work due to getting installed with odd permissions:
```
$ lua5.1 -ljwc.jwk
lua5.1: module 'jwc.jwk' not found:
	no field package.preload['jwc.jwk']
	no file './jwc/jwk/init.lua'
	no file './jwc/jwk.lua'
	no file './jwc/jwk.lua'
	no file '/usr/share/lua/5.1/jwc/jwk.lua'
	no file '/usr/share/lua/5.1/jwc/jwk/init.lua'
	no file '/usr/lib/lua/5.1/jwc/jwk.lua'
	no file '/usr/lib/lua/5.1/jwc/jwk/init.lua'
	no file '/usr/local/share/lua/5.1/jwc/jwk/init.lua'
	no file '/usr/local/share/lua/5.1/jwc/jwk.lua'
	no file './jwc/jwk.so'
	no file '/usr/lib/lua/5.1/jwc/jwk.so'
	no file '/usr/lib/lua/5.1/loadall.so'
	no file '/usr/local/lib/lua/5.1/jwc/jwk.so'
	no file './jwc.so'
	no file '/usr/lib/lua/5.1/jwc.so'
	no file '/usr/lib/lua/5.1/loadall.so'
	no file '/usr/local/lib/lua/5.1/jwc.so'
stack traceback:
	[C]: ?
	[C]: ?
```
```
$ luarocks-5.1 show lua-jwc

Error: Could not load rockspec file /usr/lib/luarocks/rocks-5.1/lua-jwc/0.3-1/lua-jwc-0.3-1.rockspec (/usr/lib/luarocks/rocks-5.1/lua-jwc/0.3-1/lua-jwc-0.3-1.rockspec: Permission denied)
```
```
$ sudo luarocks-5.1 show lua-jwc

lua-jwc 0.3-1 - JSON Web Crypto

License: 	MIT/X11
Homepage: 	https://code.zash.se/lua-jwc
[...]

PR #795, opened 2018-05-13
Replace fs.chmod with fs.set_permissions

georgeroman:


src/luarocks/build.lua
```diff
@@ -303,26 +303,26 @@ function build.build_rockspec(rockspec_file, need_to_fetch, minimal_mode, deps_m
       if fs.is_dir(copy_dir) then
          local dest = dir.path(path.install_dir(name, version), copy_dir)
          fs.make_dir(dest)
-         fs.copy_contents(copy_dir, dest)
+         fs.copy_contents(copy_dir, dest, "read")
```
hishamhm: I am concerned about the behavior of `copy_contents`: since a tree of directories contains a mix of readable and executable files (plus all subdirectories themselves need to be executable), the `copy_contents` function needs a mode where it preserves each file's permissions (in other words, the third argument should really be optional). 

I think you'll need to make revert most (all?) places where `copy_contents` was being called with 2 args only.

src/luarocks/fs/lua.lua
```diff
@@ -290,7 +289,6 @@ function fs_lua.copy(src, dest, perms)
    if destmode == "directory" then
       dest = dir.path(dest, dir.base_name(src))
    end
-   if not perms then perms = fs.attributes(src, "permissions") end
```
hishamhm: (Edited a comment I wrote earlier, which was wrong)

This is tricky: this needs to come back. We need `fs.attributes` to get the exact permissions when copying the files over. We could do this using `lfs.attributes` here (which is portable for Windows and Linux) but to _set_ the attributes we need Unix-specific code (on the other hand, we only really need to worry about these permissions on Unix, I think).

My current suggestion is this:

```lua
local fullattrs
if not perms then
   fullattrs = lfs.attributes(src, "permissions")
end
```

and then below for the second part...

src/luarocks/fs/lua.lua
```diff
@@ -302,16 +300,15 @@ function fs_lua.copy(src, dest, perms)
    end
    src_h:close()
    dest_h:close()
-   fs.chmod(dest, perms)
+   fs.set_permissions(dest, perms, "all")
```
hishamhm: ...and then here do

```lua
   if fullattrs and posix_ok then
      -- we can set permissions precisely
      posix.chmod(file, fullattrs)
   else
      -- do an approximation to preserve the executable bit
      if not perms then
         perms = fullattrs:match("x") and "exec" or "read"
      end
      fs.set_permissions(dest, perms, "all")
   end
```

src/luarocks/fs/win32/tools.lua
```diff
@@ -159,6 +159,33 @@ function tools.is_file(file)
...
+-- @param scope string ("user" or "all"): the user(s) to whom the permission applies
+-- @return boolean or (boolean, string): true on success, false on failure,
+-- plus an error message
+function tools.set_permissions(filename, mode, scope)
+   assert(filename and mode and scope)
+
+   local who, what
+   if scope == "user" then
+      who = "%USERNAME%"
+   elseif scope == "all" then
```
hishamhm: I'm not sure if it's the case or not, but please double-check that when the scope is set to "user" then everyone else is being _denied_ access (I know this is correct in your code for Unix with the `00` in the values, I'm not sure on Windows)

This is used in the code to set permissions of the API key for uploads, to make sure that only its owner can see it.

src/luarocks/repos.lua
```diff
@@ -208,7 +208,7 @@ local function move_suffixed(from_file, to_file, suffix)
 
    suffix = suffixed_from_file:sub(#from_file + 1)
    local suffixed_to_file = to_file .. suffix
-   return fs.move(suffixed_from_file, suffixed_to_file)
+   return fs.move(suffixed_from_file, suffixed_to_file, "read")
```
hishamhm: these calls to fs.move should not have "read" too (they are also used copy executables). Just removing these arguments will suffice once the changes to fs.copy discussed above are in place

src/luarocks/upload/api.lua
```diff
@@ -286,4 +286,3 @@ function api.new(flags)
 end
 
 return api
-
```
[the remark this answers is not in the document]
georgeroman: Sorry about that. These were unintended changes. It looks like my editor automatically removed the trailing whitespaces. I'll disable the option from the editor and commit the files without any whitespace change.

src/luarocks/fs/unix/tools.lua
```diff
@@ -159,12 +159,28 @@ function tools.is_file(file)
...
+-- @param mode string ("read" or "exec"): permissions to set
+-- @param scope string ("user" or "all"): the user(s) to whom the permission applies
+-- @return boolean or (boolean, string): true on success, false on failure,
+-- plus an error message
+function tools.set_permissions(filename, mode, scope)
+   assert(filename and mode and scope)
+
+   local perms
+   if mode == "read" and scope == "user" then
+      perms = "600"
```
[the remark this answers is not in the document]
georgeroman: Could you, please, elaborate a bit more on this? How should I moderate the permission based on the umask?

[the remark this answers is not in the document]
daurnimator: A umask is the default permissions to apply when creating a file.
https://en.wikipedia.org/wiki/Umask

e.g. if you want to create a file, you start with `777` then look at umask, lets say it contains `022`, and subtract it: so the default file creation mask is `755`

It's also usually applied by chmod, from example at https://en.wikipedia.org/wiki/Umask#Exceptions:
```bash
$ umask 0000
$ chmod +rwx filename
$ ls -l filename
-rwxrwxrwx filename
$ umask 0022
$ chmod +rwx filename
$ ls -l filename
-rwxr-xr-x filename
```

Some common default umasks are `0022` and `0007`. Usually a distro will have a default, but then it's overridable by the user (which they set in the .bash_profile or similar).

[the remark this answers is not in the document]
daurnimator: See also https://github.com/luarocks/luarocks/issues/546

hishamhm:
@georgeroman Great job on this! Very cool to see API improvements coming out of the testing work.

PR #801, opened 2018-05-28
Tests: cache some binary rocks to speed up tests

hishamhm:
The Linux job for "lua=5.1" on Travis went to about 159s to 142s.
The Windows job for "lua=5.1,msvc" on Appveyor went from about 654s to 569s.

Since Appveyor does not run jobs in parallel, the gains are considerable. Total Appveyor time went from 1h40 to 2h20 once all mock-server tests were enabled; this at least brings the total time down to the previous ballpark.

### DIFF
--- a/spec/build_spec.lua
+++ b/spec/build_spec.lua
@@ -5,7 +5,7 @@ local testing_paths = test_env.testing_paths
 
 test_env.unload_luarocks()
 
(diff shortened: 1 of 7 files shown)


hishamhm:
@georgeroman I made this into a PR first instead of merging directly so you could take a look!

PR #802, opened 2018-05-28
New command: luarocks init

hishamhm:
This implements the plan outlined in "[Project: LuaRocks per-project workflow](https://github.com/luarocks/luarocks/wiki/Project:-LuaRocks-per-project-workflow)".

Given the [new distribution model](https://github.com/luarocks/luarocks/wiki/Project:-LuaRocks-new-distribution-model) is not here yet, it implements "Option 1" in the document linked above.

There are still some avenues for improvement:

* It does not yet process a config file stored inside `.luarocks`. The config file loading scheme is due to be changed with the [new distribution model](https://github.com/luarocks/luarocks/wiki/Project:-LuaRocks-new-distribution-model), so I thought it would be better do it all at once
* There is little configurability to how the project is initialized and the paths in use at this point. Some of it is deliberate, but it would be good to support some "project profiles" (e.g. appropriate settings for Löve2D projects)
* No tests for `luarocks init` yet! @georgeroman ? :)

### DIFF
--- a/src/luarocks/cmd/init.lua
+++ b/src/luarocks/cmd/init.lua
@@ -1,0 +1,68 @@
+
+local init = {}
+
+local cfg = require("luarocks.core.cfg")
+local fs = require("luarocks.fs")
+local path = require("luarocks.path")
+local dir = require("luarocks.dir")
+local util = require("luarocks.util")
+local write_rockspec = require("luarocks.cmd.write_rockspec")
+
+init.help_summary = "Initialize a directory for a Lua project using LuaRocks."
+init.help_arguments = "[<name> [<version>]]"
+init.help = [[
+<name> is the project name.
+<version> is an optional project version.
+
+--license="<string>"     A license string, such as "MIT/X11" or "GNU GPL v3".
+--summary="<txt>"        A short one-line description summary.
+--detailed="<txt>"       A longer description string.
+--homepage=<url>         Project homepage.
+--lua-version=<ver>      Supported Lua versions. Accepted values are "5.1", "5.2",
+                         "5.3", "5.1,5.2", "5.2,5.3", or "5.1,5.2,5.3".
+--rockspec-format=<ver>  Rockspec format version, such as "1.0" or "1.1".
+--lib=<lib>[,<lib>]      A comma-separated list of libraries that C files need to
+                         link to.
+]]
+
+local function write_gitignore()
+   if fs.exists(".gitignore") then
+      return
+   end
+   local fd = io.open(".gitignore", "w")
+   fd:write("lua_modules\n")
+   fd:write("lua\n")
+   fd:close()
+end
+
+--- Driver function for "init" command.
+-- @return boolean: True if succeeded, nil on errors.
+function init.command(flags, name, version)
+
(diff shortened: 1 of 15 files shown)


src/luarocks/cmd/init.lua
```diff
@@ -0,0 +1,108 @@
...
+   util.printout("Preparing ./.luarocks/ ...")
+   fs.make_dir(".luarocks")
+   local config_file = ".luarocks/config-" .. cfg.lua_version .. ".lua"
+   if not fs.exists(config_file) then
+      local fd = io.open(config_file, "w")
+      fd:write("-- add your configuration here\n")
+      fd:close()
+   end
+
+   util.printout("Preparing lua_modules/ ...")
```
daurnimator: `./.luarocks/` above has a leading `./` while this line doesn't.

src/luarocks/cmd/list.lua
```diff
@@ -71,8 +71,10 @@ end
 function list.command(flags, filter, version)
    local query = queries.new(filter and filter:lower() or "", version, true)
    local trees = cfg.rocks_trees
+   local title = "Rocks installed for Lua "..cfg.lua_version
    if flags["tree"] then
       trees = { flags["tree"] }
+      title = "Rocks installed for Lua "..cfg.lua_version.." in " .. flags["tree"]
```
daurnimator: Could just append to `title` here instead of creating afresh?

hishamhm: Sure, why not (building sentences out of fragments in code irks me a bit, but we don't have i18n anyway)

hishamhm:
> I like; 

@daurnimator awesome!

> it picks git: do we need to support other VCS systems too?

Let's do it on-demand: we'll add more if people ask for (or send PRs) for them. `write_rockspec` does have some built-in support for `git` and `hg`.

> luacheck conf file

Any suggestion on what to do regarding this, if anything?

> busted spec/ directory.

As in create an empty one?

Currently, `luarocks test` detects Busted only if you have a `.busted` file. Is it safe to detect it with `spec` alone? I think some other test frameworks use(d) the same directory name, but are any of them still maintained?

daurnimator:
>  Let's do it on-demand: we'll add more if people ask for (or send PRs) for them.

Should we add the flag `--vcs` in anticipation? (and have `--vcs=git` as the default?)

>> luacheck conf file
> Any suggestion on what to do regarding this, if anything?

The `std` field should match the version of lua the project is using. `std="min"` is otherwise a good default.

Additionally, related to the next point, you'll want this snippet:
```lua
files["spec"] = {
	std = "+busted";
}
```

>> busted spec/ directory.
> As in "create an empty one"?

Sure :)

> Currently, luarocks test detects Busted only if you have a .busted file. 

Could also add an empty `.busted` file? or one containing:

```lua
return {
}
```

hishamhm:
@daurnimator cool suggestions, let's think about adding them in future PRs! Maybe the Busted additions need some degree of detection — we don't want to add busted stuff if a project is already using some other test framework. I'm merging this as is now, and @georgeroman will work on the tests.

PR #803, opened 2018-05-29
Add a temporary path utility function to test_env

georgeroman:
This is going to be used a lot in the unit tests so I think it's worth adding it to test_env in order to be accessible anywhere in the tests.


hishamhm:
Sure, makes sense. The Appveyor failure was a network hiccup on their end, so I'm merging this.

PR #806, opened 2018-05-30
Remove the empty file resulted when fs.download fails

georgeroman:
This commit removes the empty file resulted on the failure of fs.download (when it uses `wget` or `curl`)


hishamhm:
Thank you!

PR #809, opened 2018-05-31
Separate the test types (integration and unit)

georgeroman:


hishamhm:
While we're doing this cleanup, I think you can remove the `#i_*` and `#u_*` hashtags too. We can always invoke busted on specific files instead of using them.

hishamhm:
The original plan was to run these as fully separate CI jobs, so we could get a success/failure from the unit tests faster. But I'll merge this as is, for now, because I'm worried that the additional setup/teardown of doubling the number of CI jobs would end up making the overall execution noticeably slower (and the final "green" status).

PR #813, opened 2018-06-05
Test core.util.show_table

georgeroman:


hishamhm:
@georgeroman I was puzzled that this commit did not produce coverage changes and just realized something crazy: we're not actually collecting coverage data of the unit tests!

We launch it with the `run_luarocks*` functions on integration tests, but we weren't loading it on unit. We need to add something like this to all unit test blocks:

```diff
 describe("Luarocks util test #unit", function()
+
+   local runner
+   setup(function()
+      runner = require("luacov.runner")
+      runner.init(test_env.testing_paths.testrun_dir .. "/luacov.config")
+   end)
+   teardown(function()
+      runner.shutdown()
+   end)
+
```

PR #822, opened 2018-06-14
Tests: test command

georgeroman:


spec/test_spec.lua
```diff
@@ -76,3 +78,173 @@ describe("luarocks test #integration", function()
...
+            create_tmp_dir()
+            write_file("test.lua", "", finally)
+         end)
+         
+         after_each(function()
+            destroy_tmp_dir()
+         end)
+         
+         it("returns true if test.lua exists", function()
+            assert.truthy(test_command.detect_type())
```
hishamhm: write the file inside this test instead of `before_each`. this allows the tests to run more independently of each other (and in any order).

PR #825, opened 2018-06-18
Tests: replace small fixture files with inline content

georgeroman:


hishamhm:
Nice! That's a great cleanup.

PR #830, opened 2018-06-24
Tests: add unit tests for build.builtin

georgeroman:


spec/build_spec.lua
```diff
@@ -451,3 +448,434 @@ describe("LuaRocks build tests #integration", function()
...
+               }
+            }
+            build_builtin.autodetect_external_dependencies(build_table)
+            assert.same(build_table, {
+               type = "builtin",
+               modules = {
+                  module1 = {
+                     libraries = "lib",
+                     incdirs = { "$(LIB_INCDIR)" },
+                     libdirs = { "$(LIB_LIBDIR)" }
```
hishamhm: minor, but: could you use something like `foo` instead of `lib` in the examples? This leads to confusing names like `LIB_LIBDIR` :)

spec/build_spec.lua
```diff
@@ -451,3 +448,434 @@ describe("LuaRocks build tests #integration", function()
...
+
+            int luaopen_c_module(lua_State* L) {
+              lua_newtable(L);
+              lua_pushinteger(L, 1);
+              lua_setfield(L, -2, "c_module");
+              return 1;
+            }
+         ]]
+         
+         it("returns false if the rockspec has no build modules and the its format does not support autoextraction", function()
```
hishamhm: typo: "the its"

spec/build_spec.lua
```diff
@@ -451,3 +448,434 @@ describe("LuaRocks build tests #integration", function()
...
+            write_file("test-1.0-1.rockspec", [[
+               rockspec_format = "1.0"
```
hishamhm: probably worth testing with `rockspec_format` not set at all

spec/build_spec.lua
```diff
@@ -451,3 +448,434 @@ describe("LuaRocks build tests #integration", function()
...
+               package = "test"
+               version = "1.0-1"
+               source = {
+                  url = "http://example.com/foo/tar.gz"
+               }
+               build = {}
+            ]], finally)
+            
+            local rockspec = fetch.load_local_rockspec("test-1.0-1.rockspec", true)
```
hishamhm: instead of writing a file and using `fetch.load_local_rockspec` the test will be more specific (ie, using less code that is not the focus of the test) if instead you do `rockspecs.from_persisted_table("test-1.0-1.rockspec", tbl)` (and run slightly faster due to less disk I/O :) )

spec/build_spec.lua
```diff
@@ -132,21 +129,21 @@ describe("LuaRocks build tests #integration", function()
    end)
 
    describe("LuaRocks build - basic builds", function()
-      it("LuaRocks build luadoc", function()
+      pending("LuaRocks build luadoc", function()
          assert.is_true(run.luarocks_bool("build luadoc"))
       end)
       
-      it("LuaRocks build luacov diff version", function()
+      pending("LuaRocks build luacov diff version", function()
```
hishamhm: this one (or preferrably something like this using a fixture) is worth keeping because it tests `luarocks build` receiving a version number as a second argument.

spec/build_spec.lua
```diff
@@ -132,21 +129,21 @@ describe("LuaRocks build tests #integration", function()
...
          assert.is_true(run.luarocks_bool("build luacov 0.11.0-1"))
          assert.is.truthy(lfs.attributes(testing_paths.testing_sys_rocks .. "/luacov/0.11.0-1/luacov-0.11.0-1.rockspec"))
       end)
       
-      it("LuaRocks build command stdlib", function()
+      pending("LuaRocks build command stdlib", function()
```
hishamhm: This kind of test is probably a better candidate for removal because it runs the same command as another test, only using a different existing rock which happens to trigger a different build path (I think this one was in because stdlib used to use `build.type = "command"`...)

spec/build_spec.lua
```diff
@@ -132,21 +129,21 @@ describe("LuaRocks build tests #integration", function()
    end)
 
    describe("LuaRocks build - basic builds", function()
-      it("LuaRocks build luadoc", function()
+      pending("LuaRocks build luadoc", function()
          assert.is_true(run.luarocks_bool("build luadoc"))
       end)
       
-      it("LuaRocks build luacov diff version", function()
```
hishamhm: This one (or preferrably a version of it using fixtures instead of a real module) should probably stay in, as it tests receiving an explicit version number as a second argument (and the choice of luacov was to test building a package that was already installed (ugly, I know — very "implicit" :\ ) )

spec/build_spec.lua
```diff
@@ -451,3 +451,447 @@ describe("LuaRocks build tests #integration", function()
...
+               },
+               build = {
+                  type = "builtin"
+               }
+            }
+            write_file("c_module.c", c_module_source, finally)
+
+            rockspecs.from_persisted_table("c_module-1.0-1.rockspec", rockspec)
+            rockspec.variables["missing_INCDIR"] = lfs.currentdir()
+            rockspec.variables["missing_LIBDIR"] = lfs.currentdir()
```
hishamhm: the variable names that are generated from that `external_dependencies` entry would be `ext_dep_LIBDIR` and `ext_dep_INCDIR`.

spec/build_spec.lua
```diff
@@ -451,3 +451,447 @@ describe("LuaRocks build tests #integration", function()
...
+               local mingw_dir = ""
+               if test_env.MINGW then
+                  mingw_dir = "\\MinGW"
+               end
+               
+               rockspec.variables["openssl_INCDIR"] = "C:\\OpenSSL-Win32\\include" .. mingw_dir
+               rockspec.variables["openssl_LIBDIR"] = "C:\\OpenSSL-Win32\\lib" .. mingw_dir
+            else
+               rockspec.variables["openssl_INCDIR"] = "/usr/local/opt/openssl/include"
+               rockspec.variables["openssl_LIBDIR"] = "/usr/local/opt/openssl/lib"
```
hishamhm: These locations (especially the Windows one) are already declared somewhere in `test_env.lua` (but I think they're declared in a single string, like `OPENSSL_INCDIR=... OPENSSL_LIBDIR=...`. Please edit that to create separate `test_env` variables and reuse them here (and in other places the OpenSSL libs are being used). Better not have hardcoded paths inside the test files that point to specific locations of the CI environment.

spec/build_spec.lua
```diff
@@ -451,3 +451,447 @@ describe("LuaRocks build tests #integration", function()
...
+
+            local rockspec = {
+               rockspec_format = "3.0",
+               package = "c_module",
+               version = "1.0-1",
+               source = {
+                  url = "http://example.com/c_module"
+               },
+               external_dependencies = {
+                  openssl = {
```
hishamhm: As a convention, please use `OPENSSL` here (meaning the variables change from `openssl_INCDIR` etc to `OPENSSL_INCDIR` etc).

spec/build_spec.lua
```diff
@@ -451,3 +451,447 @@ describe("LuaRocks build tests #integration", function()
...
+                     c_module = "c_module.c"
+                  }
+               }
+            }
+            write_file("c_module.c", c_module_source, finally)
+            
+            rockspecs.from_persisted_table("c_module-1.0-1.rockspec", rockspec)
+            assert.truthy(build_builtin.run(rockspec))
+            local lib_file = "lib/luarocks/rocks-" .. test_env.lua_version .. "/c_module/1.0-1/lib/c_module"
+            if test_env.TEST_TARGET_OS == "windows" then
```
hishamhm: we have `test_env.lib_extension`, use that instead

spec/util/test_env.lua
```diff
@@ -224,11 +225,13 @@ function test_env.set_args()
...
          test_env.APPVEYOR = true
-         test_env.OPENSSL_DIRS = "OPENSSL_LIBDIR=C:\\OpenSSL-Win32\\lib OPENSSL_INCDIR=C:\\OpenSSL-Win32\\include"
-      elseif argument:find("^os=") then
+			test_env.OPENSSL_INCDIR = "C:\\OpenSSL-Win32\\include"
+		   test_env.OPENSSL_LIBDIR = "C:\\OpenSSL-Win32\\lib"
+		elseif argument:find("^os=") then
          test_env.TEST_TARGET_OS = argument:match("^os=(.*)$")
       elseif argument == "mingw" then
          test_env.MINGW = true
+			test_env.OPENSSL_LIBDIR = "C:\\OpenSSL-Win32\\lib\\MinGW"
```
hishamhm: indentation is off — hard tabs sneaked in, should be spaces

spec/build_spec.lua
```diff
@@ -451,3 +450,433 @@ describe("LuaRocks build tests #integration", function()
...
+            local rockspec = {
+               rockspec_format = "3.0",
+               package = "c_module",
+               version = "1.0-1",
+               source = {
+                  url = "https://example.com/c_module"
+               },
+               external_dependencies = {
+                  EXTDEP = { 
+						  library = "missing"
```
hishamhm: tabs in this file too

spec/util/test_env.lua
georgeroman: Sorry about that, I changed my text editor and I don't have it fully configured yet. I'm on it right now.

PR #834, opened 2018-06-28
initial implementation for dynamic commands

Tieske:
Instead of a fixed list of commands, a dynamic list of modules
resideding within the `luarocks.cm.<cmd_name>` in the Lua path
will be dynamically loaded. This allows extending LuaRocks with
additional commands.

### DIFF
--- a/src/luarocks/fs/lua.lua
+++ b/src/luarocks/fs/lua.lua
@@ -166,6 +166,52 @@ function fs_lua.dir(at)
    return coroutine.wrap(function() fs.dir_iterator(at) end)
 end
 
+--- List the Lua modules at a specific require path.
+-- eg. `modules("luarocks.cmd")` would return a list of all LuaRocks command
+-- modules, in the current Lua path.
+function fs_lua.modules(at)
+   at = at or ""
+   if #at > 0 then
+      -- turn require path into file path
+      at = at:gsub("%.", package.config:sub(1,1)) .. package.config:sub(1,1)
+   end
+
+   local path = package.path:sub(-1, -1) == ";" and package.path or package.path .. ";"
+   local paths = {}
+   for location in path:gmatch("(.-);") do
+      if location:lower() == "?.lua" then
+         location = "./?.lua"
+      end
+      local _, q_count = location:gsub("%?", "") -- only use the ones with a single '?'
+      if location:match("%?%.[lL][uU][aA]$") and q_count == 1 then  -- only use when ending with "?.lua"
+         location = location:gsub("%?%.[lL][uU][aA]$", at)
+         table.insert(paths, location)
+      end
+   end
+
+   if #paths == 0 then
+      return {}
+   end
+
+   local modules = {}
+   local is_duplicate = {}
+   for _, path in ipairs(paths) do
+      local files = fs.list_dir(path)
+      for _, filename in ipairs(files or {}) do
+         if filename:match("%.[lL][uU][aA]$") then
+           filename = filename:sub(1,-5) -- drop the extension
+           if not is_duplicate[filename] then
+              is_duplicate[filename] = true
+              table.insert(modules, filename)
+           end
+         end
+      end
+   end
+
+   return modules
+end
+
+
 ---------------------------------------------------------------------
 -- LuaFileSystem functions
 ---------------------------------------------------------------------
--- a/src/bin/luarocks
+++ b/src/bin/luarocks
@@ -1,8 +1,5 @@
 #!/usr/bin/env lua
 
--- this should be loaded first.
-local cfg = require("luarocks.core.cfg")
-
 local loader = require("luarocks.loader")
 local cmd = require("luarocks.cmd")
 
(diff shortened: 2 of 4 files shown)


src/luarocks/fs/lua.lua
```diff
@@ -157,6 +157,53 @@ function fs_lua.dir(at)
...
+   at = at or ""
+   if #at > 0 then
+      -- turn require path into file path
+      at = at:gsub("%.", package.config:sub(1,1)) .. package.config:sub(1,1)
+   end
+
+   local path = package.path:sub(-1, -1) == ";" and package.path or package.path .. ";"
+   local paths = {}
+   for location in path:gmatch("(.-);") do
+      location = location:match("^%s*(.-)%s*$")  -- trim whitespace
```
hishamhm: does package.path items ever have whitespace? I don't think it's allowed...

hishamhm:
Looks good, needs the same for `luarocks-admin`

src/luarocks/fs/lua.lua
Tieske: I had LuaJIT available and it does...
```
$ LUA_PATH="/some/dir with;spaces/ " luajit
LuaJIT 2.1.0-beta3 -- Copyright (C) 2005-2017 Mike Pall. http://luajit.org/
JIT: ON SSE2 SSE3 SSE4.1 BMI2 fold cse dce fwd dse narrow loop abc sink fuse
> print(package.path)
/some/dir with;spaces/
> print(package.path .."-")
/some/dir with;spaces/ -
>
```

So I guess it can be done. Not wise, but can be done.

Tieske:
@hishamhm Admin script has also been updated

hishamhm:
@Tieske I just realized we have a problem: when using the all-in-one binary, the `luarocks.cmd.*` modules are not loaded from the filesystem, but via an internal custom package loader.

This feature will have to look both in the filesystem (for third-party commands) and cooperate with some internal feature of the all-in-one-binary to get the default list (e.g. the all-in-one-binary will need to export a global table or something — currently it's all hidden in the registry)

Tieske:
@hishamhm How about we leave the original list in there, and then add the results of the scan to that list? the scan can then remain on the filesystem, because we are only looking for extensions.

Tieske:
and btw, we might then want to scan for `luarocks.cmd.external`. This would allow us to override the internal commands, because entry `list = "luarocks.cmd.list"` will be overwritten by `list = "luarocks.cmd.external.list"`.

hishamhm:
> @hishamhm How about we leave the original list in there, and then add the results of the scan to that list? the scan can then remain on the filesystem, because we are only looking for extensions.

Oh, good idea. That will work, and it's the simplest solution.

> This would allow us to override the internal commands

Not sure it's a good idea to allow third-parties to override the standard commands.

Tieske:
> Not sure it's a good idea to allow third-parties to override the standard commands.

Then let's not advertise it

Tieske:
K, did a quick update (no local testing), waiting for CI

hishamhm:
> Then let's not advertise it

I'm afraid people will do it if we keep it possible. Let's add modules to the `commands` table only if an entry doesn't already exist. Builtin commands should take precedence or else it's a support nightmare ("is your `luarocks build` command the vanilla one?").

The other changes look good!

Tieske:
@hishamhm You had me at "support" 😄

Tieske:
@hishamhm the CI failure seems odd to me. Any ideas?

src/luarocks/fs/lua.lua
hishamhm: But then LUA_PATH is misconfigured: Lua/LuaJIT will use the space like any other character (after all, filenames can _end_ with spaces, as crazy as that may sound):

```
LUA_PATH="$LUA_PATH;/foo bar/?.lua;/bla luajit
LuaJIT 2.1.0-beta3 -- Copyright (C) 2005-2017 Mike Pall. http://luajit.org/
JIT: ON SSE2 SSE3 SSE4.1 BMI2 fold cse dce fwd dse narrow loop abc sink fuse
> require"foo"
stdin:1: module 'foo' not found:
	no field package.preload['foo']
	no file './foo.lua'
	no file './foo/init.lua'
	no file './foo.lua'
	no file '/usr/local/share/lua/5.1/foo.lua'
	no file '/usr/local/share/lua/5.1/foo/init.lua'
	no file '/foo bar/foo.lua'
	no file '/bla '
	no file '/usr/local/lib/lua/5.1/foo.so'
	no file './foo.so'
stack traceback:
	[C]: in function 'require'
	stdin:1: in main chunk
	[C]: at 0x00404c30
```

Spot the space in `'bla '`, it is simply assumed to be part of the name. I'll drop the trim when merging.

hishamhm:
The CI failure was just a Travis/Github hiccup accesing a URL (symptom of relying on lots of live URLs in tests... @georgeroman is fixing some of those as he works on the suite!)

I restarted the suite, should be fine. I'll merge this later today!

hishamhm:
@Tieske thank you for the cleanup! It's merged!

PR #841, opened 2018-07-10
Use $(CURDIR) instead of $(PWD) in Makefiles

mpeterv:
$(CURDIR) is built-in and is guaranteed to always work.
$(PWD) is inherited from the shell and may be incorrect or missing.
In particular, it is invalid if make is run by some program that
changes current directory prior to the execution.

(For example, using PWD breaks hererocks.)


hishamhm:
@mpeterv Merged, thank you! I wasn't aware of `$(CURDIR)`.

PR #843, opened 2018-07-12
Tests: add tests for the dynamic commands

georgeroman:


hishamhm:
@georgeroman Approving and merging this one — wondering if we also need an integration test that actually adds an external command and runs it. That could involve some tricky messing with LUA_PATH within the test suite. Not a priority, but probably worth adding to the end of the log.

PR #859, opened 2018-07-26
Reorganize the doc API and make the cmd doc module use the API

georgeroman:


src/luarocks/init.lua
```diff
@@ -1,6 +1,8 @@
 --- LuaRocks public programmatic API, version 3.0
 local luarocks = {}
 
+local doc = require("luarocks.doc")
+
```
Tieske: is there a reason this one goes ahead of the rest?

src/luarocks/init.lua
```diff
@@ -1295,4 +1228,9 @@ function luarocks.check_missing_dependencies(name, version, tree)
    return nil, "No .rockspec or .rock can be found."
 end
 
-return luarocks
\ No newline at end of file
+for name, func in pairs(doc) do
+   luarocks[name] = func
+end
```
Tieske: I don't like this "mindless" copying of contents:

-  going forward, when more parts get factored out, this API becomes a long list of those copy-loops, which makes the code base harder to understand
- doesn't allow for adding `ldoc` inline comments for documentation generation for the api.
- future accidental additions to modules get automatically exported this way
- above all: I prefer explicitness

src/luarocks/init.lua
georgeroman: I just wanted to separate the API modules from the regular ones, no particular reason besides this.

Tieske: so if you want to create this separation (and migrate stuff as a work in progress), I would start at the top with the clean part. So doing the API=related requires, followed by the explicit exporting of functions (see comment below)

Then a clear comment marker, and everything else below.

But that's how I would do it.

@hishamhm thoughts?

src/luarocks/init.lua
georgeroman: In the last update I exported the API functions in the following way:
```
luarocks = {
   doc = doc.doc,
   homepage = doc.homepage
}
```
Is this what you meant through explicit exporting?

src/luarocks/init.lua
```diff
@@ -1,6 +1,12 @@
 --- LuaRocks public programmatic API, version 3.0
 local luarocks = {}
 
+local doc = require("luarocks.doc")
+luarocks = {
+   doc = doc.doc,
+   homepage = doc.homepage
+}
```
Tieske: you're assigning an empty table to `luarocks` and then overwriting it with another table.

also the idea was to keep whatever should stay at the top, and then work down refactoring the code. As such I think that the `cfg` table and the call to `init` belong in the top part because they will not be refactored

PR #868, opened 2018-08-01
 Do not rewrite path in util.cleanup_path if the desired version is a substring of the version

### LINKED ISSUE #867: util.cleanup_path breaks package.path
I'm using `lenv` (https://github.com/mah0x211/lenv) to install and manage Lua and LuaRocks. Specifically, I am using my fork (https://github.com/ryanplusplus/lenv) which has been patched to look at the new releases URL. 

`lenv` allows multiple releases of 5.x to be installed simultaneously. For example, you can install 5.3.4 and 5.3.5 at the same time (though only one will be active). It does this by including the .4 and .5 in the install path. For example, the `configure` call when install LuaRocks for 5.3.5 is:

```shell
./configure --prefix=/home/ryan/.lenv/lua/5.3.5 \
    --sysconfdir=/home/ryan/.lenv/lua/5.3.5/luarocks \
    --with-lua=/home/ryan/.lenv/lua/5.3.5 \
    --with-lua-bin=/home/ryan/.lenv/lua/5.3.5/bin \
    --with-lua-include=/home/ryan/.lenv/lua/5.3.5/include \
    --with-lua-lib=/home/ryan/.lenv/lua/5.3.5/lib
```

The issue is that `util.cleanup_path` detects `/lua/5.3.5/` as belonging to an invalid version of Lua and replaces it with `/lua/5.3/` thus breaking `package.path` and breaking LuaRocks itself since it can no longer find its dependencies.

I assume that this part of the cleanup needs to be done in order to support multiple versions of Lua in one rocks tree, but could we be pickier about how we match this? Locally I have patched to _not_ rewrite the part if it matches the major and minor versions and this avoids my specific problem. I'm not sure if this is the best path forward, but I will submit this patch for your review.

ryanplusplus:
Closes #867 

This resolves the issue by checking whether the provided version is a substring of the matched version (anchored at the beginning of the string). This allows `/lua/5.3.5` to remain if the provided version is `'5.3'` while still filtering out `/lua/5.2` and `/lua/5.2.5`.

### DIFF
--- a/spec/util_spec.lua
+++ b/spec/util_spec.lua
@@ -155,4 +155,18 @@ describe("Luarocks util test #unit", function()
          assert.truthy(result:find("[\"e\"] = \"4\"", 1, true))
       end)
    end)
+      
+   describe("core.util.cleanup_path", function()
+     it("rewrites versions that do not match the provided version", function()
+        local expected = 'a/b/lua/5.3/?.lua;a/b/c/lua/5.3/?.lua'
+        local result = core_util.cleanup_path('a/b/lua/5.2/?.lua;a/b/c/lua/5.3/?.lua', ';', '5.3')
+        assert.are.equal(expected, result)
+     end)
+
+      it("does not rewrite versions for which the provided version is a substring", function()
+         local expected = 'a/b/lua/5.3/?.lua;a/b/c/lua/5.3.4/?.lua'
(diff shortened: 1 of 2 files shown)


ryanplusplus:
Thanks!

hishamhm:
@ryanplusplus Thank you for the contribution! :)

PR #869, opened 2018-08-02
Rename init.lua to luarocks.lua and move it up one directory

georgeroman:


daurnimator:
Why the move?

georgeroman:
Having the API reside in the previous `init.lua` file, I had to require `luarocks.init` in the `cmd` modules that use the API. Moving this file to `../luarocks.lua` solves this issue and now `require("luarocks")` can be used instead. This would be one reason. Besides this, the new naming seems more suggestive to me.

daurnimator:
I understand the removing `.init` from the require lines.
However the move + rename of the file shouldn't have been required: the lua search path should contain both `?.lua` and `?/init.lua` variants.

georgeroman:
Well, for some reason `?/init.lua` was not in the search path. This happened both locally and in the CI servers. Here is a link to a Travis build which errored because of this: https://travis-ci.org/luarocks/luarocks/jobs/408460972

daurnimator:
> Well, for some reason ?/init.lua was not in the search path

Ah. I think the local form is only in the search path by default for Lua 5.3+.
For lua 5.1 and 5.2 it's in the system-wide path, but not the local one (i.e. `./?.lua` is missing).

Alternatively, perhaps luarocks's or its test harness are modifying LUA_PATH without adding the `init.lua` form.

jiteshpabla:
@georgeroman Have you tried calling it from a simple lua script after installing this branch a luarocks? Because going ```require("luarocks")``` doesn't for me anymore. Moving this file up 1 directory doesn't include it anywhere in the installation.

georgeroman:
Hmm, the source directory in the loaded `luarocks-dev-1.rockspec` is set to `luarocks` so I guess this is why the API file doesn't get copied. After moving it manually, `require("luarocks")` works?

daurnimator:
Note that the Makefile (and new one in my PR) only look for files inside of src/luarocks/.
I feel like this PR was the wrong change to make

jiteshpabla:
Yea, so there are 2 ways to fix this:
 - Undoing this commit
 - changing the makefile
But I'm not sure what would be the best practice here. @hishamhm what would be the best move?

daurnimator:
> changing the makefile

Should happen after #871 (if at all).

-----

I think it made sense for all lua files to be under their own subdir; so I think I would prefer to undo this commit (partially: keep the change in require string, but don't move the file)

georgeroman:
All of that being said, I should probably undo this change.

hishamhm:
@daurnimator I was the one who originally asked for this change. The reasons for that were:

* `luarocks/init.lua` is a bit confusing given that we do have a `luarocks init` command;
* when those root modules are called `init.lua` we end up referring to them as "the init module" in communication (which adds to the confusion — I'm personally partial against the `?/init.lua` pattern, I admit :) )
* I'd like the "require" call of the LuaRocks API to be consistently `require("luarocks")`. While we can police ourselves to do that in the LR codebase, end-users won't. This limits our future options if people start to assume that the API is available as `require("luarocks.init")` — for example, if we decide to bundle the API as a `luarocks.dll` redistributable later.

My suggestion then is to keep this change and fix the Makefile accordingly.

PR #870, opened 2018-08-02
Move some CLI utilities from util to the cmd module

georgeroman:


hishamhm:
Nice!

 `util.printerr` should move out of `luarocks.core.util` as well (it shouldn't even be there in the first place, it is not used anymore in `luarocks.loader` or the `luarocks.core.*` modules)

src/luarocks/build.lua
```diff
@@ -436,7 +438,8 @@ function build.build_rockspec(rockspec, opts)
    ok, err = repos.run_hook(rockspec, "post_install")
    if not ok then return nil, err end
 
-   util.announce_install(rockspec)
+   local cmd = require("luarocks.cmd")
+   cmd.announce_install(rockspec)
```
hishamhm: Would it be possible to move some of these messages out of these internal module functions and into the `cmd` functions that call them?

No need to handle them all yet, but at least some of these messages that run at the very beginning or the very end of these internal functions should be easy to move out into the cmd caller.

georgeroman: Yeah, sure, I'll do it!

hishamhm:
@georgeroman could you check the conflict? it's probably simple

georgeroman:
I rebased on top of the updated luarocks-api branch and the conflict is now gone. Regarding the failing tests, it's only a forgotten `require` somewhere in the code, I'll fix it in a minute.

Issue #853, opened 2018-07-24
How do you remove a bootstrapped luarocks?


THofstee:
I have a bit of a problem. I installed luajit from source, as well as luarocks. When I attempt to do anything in luarocks I get the following error:

```
~/s/luajit master · luarocks remove luarocks
/usr/local/bin/luajit: ...cal/lib/luarocks/rocks-5.1/luarocks/3.0.0-1/bin/luarocks:6: loop or previous error loading module 'luarocks.loader'
stack traceback:
        [C]: in function 'require'
        ...cal/lib/luarocks/rocks-5.1/luarocks/3.0.0-1/bin/luarocks:6: in main chunk
        [C]: at 0x59d566df04f0
```

So I can't use luarocks to remove itself, or install anything, or even call it with no arguments. I'm not sure if the issue is on the luarocks side or the luajit side, but in any case it would help if I knew a complete list of things to remove.

I know I need to remove `/usr/local/bin/luarocks`, but is there anything else that I should remove to avoid conflicts with future installations?

daurnimator:
We should add a `make uninstall` target.

hishamhm:
@daurnimator can we do that without keeping a list of all `luarocks.*` modules inside the Makefile? (I was happy to have gotten rid of that)

daurnimator:
The `make uninstall` now exists. Can this be closed?

THofstee:
Yes go ahead, I am sadly no longer actively using Lua but I trust that this is done.

PR #871, opened 2018-08-05
Build process refactor

### LINKED ISSUE #719: Target Lua version variable

### LINKED ISSUE #857: Installation weirdness
When installing (from current git) the files in `/usr/local/share/lua/5.3/luarocks/` have ownership `zash:zash` (ie me). I'm guessing that this is because the Makefile uses `cp -a` when installing, which preserves the ownership and permissions on the files in my git clone.
That doesn't seem right, normally I would expect `root:root rw-r--r--` on files installed system-wide.

### LINKED ISSUE #863: add support for $(DESTDIR) in the Makefile

daurnimator:
Many fixes to the makefile and configure

Closes #863
Should fix #857
Adds `uninstall` target as noted in #853 
Should fix #719


daurnimator:
I unpushed the last (whitespace) commit due to a conflict in the master branch.
Please merge soon, before other conflicts occur!

hishamhm:
I had some minor tweaks to suggest but I think I'll just merge this as is and apply the tweaks in subsequent commits. Thank you for working through this!

PR #872, opened 2018-08-05
Detect freebsd header location

### LINKED ISSUE #681: LuaRocks ./configure --lua-suffix=51 on FreeBSD
LuaRocks 2.4.2 on a FreeBSD 11 VM

```
# ./configure --lua-suffix=51
Lua version detected: 5.1
Lua interpreter found: /usr/local/bin/lua51...
Checking Lua includes... lua.h not found (looked in /usr/local/include, /usr/local/include/lua/5.1, /usr/local/include/lua5.1)
You may want to use the flag --with-lua or --with-lua-include. See --help.

configure failed.
```

I would have expected it to find `/usr/local/include/lua51` when being explicitly told that the suffix is `51`, yet it uses `$LUA_VERSION` when searching.

daurnimator:
Closes #681

### DIFF
--- a/configure
+++ b/configure
@@ -431,7 +431,6 @@ check_incdir() {
       "$incdir/lua/$LUA_VERSION/lua.h" \
       "$incdir/lua$LUA_VERSION/lua.h" \
       "$incdir/lua.h" \
-      "$incdir/lua$LUA_VERSION/lua.h" \
       $("$LUA_BINDIR/$LUA_INTERPRETER" -e 'print(jit and [['"$incdir"'/luajit-]]..jit.version:match("(%d+%.%d+)")..[[/lua.h]] or "")')
    do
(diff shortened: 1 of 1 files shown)


configure
```diff
@@ -430,8 +430,8 @@ check_incdir() {
    for lua_h in \
       "$incdir/lua/$LUA_VERSION/lua.h" \
       "$incdir/lua$LUA_VERSION/lua.h" \
+      "$incdir/lua${LUA_VERSION/.//}/lua.h" \
```
daurnimator: This is not POSIX sh compatible. See https://stackoverflow.com/a/21913014/282536
I recommend using `tr` for this case instead.

hishamhm: yep, that was the shellcheck failure

PR #883, opened 2018-08-08
fix luajit detection on Buildroot

fperrad:
On Buildroot (cross compiled linux), when building a module for LuaJIT,
`luarocks` is runned by a host Lua 5.1,
and `LUA_LIBDIR` is a directory which contains only a target LuaJIT library.

see https://git.busybox.net/buildroot/commit/?id=bcccf228145bd8a2863279117810531bb7951ed0
and http://autobuild.buildroot.net/results/25c5e7ff5fab5ac45d959d50eded47bec4fa83b5/build-end.log

### DIFF
--- a/src/luarocks/deps.lua
+++ b/src/luarocks/deps.lua
@@ -506,15 +506,13 @@ function deps.check_lua(vars)
    end
    local shortv = cfg.lua_version:gsub("%.", "")
    local libnames = {
+      "luajit-" .. cfg.lua_version,
       "lua" .. cfg.lua_version,
       "lua" .. shortv,
       "lua-" .. cfg.lua_version,
       "lua-" .. shortv,
       "lua",
    }
-   if cfg.luajit_version then
-      table.insert(libnames, 1, "luajit-" .. cfg.lua_version)
-   end
    for _, libname in ipairs(libnames) do
       local ok = check_external_dependency("LUA", { library = libname }, vars, "build")
       if ok then


src/luarocks/deps.lua
```diff
@@ -506,15 +506,13 @@ function deps.check_lua(vars)
...
    local shortv = cfg.lua_version:gsub("%.", "")
    local libnames = {
+      "luajit-" .. cfg.lua_version,
       "lua" .. cfg.lua_version,
       "lua" .. shortv,
       "lua-" .. cfg.lua_version,
       "lua-" .. shortv,
       "lua",
    }
-   if cfg.luajit_version then
```
hishamhm: Would it be possible to make this scenario detect `cfg.luajit_version` instead? I'm worried if adding luajit to the list inconditionally would make it pick up the luajit library when the intention is to use PUC Lua

blueyed: Note that `luajit_version` is typically missing in the config (#1042, #1043).

hishamhm:
@fperrad please let me know if the latest code in `master` renders this PR unnecessary — it should be performing LuaJIT detection dynamically. If not, please consider if setting `LUA_INCDIR` in the LuaRocks configuration would do the trick.

fperrad:
The merge of #1051 does not help.

On Buildroot, in order to cross-compile, the Luarocks config file is appended with the content of 
https://git.busybox.net/buildroot/tree/package/luarocks/luarocks-br-config.lua

hishamhm:
Wouldn't it be sufficient to set LUA_INCDIR in that file to the proper
LuaJIT include dir, then? Am I missing something?

hishamhm:
Alternatively, _not_ setting LUA_INCDIR and letting it detect it at runtime entirely could also fix it, if the LuaRocks configuration points to the LuaJIT interpreter (even if the `luarocks` command itself is running on Lua 5.3).

fperrad:
Ok, I found the trick.
And it works already with 3.1.3

PR #886, opened 2018-08-10
Add function to open doc files to the API.

jiteshpabla:
This PR also:
 - Uses that function for the CLI.
 - Fixes the issue of the doc file not opening due to incorrect path.

### DIFF
--- a/src/luarocks/api/doc.lua
+++ b/src/luarocks/api/doc.lua
@@ -119,6 +119,9 @@ function doc.open_doc(docdir, docfile)
    end
 
    local pathname = dir.path(docdir, docfile)
+   if not fs.browser then
+      fs.init()
+   end
    local ok = fs.browser(pathname)
    if not ok and not docfile:match("%.html?$") then
       local fd = io.open(docfile, "r")
--- a/src/luarocks/api/doc.lua
+++ b/src/luarocks/api/doc.lua
@@ -119,14 +119,16 @@ function doc.open_doc(docdir, docfile)
    end
 
    local pathname = dir.path(docdir, docfile)
+   if not pathname then
+      return nil, "Cannot find path to doc file."
+   end
+   
    if not fs.browser then
       fs.init()
    end
    local ok = fs.browser(pathname)
    if not ok and not docfile:match("%.html?$") then
-      local fd = io.open(docfile, "r")
-      cmd.printout(fd:read("*a"))
-      fd:close()
+      return true, pathname
    end
    
    return true
--- a/src/luarocks/api/doc.lua
+++ b/src/luarocks/api/doc.lua
@@ -109,4 +109,24 @@ function doc.doc(name, version, tree)
    return docdir, nil, files 
 end
 
+function doc.open_doc(docdir, docfile)
+   if not docdir then
+      return nil, "No doc directory given as argument."
+   end
+   
+   if not docfile then
+      return nil, "No doc files given as argument."
+   end
+
+   local pathname = dir.path(docdir, docfile)
+   local ok = fs.browser(pathname)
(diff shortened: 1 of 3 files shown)


hishamhm:
@jiteshpabla you are not exposing the function in the `src/luarocks.lua` module:
https://travis-ci.org/luarocks/luarocks/jobs/414469918#L1502

jiteshpabla:
@hishamhm thanks for pointing that out!
I couldn't figure this out on my own by reading the failing tests on AppVeyor; they seemed to be completely different.

src/luarocks/api/doc.lua
```diff
@@ -109,4 +109,27 @@ function doc.doc(name, version, tree)
...
+      return nil, "No doc files given as argument."
+   end
+
+   local pathname = dir.path(docdir, docfile)
+   if not fs.browser then
+      fs.init()
+   end
+   local ok = fs.browser(pathname)
+   if not ok and not docfile:match("%.html?$") then
+      local fd = io.open(pathaname, "r")
```
hishamhm: there's a typo here (pathaname) — (which means this code path is untested!)

src/luarocks/api/doc.lua
```diff
@@ -109,4 +109,27 @@ function doc.doc(name, version, tree)
...
+      cmd.printout(fd:read("*a"))
```
hishamhm: In any case, this part is CLI-specific (it prints the doc to the terminal), and is not a behavior you'll want in the GUI. This part should stay in `luarocks.cmd.doc`.

For the GUI, if the doc file is not an `.html`, I think it should open the file, read its contents and embed it in the page. It will often be a `.txt` or a `.md` (in the latter case it would be really cool to load a Markdown library and render HTML with it (there are a few Lua libraries that can do it in a couple of lines).

src/luarocks/api/doc.lua
jiteshpabla: I forgot to push that in a hurry.

src/luarocks/api/doc.lua
jiteshpabla: Keeping the GUI on the side, shouldn't the doc function try to open up that file using ```xdg-open``` (or other equivalent commands)? Since LuaRocks already detects the OS specific command, should there be a more generalised function to open files other than fs.browser (if a function like that doesn't exist already)? This could be useful for more features in the API, for example - opening up the directory of a specific rock-tree (like I do in the GUI).

jiteshpabla:
@hishamhm Are the changes fine now? I want to include this commit in my GSoC submission.

src/luarocks/api/doc.lua
hishamhm: > Since LuaRocks already detects the OS specific command, should there be a more generalised function to open files other than fs.browser (if a function like that doesn't exist already)? This could be useful for more features in the API, for example - opening up the directory of a specific rock-tree (like I do in the GUI).

That would be one possibility. It could use `start` on Windows, `open` on the Mac and `xdg-open` on Linux.

> IIRC, fs.browser can open up a bunch of different extensions and directories too, so I fail to understand the need for this specific case - and not docfile:match("%.html?$").

It is a bit arbitrary indeed. It arose from the observation that printing out README files and such to the terminal is usable (though perhaps it would be best to use a pager program on Unix, such as `less`) but printing HTML reads like "garbage" on the terminal. (We can't always assume that the `luarocks` CLI user has GUI access, they may be using it from an SSH session).

jiteshpabla: The SSH session is an interesting case, I'm not sure how something like `xdg-open` acts in such a situation. Anyway, I guess dwelling too much on this is overkill.

PR #911, opened 2018-08-31
show: base listing of modules on rock_manifest

hishamhm:
List modules based on rock_manifest, for better precision. Overall management
of modules should be moved to use rock_manifest instead of the modules list
from the general local manifest, but this is a good first step.

### DIFF
--- a/src/luarocks/cmd/show.lua
+++ b/src/luarocks/cmd/show.lua
@@ -4,6 +4,7 @@ local show = {}
 
 local queries = require("luarocks.queries")
 local search = require("luarocks.search")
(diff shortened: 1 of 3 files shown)


hishamhm:
@tst2005 I loosened the error protocol in that function on purpose so that callers don't need to add a `return true` to the traversal. So it's only really an error if you get an error. I know this is a bit nonstandard, but I thought it would reduce the boilerplate in practice.

tst2005:
@hishamhm I understand your approach. It is not a mistake like I thought.

hishamhm:
@tst2005 cool, thank you for the careful review :)

PR #912, opened 2018-09-03
Fix package removal

hishamhm:
This reworks the logic of upgrading/downgrading multiple versions of the same package (or different packages) when they have different files that resolve to the same module name (foo.lua and foo.so).

For a concrete example, this was hitting LuaSec downgrades.


spec/make_spec.lua
```diff
@@ -109,6 +109,109 @@ describe("LuaRocks make tests #integration", function()
...
+         print(run.luarocks("make ddt/double_deploy_type-0.2.0-1.rockspec"))
+         assert.is.truthy(lfs.attributes(deploy_lib_dir.."/ddt."..so))
+         assert.is.truthy(lfs.attributes(deploy_lua_dir.."/ddt.lua"))
+         assert.same("ddt2", loadfile(deploy_lua_dir.."/ddt.lua")())
+         assert.is.truthy(lfs.attributes(deploy_lua_dir.."/ddt_file"))
+         assert.is.falsy(lfs.attributes(deploy_lib_dir.."/ddt."..so.."~"))
+         assert.is.falsy(lfs.attributes(deploy_lua_dir.."/ddt.lua~"))
+         assert.is.falsy(lfs.attributes(deploy_lua_dir.."/ddt_file~"))
+
+         print(run.luarocks("make ddt/double_deploy_type-0.1.0-1.rockspec --keep"))
```
daurnimator: prints meant to be in here?

hishamhm:
@daurnimator thanks for catching that!

PR #916, opened 2018-09-16
Remove unnecessary old check from fs.set_permissions

georgeroman:


hishamhm:
Thank you! The Appveyor failure seems unrelated, right?

georgeroman:
Yes, as the branch was always taken before these modifications, the behavior is the same as before.

hishamhm:
Cool, I believe I fixed that Appveyor issue! Thanks!

PR #932, opened 2018-10-07
Add the unpack command to the API

georgeroman:


src/luarocks/cmd/unpack.lua
```diff
@@ -21,126 +16,6 @@ In the latter case, the app version may be given as a second argument.
...
-      if rockspec.source.dir ~= "." then
-         local ok = fs.copy(rockspec.local_abs_filename, rockspec.source.dir, "read")
-         if not ok then
-            return nil, "Failed copying unpacked rockspec into unpacked source directory."
-         end
-      end
-      cmd.printout()   
-      cmd.printout("Done. You may now enter directory ")
-      cmd.printout(dir.path(dir_name, rockspec.source.dir))
-      cmd.printout("and type 'luarocks make' to build.")
```
hishamhm: Looks like these messages were lost in the conversion — these could remain in the `luarocks.cmd.unpack` module. One way to do this would be to return `kind, nil, dir.path(dir_name, rockspec.source.dir)` on success, and then in `unpack.command` to something like

```lua
local kind, err, srcdir = luarocks.unpack(ns_name, flags["force"])
if not kind then
   return nil, err
end
if kind == "src" or kind == "rockspec" then
   cfg.log()
   cfg.log("Done. You may now enter directory ")
   cfg.log(srcdir)
   cfg.log("and type 'luarocks make' to build.")
end
return true
```

hishamhm:
Yay! Left a comment for a minor review item, otherwise looks good!

src/luarocks/cmd/unpack.lua
georgeroman: Since these messages are only relevant when using `luarocks` from the terminal, I kept using `cmd.printout` instead of `cfg.log`.

georgeroman:
Great, thank you! I'm really looking forward to keep working on this API!

src/luarocks/cmd/unpack.lua
hishamhm: @georgeroman Makes total sense!

PR #933, opened 2018-10-17
Fix packing an installed rock when custom module deployment paths are set

THE-FYP:
LuaRocks fails to copy module files of an installed rock if the deployment paths are changed in the rock tree, it ignores them and tries to copy from the default paths.

### DIFF
--- a/src/luarocks/pack.lua
+++ b/src/luarocks/pack.lua
@@ -102,12 +102,12 @@ function pack.pack_installed_rock(query, tree)
 
    local is_binary = false
    if rock_manifest.lib then
-      local ok, err = copy_back_files(name, version, rock_manifest.lib, path.deploy_lib_dir(root), dir.path(temp_dir, "lib"), "exec")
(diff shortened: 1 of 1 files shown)


hishamhm:
Hi! Thanks for the patch! Could you give an example of a configuration that demonstrates the problem?

THE-FYP:
Sure.
```lua
rocks_trees = {
    home..[[/luarocks]],
    { name = [[user]],
         root    = home..[[/luarocks]],
    },
    { name = [[system]],
         root    = [[D:\luarocks-test\luarocks\systree]],
         lib_dir = [[D:\luarocks-test\lib]],
         lua_dir = [[D:\luarocks-test\lib]],
    },
}
variables = {
    MSVCRT = 'MSVCR80',
    LUALIB = 'lua51.lib',
}
verbose = false   -- set to 'true' to enable verbose output
```
It wasn't edited manually, `lib_dir` and `lua_dir` were set by installation with flags `/LUAMOD` and `/CMOD`.

hishamhm:
@THE-FYP thank you! Merging this. Thank you for the fix!

PR #935, opened 2018-10-21
Fix setting up deployment paths from a specified tree

THE-FYP:
Fixes the problem when a tree to operate in is specified explicitly using the `--tree` flag, the deployment paths are set to default instead of those specified in the config.

**Reproducing**
With config
```lua
rocks_trees = {
    { name = [[system]],
         root    = [[D:\luarocks-test\luarocks\systree]],
         lua_dir = [[D:\luarocks-test\lib]],
    },
}
```
run `luarocks config --tree=system` and you'll see that variable `deploy_lua_dir` points to default path, although running luarocks without the `--tree=system` parameter the path is correct.

### DIFF
--- a/src/luarocks/cmd.lua
+++ b/src/luarocks/cmd.lua
@@ -150,10 +150,10 @@ end
 
 local process_tree_flags
 do
-   local function replace_tree(flags, tree)
-      tree = dir.normalize(tree)
-      flags["tree"] = tree
-      path.use_tree(tree)
+   local function replace_tree(flags, root, tree)
+      root = dir.normalize(root)
+      flags["tree"] = root
+      path.use_tree(tree or root)
    end
 
    local function find_project_dir()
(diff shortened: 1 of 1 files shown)


hishamhm:
Thanks for the patch! Would you consider adding a test to this, to avoid a
future regression? In the `spec/config_spec.lua` and `spec/upload_spec.lua`
files there are some examples of tests which pass a custom config via
`LUAROCKS_CONFIG`. You could create a temporary one using
`test_env.write_file` and then check the output of `run.luarocks("config",
{LUAROCKS_CONFIG = custom_config_file }`.

Using LR3, the LuaRocks test suite should be runnable using `luarocks
test`, you can also do `luarocks test -- spec/config_spec.lua` to run only
the relevant file.

THE-FYP:
I ran the command `luarocks test -- spec/config_spec.lua` and got this error: `Error: Argument missing: please specify a rockspec to use on current directory`. So the description of the `test` command says: "If the first argument is a rockspec, it will use it to determine the parameters for running tests; otherwise, it will attempt to detect the rockspec." which in my mind means that this command is intended for testing rockspecs, am I getting something wrong?
Anyway I don't feel I want to spend time understanding how the LuaRocks test suite works, particularly for such a small change, sorry.

Issue #931, opened 2018-10-06
Do not use `arg` as name for argument in vararg functions


moteus:
There known `feature` in Lua 5.1 that `arg` is special name for varargs.
So when you use `...` and `arg` in same time then `arg` will always be nil.
Here some magic :)
```Lua
function foo(arg, ...) assert(nil == arg) local args = {...} end foo(true)
function boo(arg, ...) assert(type(arg) == 'table') end boo(true)
```
So as result code for test does not work when try to specify rockspec in command line
```
function cmd_test.command(flags, arg, ...)
   assert(type(arg) == "string" or not arg)

   local args = { ... }

   if arg and arg:match("rockspec$") then --- < here arg always nil on Lua 5.1
      return test.run_test_suite(arg, flags["test-type"], args)
   end
```

PR #936, opened 2018-10-22
Rename a misinterpretable vararg

firoxer:
Hi,

I noticed issue #931 and thought I could fix it because the problem looks simple enough. I hope this solution is alright.

Thanks for LuaRocks!


hishamhm:
@firoxer Thank you for the fix!!

PR #943, opened 2018-11-16
Fix granting permissions for all users on localized Windows

THE-FYP:
On localized Windows systems granting permissions using `icacls` to `Everyone` is not correct.
Changing it to `*S-1-1-0` resolves the problem.
Details: https://support.microsoft.com/en-us/help/243330/well-known-security-identifiers-in-windows-operating-systems

### DIFF
--- a/src/luarocks/fs/win32/tools.lua
+++ b/src/luarocks/fs/win32/tools.lua
@@ -278,7 +278,7 @@ function tools.set_permissions(filename, mode, scope)
 
       local ok
       -- Grant permissions available to all users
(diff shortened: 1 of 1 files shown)


Tieske:
lgtm, but would it be possible to add a test validating the behaviour?

THE-FYP:
@Tieske 
is there any reason to add a specific test for this particular case considering that any install test would fail in given environment without this fix?
if you are wondering if this change breaks anything, you can see that all CI checks were successful. i suppose it wouldn't be so if this change breaks something.

hishamhm:
@THE-FYP thank you!

PR #956, opened 2019-01-16
Support addon command

fperrad:
On Buildroot, I wrote an addon, see https://git.busybox.net/buildroot/tree/package/luarocks/0002-Buildroot-addon.patch

Now, I've rebased the first commit of the branch https://github.com/xiaq/luarocks/tree/addon-prototype which allows "addon command".

### DIFF
--- a/src/luarocks/cmd.lua
+++ b/src/luarocks/cmd.lua
@@ -305,6 +305,17 @@ function cmd.run_command(description, commands, external_namespace, ...)
       os.exit(exitcode or cmd.errorcodes.UNSPECIFIED)
    end
 
+   --- Given a command name, find the corresponding module. This can either be a
+   -- module named in the global commands table, or an addon.
+   local function find_command(name)
+      -- Look for builtin command
+      local cmd = commands[name]
+      if cmd then return require(cmd) end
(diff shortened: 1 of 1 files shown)


Tieske:
isn't this a duplicate of https://github.com/luarocks/luarocks/pull/834 (at least same functionality)?

See https://github.com/luarocks/luarocks-gui/blob/master/luarocks-gui-0.0.0-1.rockspec#L26 for an example adding a `gui` command.

fperrad:
@Tieske you are right.

This PR is based on obsolete information from https://github.com/luarocks/luarocks/wiki/Addon-author's-guide

As #834 was merged, this PR is useless.

From an user's point of view, the only difference is the location of the additional command 
- `luarocks/cmd/external/foo.lua` in #834 
- `luarocks/addon/foo.lua` in https://github.com/xiaq/luarocks/tree/addon-prototype

Tieske:
@hishamhm when we came up with #834 I never even though about the addons, should we adjust the name. Or are the addon's a different purpose all together?

Iirc the addon's were also supposed to be able to add a new backend/build right? So they are not the same?

PR #957, opened 2019-01-23
Fix small misspellings

MisterDA:
Found with https://github.com/client9/misspell.


src/luarocks/core/util.lua
```diff
@@ -42,7 +42,7 @@ end
 -- Formats tables with cycles recursively to any depth.
 -- References to other tables are shown as values.
 -- Self references are indicated.
--- The string returned is "Lua code", which can be procesed
+-- The string returned is "Lua code", which can be proceeds
```
georgeroman: I think this should be "processed".

hishamhm:
Merged, thank you!!

PR #958, opened 2019-01-24
Update install.bat

tilkinsc:
The problem here is that `%errorlevel%` is set on `exit /b ]]` and you pass it something funny. By adding `rem=rem` you can reuse a hack and get it still to execute correctly. You don't actually need to even set `rem=rem` you can just pass it nil, since nothing else in the script uses rem.

### DIFF
--- a/install.bat
+++ b/install.bat
@@ -1,7 +1,7 @@
-rem=nil --[[--lua
+rem=rem --[[--lua
 @setlocal&  set luafile="%~f0" & if exist "%~f0.bat" set luafile="%~f0.bat"
 @win32\lua5.1\bin\lua5.1.exe %luafile% %*&  exit /b 0
-rem=nil ]]
+rem=rem ]]
 
 local vars = {}
 


Tieske:
`rem=nil` alters the Lua environment if `rem` was already set to something else. The original `rem=rem` is better imo, since it does not alter the Lua environment.

Also if `exit /b ]]` returns a weird exit code (I didn't test it), then surely the right solution is to use `exit /b` and not `exit /b 0` because in the latter case it always returns exit code 0, where in the former case it would leave the exit code unaltered and return the exit code that LuaRocks exited with, which is the right thing to do imo.

Or even better, make it explicit and do `exit /b %ERRORLEVEL%`.

tilkinsc:
Ahh thats true.
The ]] still gets processed as an argument.

Tieske:
> Also, whats up with the dramatic code coverage declination?

no idea, did you use the latest master branch to create this PR?

tilkinsc:
Yes I did. I edited directly from the website.

Tieske:
nvm the coverage for now, it needs updating according to previous comments

tilkinsc:
The changes were made for rem=rem

Tieske:
But it is still forcing the exit code to 0 instead of returning the result from the call into Lua

hishamhm:
Do we still need this?

Looks like `exit /b` (with or without `%ERRORLEVEL%`) is a big rabbit hole, which I'm not keen on getting into:
https://stackoverflow.com/questions/66248276/difference-between-exit-b-and-exit-b-errorlevel

And from Lua's perspective that second added `rem=rem` is inside a Lua string, so it should never run, so I don't think that's necessary either.

But in practice, I don't think a lot of people rely on our `install.bat`'s produced errorlevel (especially if it doesn't currently work!), and to be honest I'd rather eventually remove this script in favor of other installation methods.

If the current behavior is producing a spurious error message thanks to `exit \b ]]` I'd be happy to just change it to plain `exit \b` and put the `]]` on its own line below.

If this is just producing a bogus errorlevel which goes mainly unnoticed, then I don't care much either way and I would just close the PR.

PR #959, opened 2019-01-25
Support installations without documentation

georgeroman:


src/luarocks/cmd/build.lua
```diff
@@ -156,6 +157,15 @@ function cmd_build.command(flags, name, version)
    if not ok then return nil, err end
    name, version = ok, err
 
+   if flags["no-doc"] then
+      local install_dir = path.install_dir(name, version)
+      for _, file in ipairs(fs.list_dir(install_dir)) do
+         if string.find(file, "doc") then
```
hishamhm: elsewhere in `luarocks doc` we use the heuristic of checking for the exact strings `"doc"` and `"docs"`... deleteing any "doc" substring sounds a bit dangerous :)

hishamhm: Also, I think this should probably be done inside `do_build`, so that combining `--pack-binary-rock` and `--no-doc` works as expected! (Remembering to test flag combinations is always tricky)

hishamhm:
Merged!!

PR #966, opened 2019-02-12
Make environment variable changes local in Windows wrapper scripts

dwenegar:
On Windows, the changes made by the wrapper script to the environment variables are global.
After the execution of a command installed with luarocks (i.e. busted) the variables in the parent environment are left modified.
I assume this is an unwanted effect, and this PR fix it.


hishamhm:
@dwenegar yes, it was unintended! Merged, thank you!

PR #967, opened 2019-02-13
Fix % expansion in windows wrapper.

dwenegar:
While writing the windows wrapper script expanding `%` into `%%`breaks variable expansions.
It should be `set LUA_PATH=....;%LUA_PATH%` while now it is `set LUA_PATH=....;%%LUA_PATH%%`

### DIFF
--- a/src/luarocks/fs/win32.lua
+++ b/src/luarocks/fs/win32.lua
@@ -91,7 +91,7 @@ function win32.Qb(arg)
    arg = arg:gsub('\\(\\*)"', '\\%1%1"')
    arg = arg:gsub('\\+$', '%0%0')
    arg = arg:gsub('"', '\\"')
-   arg = arg:gsub('%%', '%%%%')
+   -- arg = arg:gsub('%%', '%%%%')
    return '"' .. arg .. '"'
 end
 


hishamhm:
@Tieske do you happen to know if this fix is correct? Was it you or @ignacio who originally did the `Qb` function for batch files?

Tieske:
@dwenegar I do not see why this would have broken recently.

Can you provide reproduction steps? So I could try to reproduce. What are the inputs, what do you expect, and what did you get?

dwenegar:
> Can you provide reproduction steps?

siffiejoe:
@hishamhm I think I added the `Qb` function originally. The only purpose of escaping `%` was to avoid (accidental) variable expansions. If variable expansion is intended in quoted strings, this change should do it. I don't have access to a Windows machine anymore, though, so I can't test it.

hishamhm:
Ahhh, I see now. `Qb` was originally being used to quote Lua code when passed as an argument to `-e` in the batch script, and I started using for setting variables in 9941d6d52. 

Issue #980 also reported problems with using `set` instead of `-e` on Linux, so maybe I should just revert that change. TBH, I don't remember why I made that change (should have documented it better in the commit message), maybe it was just to simplify the command line; I don't remember if there were any advantages to setting the variable instead of doing it via `-e`; it was done in the process of making `luarocks` itself wrappable via `wrap_script`.

hishamhm:
Will merge this in the meantime since it fixes the current behavior.

Issue #952, opened 2018-12-25
Warning: Failed searching manifest: Failed fetching manifest - Failed downloading manifest


jensenr30:
I followed the instructions to install luarocks on Windows 7, but I cannot install any rocks.  It seems that luarocks is incapable of downloading the manifest.
I have successfully manually visited [the manifest page](https://luarocks.org/manifest).  My web browser is capable of [downloading the manifest and displaying it to me](https://i.imgur.com/URGAOyM.png).

Does anyone know how this failure occurs?

command line log:
```
Microsoft Windows [Version 6.1.7601]
Copyright (c) 2009 Microsoft Corporation.  All rights reserved.

C:\Windows\system32>luarocks install graph-toolkit

Warning: Failed searching manifest: Failed fetching manifest for https://luarocks.org - Failed downloading https://luarocks.org/manifest - C:\Users\jen\AppData\Local/LuaRocks/Cache/https___luarocks.org/manifest
Warning: Failed searching manifest: Failed fetching manifest for https://raw.githubusercontent.com/rocks-moonscript-org/moonrocks-mirror/master/ - Failed downloading https://raw.githubusercontent.com/rocks-moonscript-org/moonrocks-mirror/master/manifest - C:\Users\jen\AppData\Local/LuaRocks/Cache/https___raw.githubusercontent.com_rocks-moonscript-org_moonrocks-mirror_master_/manifest
Warning: Failed searching manifest: Failed fetching manifest for http://luafr.org/moonrocks/ - Failed downloading http://luafr.org/moonrocks/manifest - C:\Users\jen\AppData\Local/LuaRocks/Cache/http___luafr.org_moonrocks_/manifest
Warning: Failed searching manifest: Failed fetching manifest for http://luarocks.logiceditor.com/rocks - Failed downloading http://luarocks.logiceditor.com/rocks/manifest - C:\Users\jen\AppData\Local/LuaRocks/Cache/http___luarocks.logiceditor.com_rocks/manifest

Error: No results matching query were found.

C:\Windows\system32>
```

vincentkerdraon:
Hello,
same here in Docker

```
FROM alpine

COPY po_vault.lua validate-po.bash ./

RUN apk add --no-cache --purge -uU lua lua-dev git  \
    && apk add --no-cache --virtual .build-deps build-base unzip \
    && cd /tmp \
    && git clone https://github.com/keplerproject/luarocks.git \
    && cd luarocks \
    && sh ./configure \
[...]

PR #968, opened 2019-02-19
Put quotes around platform tool paths

TomSputz:
A solution to #952 

This is completely untested, and I don't know enough about the codebase to know if it's the best solution - I just did this to get my installation working, and hope it can be of use to fixing this problem.

To elaborate -

On windows, luarocks finds the paths of the 'tools' at runtime, using a hardcoded path which is generated by `install.bat`, and it uses this to `os.execute` the tools.

However, if the path has a space in it (which it does by *default* in `Program Files (x86)`), then cmd fails to find the executable and the call fails. This means that the default installation of LuaRocks couldn't function on any windows machine

### DIFF
--- a/src/luarocks/fs/lua.lua
+++ b/src/luarocks/fs/lua.lua
@@ -111,7 +111,7 @@ function fs_lua.is_tool_available(tool_cmd, tool_name, arg)
    arg = arg or "--version"
    assert(type(arg) == "string")
 
-   if not fs.execute_quiet(tool_cmd, arg) then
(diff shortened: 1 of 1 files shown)


hishamhm:
> completely untested [...] I just did this to get my installation working

Well, did it work on _your_ machine at least? :)

hishamhm:
@TomSputz thank you for bringing the problem to my attention. I implemented a different solution in #984 which will feature in the next release.

PR #987, opened 2019-03-20
Introduce --sign and --verify flags

hishamhm:
* Adds a `--sign` flag to `pack`, `upload`, `build --pack--binary-rock` and `make --pack-binary-rock`
* Adds a `--verify` flag to `build` and `install`

The changes to the `upload` command assume extensions to the luarocks.org server-side API.

The `--verify` command assumes that, for every file `F` being used, a `F.asc` file exists alongside it (locally in the same directory or remotely in the same server path).


src/luarocks/cmd/install.lua
```diff
@@ -29,22 +29,40 @@ or a filename of a locally available rock.
...
 --only-deps         Installs only the dependencies of the rock.
+
 --no-doc            Installs the rock without its documentation.
+
+--verify            Verify signature of the rock being installed.
+                    If rock is being downloaded, LuaRocks will attempt
+                    to download the signature as well. If the rock is
+                    local, the signature file should be in the same
+                    directory.
+                    You need the signer’s public key in your local
```
daurnimator: This shouldn't be required IMO: luarocks should be verifying that the package was signed by the luarocks.org user that uploaded for it. Otherwise I could sign someone else's package

src/luarocks/cmd/make.lua
```diff
@@ -84,8 +87,12 @@ function make.command(flags, rockspec_filename)
       branch = not not flags["branch"],
    })
 
+   if flags["sign"] and not flags["pack-binary-rock"] then
+      return nil, "In the make command, --sign is meant to be used only with --pack-binary-rock"
```
daurnimator: why not allow people to sign a src rock?

src/luarocks/cmd/upload.lua
```diff
@@ -12,14 +13,20 @@ upload.help_arguments = "[--skip-pack] [--api-key=<key>] [--force] <rockspec>"
...
+
 --temp-key=<key> Use the given a temporary API key in this invocation only.
                  It will not be stored.
+
 --force          Replace existing rockspec if the same revision of
                  a module already exists. This should be used only 
                  in case of upload mistakes: when updating a rockspec,
                  increment the revision number instead.
+
+--sign           Upload a signature file alongside each file as well.
```
daurnimator: trailing newline?

src/luarocks/signing.lua
```diff
@@ -0,0 +1,49 @@
...
+end
+
+function signing.sign_file(file)
+   local gpg, err = get_gpg()
+   if not gpg then
+      return nil, err
+   end
+   
+   local sigfile = file .. ".asc"
+   fs.delete(sigfile)
```
daurnimator: Would it be better to throw an error if the signature already exists?

src/luarocks/cmd/make.lua
hishamhm: that's allowed, but not in the `make` command. You sign a src.rock with `luarocks pack --sign <rockspec>` (or via `luarocks upload --sign`)

src/luarocks/cmd/install.lua
hishamhm: > luarocks should be verifying that the package was signed by the luarocks.org user that uploaded for it

You mean (1) luarocks.org should verify at upload time? (Assuming the user has a way to upload their key to luarocks.org beforehand)

Or do you mean (2) that luarocks CLI should download the user key from luarocks.org for verifying? (Assuming the user has a way to upload their key to luarocks.org beforehand and the CLI has a way to download it)

AFAICT, any of these two options above would protect a user in case an API key was stolen, but not really in case luarocks.org was compromised (in which case the user public keys could be tampered too), in which case one would need --verify to check against a local key that was obtained from a third-party trusted source.

For the --sign and --verify features in this PR I basically copied the featureset from git (that sentence "You need the signer’s public key in your local..." was literally lifted from the git documentation).

Option (1) could be added to luarocks.org with no changes to this PR. So my reasoning is that this would be sufficient for a v1 of signing functionality.

src/luarocks/signing.lua
hishamhm: if we don't delete, gpg asks if we want to override. Since entering the passphrase implies an interaction anyway, I just removed the `fs.delete`.

src/luarocks/cmd/upload.lua
hishamhm: not needed, shows up fine in --help

src/luarocks/cmd/install.lua
daurnimator: > You mean (1) luarocks.org should verify at upload time? 

It probably should; but that would be separate

> Assuming the user has a way to upload their key to luarocks.org beforehand

They should! otherwise how will we know who is allowed to sign which packages?
I would make this a dependency of the `--verify` half of the PR. But not of `--sign` which can be done first.

> in case luarocks.org was compromised (in which case the user public keys could be tampered too

In such a compromise, you could ask users to mark the gpg keys in their account as correct/valid. And then be suspicious at any signatures by unknown/untrusted keys.

Issue #993, opened 2019-04-04
Incorrect handling of all-negative supported_platforms rockspec value


slact:
The [documentation](https://github.com/luarocks/luarocks/wiki/Rockspec-format) for the `supported_platforms` field states:
>  If only negative entries are listed, the rock is assumed to be portable to any platform except those listed. 

However, in luarocks 3.0.4, a rockspec with an all-negative `supported_platforms` is always treated as on an unsupported platform.

Checksum:
Ran into this with `openresty:alpinel-fat` as well. Any known workarounds?

daurnimator:
This should have been fixed with #994

PR #994, opened 2019-04-04
fix: handle all-negative supported_platforms rockspec field

slact:
fixes issue #993


hishamhm:
@slact merged, thanks!

Issue #992, opened 2019-04-04
supported_platforms error does not print current platforms


slact:
When a luarock fails the `supported_platforms` check, the error message is 
```
Error: This rockspec for ... does not support  platforms.
```

On my system, it should be
```
Error: This rockspec for ... does not support unix, linux platforms.
```

pauloue:
I think this can be closed?

PR #995, opened 2019-04-04
fix: correctly print current platforms

slact:
fixes issue #992

### DIFF
--- a/src/luarocks/core/cfg.lua
+++ b/src/luarocks/core/cfg.lua
@@ -814,7 +814,12 @@ function cfg.init(detected, warning)
    end
 
    function cfg.print_platforms()
-      return table.concat(platforms, ", ")
+      local platform_keys = {}
+      for k,_ in pairs(platforms) do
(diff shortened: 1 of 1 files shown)


src/luarocks/core/cfg.lua
```diff
@@ -814,7 +814,12 @@ function cfg.init(detected, warning)
    end
 
    function cfg.print_platforms()
-      return table.concat(platforms, ", ")
+      local platform_keys = {}
+      for k,_ in pairs(platforms) do
+         table.insert(platform_keys, k)
+      end
+      table.sort(platform_keys)
+      return table.concat(platform_keys, ", ")
```
daurnimator: Just use `platform_order` instead?

slact: Do you mean that this should print `platforms` in `platform_order` rather than ordered alphabetically?

hishamhm:
@slact merged, thank you!

PR #998, opened 2019-04-14
Add the --bin-suffix option

norcalli:
Allow setting a suffix for the luarocks and luarocks-admin binaries to
distinguish between different versions appropriately.


daurnimator:
Since luarocks 3.0 there is no longer a need to have multiple luarocks versions installed. Use the `--lua-version` command line argument.

norcalli:
@daurnimator That option isn't even documented in the command line arguments in the `--help`, the website, the README, or anywhere that it should be. I'm honestly kind of angry at how much time I wasted because of this. Someone should make a PR to add that to the help options, but I'm not going to waste anymore time on this project.

PR #1012, opened 2019-05-14
test: fix reporting failures on 'command' backend

### LINKED ISSUE #1011: Falling tests produce assertion
The following testing scenario produces an assertion:

```lua
-- myrock-dev-1.rockspec
test = {
    type = 'command',
    command = 'false',
}
```

```console
$ luarocks test
/usr/bin/lua5.2: /usr/local/share/lua/5.2/luarocks/cmd.lua:296: bad error, expected string, got: nil
```

The problem is here: https://github.com/luarocks/luarocks/blob/v3.1.2/src/luarocks/test/command.lua#L40-L42
```lua
function command.run_tests(test, args)
    ...
    return fs.execute(test.command, unpack(args))
end
```

Function `run_tests` should `return nil, err`, but `fs.execute` returns only one boolean value

rosik:
Resolve #1011

### DIFF
--- a/src/luarocks/test/command.lua
+++ b/src/luarocks/test/command.lua
@@ -32,14 +32,22 @@ function command.run_tests(test, args)
       end
    end
    
+   local ok
+
    if test.script then
       if not fs.exists(test.script) then
          return nil, "Test script " .. test.script .. " does not exist"
       end
       local lua = fs.Q(dir.path(cfg.variables["LUA_BINDIR"], cfg.lua_interpreter))  -- get lua interpreter configured
-      return fs.execute(lua, test.script, unpack(args))
+      ok = fs.execute(lua, test.script, unpack(args))
    elseif test.command then
-      return fs.execute(test.command, unpack(args))
+      ok = fs.execute(test.command, unpack(args))
+   end
+
+   if ok then
+      return true
+   else
+      return nil, "tests failed with non-zero exit code"
    end
 end
 


rosik:
@hishamhm Could you take a look at the fix, please

src/luarocks/test/command.lua
```diff
@@ -32,14 +32,22 @@ function command.run_tests(test, args)
...
    
+   local ok, err
+
    if test.script then
       if not fs.exists(test.script) then
          return nil, "Test script " .. test.script .. " does not exist"
       end
       local lua = fs.Q(dir.path(cfg.variables["LUA_BINDIR"], cfg.lua_interpreter))  -- get lua interpreter configured
-      return fs.execute(lua, test.script, unpack(args))
+      ok, err = fs.execute(lua, test.script, unpack(args))
```
daurnimator: As `fs.execute` never returns an error message, you don't need to try and get `err` from it.

src/luarocks/test/command.lua
```diff
@@ -32,14 +32,22 @@ function command.run_tests(test, args)
...
    elseif test.command then
-      return fs.execute(test.command, unpack(args))
+      ok, err = fs.execute(test.command, unpack(args))
+   end
+
+   if ok then
+      return true
+   else
+      return nil, err or "tests failed."
```
daurnimator: Perhaps a better error message? e.g. "tests failed with non-zero exit code"?

rosik:
@daurnimator done (I've amended changes)

PR #1013, opened 2019-05-19
Avoid double leading `/` in simple rooted paths

dwenegar:
Remove the extra leading `/` in the result of `canonicalpath` when the input is a simple rooted path (`/hoge`).


hishamhm:
@dwenegar thank you!

PR #1015, opened 2019-05-23
Add per CPU libraries

dwenegar:
Trying to fix the installation of some modules failing because
some libraries have different names depending on the CPU architecture (for example, on Windows, `luasec` assume 32bits libraries).


daurnimator:
Will this need a change in the rockspec validator?
Won't this be a compatibility break? and rockspec that uses it will end up not working in older luarocks releases?

dwenegar:
> Will this need a change in the rockspec validator?
AFAIK the validator doesn't check what's inside the `modules` table.
> Won't this be a compatibility break? and rockspec that uses it will end up not working in older luarocks releases?

dwenegar:
The problem I had stemmed form a funky precompiled version of OpenSSL.

Issue #879, opened 2018-08-06
Ship completion files for shells


daurnimator:
e.g. for bash+zsh

daurnimator:
I'm wondering if this should be done after an overhaul of command line argument handling.
At the moment, flags are stored in a central list: there is no per-module data (outside of help output) about which flags are used for each command.

@hishamhm how would you feel about new argument parsing for luarocks? Things I'd like it to do:
  - auto generate help text
  - validate arguments and display help if wrong
  - be introspectable for generating completion data

nightwarrior-xxx:
@daurnimator Can I work on this issue for GSOC 2019 . 
I have been wrtiting  lua and bash scripts for some time  and have idea how to do this.

daurnimator:
> Can I work on this issue for GSOC 2019 .

@nightwarrior-xxx I assume you saw my comment here? https://github.com/luarocks/luarocks/issues/962#issuecomment-460884004

> I have been wrtiting lua and bash scripts for some time and have idea how to do this.

Could you share your ideas? :)

nightwarrior-xxx:
@daurnimator Will ping you on gitter

PR #1035, opened 2019-06-18
Use argparse for command line argument parsing

pauloue:
[Argparse](https://github.com/luarocks/argparse) supports auto generation of help texts and using it allows eliminating the central list of flags. Argparse also supports generation of shell completion scripts for bash, zsh, and fish.

Closes https://github.com/luarocks/luarocks/issues/879


src/bin/luarocks
```diff
@@ -33,4 +32,4 @@ local commands = {
    test = "luarocks.cmd.test",
 }
 
-cmd.run_command(description, commands, "luarocks.cmd.external", ...)
+cmd.run_command("luarocks", description, commands, "luarocks.cmd.external", ...)
```
daurnimator: Couldn't we keep the old signature here and continue to use `util.this_program()` inside of `run_command`?

src/luarocks/admin/cmd/add.lua
```diff
@@ -11,20 +11,21 @@ local fs = require("luarocks.fs")
...
-
---no-refresh  The local cache should not be refreshed
-              prior to generation of the updated manifest.
---index       Produce an index.html file for the manifest.
-              This flag is automatically set if an index.html
-              file already exists.
-]]
+function add.add_to_parser(parser)
+   local cmd = parser:command("add", "Add a rock or rockspec to a rocks server.",
+      util.see_also())
```
daurnimator: this line wrapping looks weird to me. The luarocks codebase doesn't mind a few longer lines.

src/luarocks/cmd.lua
```diff
@@ -288,6 +286,102 @@ do
    end
 end
 
+local function get_status(status)
+   return status and "ok" or "not found"
+end
+
+local function get_config_text()
```
daurnimator: Could you pass in `cfg` here? Gives us the opportunity in future to print/preview configs without setting globals

src/luarocks/cmd.lua
```diff
@@ -322,69 +429,43 @@ function cmd.run_command(description, commands, external_namespace, ...)
...
    end
 
-   local command
-   if flags["help"] or #nonflags == 0 then
-      command = "help"
-   else
-      command = table.remove(nonflags, 1)
+   -- Compatibility for old flag
+   if args.nodeps then
+      args.deps_mode = "none"
```
daurnimator: Can this be done with argparse aliases?

src/luarocks/admin/cmd/make_manifest.lua
```diff
@@ -11,36 +11,37 @@ local deps = require("luarocks.deps")
...
    
-   local ok, err = writer.make_manifest(repo, deps.get_deps_mode(flags), not flags["local-tree"])
-   if ok and not flags["local-tree"] then
+   local ok, err = writer.make_manifest(repo, deps.get_deps_mode(args), not args["local_tree"])
+   if ok and not args["local_tree"] then
       util.printout("Generating index.html for "..repo)
       index.make_index(repo)
    end
-   if flags["local-tree"] then
+   if args["local_tree"] then
```
daurnimator: Some places in this pull request you use `args["foo"]` syntax, other you use `args.foo`. Given that argparse converts `-` to `_`, then lets go with the `.foo` form?

src/luarocks/cmd/config.lua
```diff
@@ -45,23 +46,35 @@ config_cmd.help = [[
...
+      :summary("Query information about the LuaRocks configuration.")
+      :add_help(false)
+
+   cmd:argument("key", "The configuration key.")
+      :args("?")
+   cmd:argument("value", "The configuration value.")
+      :args("?")
+
+   cmd:option("--scope", "The scope indicates which config file should be rewritten.\n"..
+      'Accepted values are "system", "user" or "project".\n'..
```
daurnimator: This duplicates what the `:choices` already tells the user

src/luarocks/cmd/config.lua
```diff
@@ -218,44 +231,39 @@ local function write_entries(keys, scope, do_unset)
...
-                 or (flags["local"] and "user")
-                 or (flags["project-tree"] and "project")
-                 or (cfg.local_by_default and "user")
-                 or "system"
-   if scope ~= "system" and scope ~= "user" and scope ~= "project" then
-      return nil, "Valid values for scope are: system, user, project"
-   end
-
-   return scope
+local function get_scope(flags)
```
daurnimator: rename the argument here to `args` to match?

src/luarocks/cmd/init.lua
```diff
@@ -10,27 +10,32 @@ local util = require("luarocks.util")
...
+      cmd:option("--license", 'A license string, such as "MIT/X11" or "GNU GPL v3".')
+         :argname("<string>"),
+      cmd:option("--summary", "A short one-line description summary.")
+         :argname("<txt>"),
+      cmd:option("--detailed", "A longer description string.")
+         :argname("<txt>"),
+      cmd:option("--homepage", "Project homepage.")
+         :argname("<url>"),
+      cmd:option("--lua-versions", "Supported Lua versions. Accepted values are "..
+         '"5.1", "5.2", "5.3", "5.1,5.2", "5.2,5.3", or "5.1,5.2,5.3".'),
```
daurnimator: should this be a `:choices`?

src/luarocks/cmd/new_version.lua
```diff
@@ -21,7 +20,7 @@ default server. If a rockspec is given, it uses it instead. If no argument
 is given, it looks for a rockspec same way 'luarocks make' does.
 
 If the version number is not given and tag is passed using --tag,
-it is used as the version, with 'v' removed from beginning.
+it is used as the version, with 'v' removed from beginnnumbering.
```
daurnimator: typo here

src/luarocks/cmd.lua
```diff
@@ -328,7 +328,7 @@ Variables:
 ]]..get_config_text()
 
    local parser = argparse(
-      program_name, "LuaRocks "..cfg.program_version..", the Lua package manager\n\n"..
+      dir.base_name(program), "LuaRocks "..cfg.program_version..", the Lua package manager\n\n"..
```
daurnimator: Is this correct for `luarocks-admin`? (isn't `program` only correct for luarocks itself?)

src/luarocks/admin/cmd/add.lua
```diff
@@ -14,7 +14,7 @@ local index = require("luarocks.admin.index")
 function add.add_to_parser(parser)
    local cmd = parser:command("add", "Add a rock or rockspec to a rocks server.",
       util.see_also())
-      :add_help(false)
+      :add_help("--help")
```
daurnimator: It this `--help` rather than `true` so that `-h` doesn't work?

pauloue: Yes, I could remove those lines if the default `-h --help` is good.

src/luarocks/cmd.lua
pauloue: It works for luarocks-admin since that's a separate binary.

src/luarocks/cmd.lua
pauloue: I don't think so since `--nodeps` is a flag while `--deps-mode` takes an argument.

src/luarocks/cmd/init.lua
pauloue: I changed this to get the options from `write_rockspec`.

src/luarocks/cmd/config.lua
pauloue: changed a bunch of occurrences of `flags` in `cmd.lua` too.

src/luarocks/cmd/install.lua
```diff
@@ -41,6 +41,10 @@ function install.add_to_parser(parser)
       "You need the signer’s public key in your local keyring for this "..
       "option to work properly.")
    util.deps_mode_option(cmd)
+   -- luarocks build options
+   parser:flag("--pack-binary-rock"):hidden(true)
```
daurnimator: TIL you can pass this to `install`.

src/luarocks/cmd.lua
```diff
@@ -306,6 +304,128 @@ do
...
+      :argname("<server>")
+   parser:option("--only-sources", "Restrict downloads to paths matching the given URL.")
+      :argname("<url>")
+   parser:option("--namespace"):description("Specify the rocks server namespace to use.")
+   parser:option("--lua-dir", "Which Lua installation to use.")
+      :argname("<prefix>")
+   parser:option("--lua-version", "Which Lua version to use.")
+      :argname("<ver>")
+   parser:option("--tree", "Which tree to operate on.")
+   parser:option("--project-tree"):description("") -- TODO
```
pauloue: Not sure what the description should be for this option.

src/luarocks/util.lua
```diff
@@ -77,162 +77,6 @@ function util.matchquote(s)
...
-   ["only-from"] = "<server>",
-   ["only-server"] = "<server>",
-   ["only-sources"] = "<url>",
-   ["only-sources-from"] = "<url>",
-   ["outdated"] = true,
-   ["output"] = "<file>",
-   ["pack-binary-rock"] = true,
-   ["porcelain"] = true,
-   ["project-tree"] = "<tree>",
-   ["quick"] = true,
```
pauloue: This `--quick` flag is the only one in this table not added to the argparse parser. It seems it was used by a `validate` command that no longer exists.

src/luarocks/cmd.lua
hishamhm: (I haven't looked properly at this branch yet, but I noticed this comment in a quick fly-by)

`--project-tree` is a slightly "internal" option, as it is used by the `./luarocks` wrapper created by `luarocks init` to basically force the use of the particular project tree, regardless of what the current directory is. Rationale:

* plain `luarocks` works like `git`: depending on your working directory, it works on a given project
* but if I have `~/project_a` and `~/project_b`, both initalized with `luarocks init`, and I do `cd ~/project_b; ../project_a/luarocks`, I would expect that wrapper to still refer to `project_a`.

(Having said that, as we are working out the kinks of running `luarocks` reliably in a `git`-style fashion, I'm getting less confident about the necessity of the `./luarocks` wrapper — I still think the `./lua` wrapper is more useful, if only to avoid having to deal with package paths (Lua package path errors are the kind of thing that are easy once you know them, but are a total mistery for newcomers))

src/luarocks/admin/cmd/make_manifest.lua
daurnimator: This inconsistency is still around (some places are changed over, some aren't).

src/luarocks/argparse.lua
```diff
@@ -0,0 +1,2013 @@
+-- The MIT License (MIT)
```
daurnimator: @hishamhm how would you like argparse brought in here? 
Vendoring it in at  src/luarocks/argparse.lua keeps working on luarocks itself as the same flow. But it's not obvious that this is a vendored dependency.

Other options:

  - vendor/argparse.lua (note that e.g. github knows not to count the top level vendor folder as part of a repository for stats purposes)
  - vendor/argparse/* as the whole argparse repository (brought in via `git subtree`)
  - vendor/argparse as a git submodule

src/luarocks/cmd/install.lua
```diff
@@ -208,51 +212,46 @@ local function install_rock_file(filename, opts)
...
 -- @return boolean or (nil, string, exitcode): True if installation was
 -- successful, nil and an error message otherwise. exitcode is optionally returned.
-function install.command(flags, name, version)
-   if type(name) ~= "string" then
-      return nil, "Argument missing. "..util.see_help("install")
-   end
-
-   name = util.adjust_name_and_namespace(name, flags)
+function install.command(args)
+   args.rock = util.adjust_name_and_namespace(args.rock, args)
```
daurnimator: I'm wondering if we should replace this with something in the argument parsing. Or alternatively just have the function return `namespace, name`.

hishamhm:
I also restarted the Appveyor job as there had been a hiccup in the CI when running it, and now both checks are green. :green_heart:

src/luarocks/cmd.lua
```diff
@@ -306,6 +304,141 @@ do
...
+   parser:flag("--local", "Use the tree in the user's home directory.\n"..
+      "To enable it, see '"..program.." help path'.")
+   parser:flag("--global", "Use the system tree when `local_by_default` is `true`.")
+   parser:flag("--verbose", "Display verbose output of commands executed.")
+   parser:option("--timeout", "Timeout on network operations, in seconds.\n"..
+      "0 means no timeout (wait forever). Default is "..
+      tostring(cfg.connection_timeout)..".")
+      :argname("<seconds>")
+      :convert(tonumber)
+   parser:option("-c"):description("Set the value of a config variable. "..
```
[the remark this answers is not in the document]
daurnimator: This was my idea:

> ```
> 2019-06-05 13:20:03     pauloue the luarocks VAR=VALUE argument to set config variables is a bit difficult to do with argparse
> 2019-06-05 13:22:09     pauloue it's optional so if an argument is in argparse, it will consume the subcommand
> 2019-06-05 13:26:39     daurnimator     I don't understand your statement there?
> 2019-06-05 13:29:29     pauloue if the main command has an optional argument and a command, then the command will always be interpreted as an argument
> 2019-06-05 13:30:06     pauloue because argparse doesn't know when to stop parsing the arguments since there can be any number of them
> 2019-06-05 13:30:22     daurnimator     I thought that those config variables could be done with: parser:argument("configuration", "configuration overrides in NAME=VALUE form"):count("*")
> 2019-06-05 13:31:45     pauloue yes but then I think if you do luarocks install, the install will be interpreted as an config argument
> 2019-06-05 13:32:17     daurnimator     oh I was thinking that you'd add that configuration argument to the install subcommand; after the install target
> 2019-06-05 13:32:58     pauloue oh that would work
> 2019-06-05 13:34:27     pauloue does it apply to any other commands as well?
> 2019-06-05 13:34:40     daurnimator     I think build, install and make?
> 2019-06-05 13:34:47     daurnimator     need to check
> 2019-06-05 13:39:33     pauloue do you think it makes sense to drop the boolean argument for add_help_command in argparse? since it's not enabled by default
> 2019-06-05 13:40:37     pauloue I did that for add_complete to avoid tracking the option index since it could change if the help optional is removed
> 2019-06-05 13:42:40     daurnimator     I don't really have an opinion on it
> 2019-06-05 13:45:06     pauloue I'll do it then since it's simpler and there's no need for add_help_command(false)
> 2019-06-05 13:45:45     daurnimator     okay
> 2019-06-05 13:46:34     daurnimator     okay I was wrong about those variables... they're used in all sorts of minor ways all over the place.... urgh
> 2019-06-05 13:47:06     daurnimator     I guess every subcommand gets a trailing :argument("variable"):count("*")
> 2019-06-05 13:49:03     daurnimator     okay maybe not all of them..... this needs some closer auditing
> 2019-06-05 13:49:52     pauloue some subcommands already have an optional argument, like the rockspec for make
> 2019-06-05 13:51:20     pauloue a workaround could be to add a separate option for each config variable to the main command
> 2019-06-05 13:51:30     pauloue options don't have to start with a -
> 2019-06-05 13:51:37     daurnimator     yeah. I like git's -C for this sort of thing much better
> 2019-06-05 13:52:01     daurnimator     ==> with git you can do: git -C myvar.mysubvar=myvalue subcommand --etc
> 2019-06-05 13:52:15     daurnimator     uh I mean -c
> 2019-06-05 13:52:26     pauloue yes that be the simplest way
> 2019-06-05 13:52:26     daurnimator     -C is to change directories :P
> 2019-06-05 13:52:51     daurnimator     Sadly that's not something we can change easily though :(
> 2019-06-05 13:53:22     daurnimator     You might have to do a pre-argparse pass to extract the variables
> 2019-06-05 13:53:52     daurnimator     that's sort of what already happens in the luarocks cmdline parsing
> 2019-06-05 13:54:10     daurnimator     https://github.com/luarocks/luarocks/blob/daf7e7e5f5ede8c3a5a00e2edaf8246209712781/src/luarocks/cmd.lua#L319
> 2019-06-05 13:54:59     pauloue ok I'll do that
> 2019-06-05 13:55:51     daurnimator     Towards the end of your GSoC time, if we get time, we could *also* add -c like git
> 2019-06-05 13:56:04     daurnimator     and say that the VAR=NAME anywhere is deprecated
> 2019-06-05 13:56:11     daurnimator     but I'd want to give it a year or two before we remove it
> 2019-06-05 13:57:05     pauloue ok
> ```

src/luarocks/cmd/doc.lua
```diff
@@ -12,19 +12,23 @@ local fetch = require("luarocks.fetch")
...
+      :summary("Show documentation for an installed rock.")
+      :add_help("--help")
+
+   cmd:argument("rock", "Name of the rock.")
+   cmd:argument("version", "Version of the rock.")
+      :args("?")
+
+   cmd:flag("--home", "Open the home page of project.")
+   cmd:flag("--list", "List documentation files only.")
+   cmd:flag("--porcelain"):description("Produce machine-friendly output.")
```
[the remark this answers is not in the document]
pauloue: They both do the same. I made it use the argument version consistently.

src/luarocks/admin/cmd/make_manifest.lua
hishamhm: AFAICT this is one of the only ones that could be changed. The other remaining ones are `args["local"]` since `local` is a keyword — so the inconsistency is unavoidable when using the `args.` form.

pauloue: I wasn't able to find any uses of `args["foo"]` other than `args["local"]`.

src/luarocks/cmd/test.lua
```diff
@@ -6,42 +6,42 @@ local cmd_test = {}
...
---test-type=<type>  Specify the test suite type manually if it was not
-                    specified in the rockspec and it could not be
-                    auto-detected.
+Any additional arguments are forwarded to the test suite.
+To make sure that test suite flags are not interpreted as LuaRocks flags, use --
+to separate LuaRocks arguments from test suite arguments.]],
+      util.see_also())
+      :summary("Run the test suite in the current directory.")
 
-]]..util.deps_mode_help()
```
pauloue: Not sure why the test command had the deps mode help (the option was unused).

pauloue:
When using `luarocks help` or `luarocks --help` argparse prints out the help message when we call `parser:parse`, which is before we call `init_config`, so the config text may be incomplete or wrong in this case.

Maybe the config should only be shown with `luarocks` and not with the help flag or command. Or luarocks could add it's own help flag and command that only print the help after cfg is inited (this is somewhat difficult to do for the help command). Another solution could be to add an option to argparse to not exit when using the help flag or command.

src/luarocks/cmd.lua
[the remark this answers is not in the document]
hishamhm: I have an issue with this feature, especially if it was made to mimic git:

In git, `-c foo=bar` is the temporary one-shot equivalent to setting `foo` to `bar` in `git config`.

But here, this is not equivalent to the same in `luarocks config`. It only applies to `luarocks config`'s `variables` subset. I don't like this inconsistency, I think it can be confusing/misleading.

I would rather remove the `"-c"` functionality for now from this PR, and consider bring it back later in the future as a `-c/--config` flag that applies to the entire configuration.

The current bare `VAR=VALUE` support was made to mimic Make, meaning `"$(VAR)"` value interpolations in the rockspec `build` section (and in retrospect should have allowed override via environment variables as well, as Make does).

hishamhm:
> Maybe the config should only be shown with luarocks and not with the help flag or command.

@pauloue That is a fair solution, especially if we add something like "run luarocks with no arguments to see the configuration" to the `--help` output.

src/luarocks/cmd.lua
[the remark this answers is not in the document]
daurnimator: Okay. Sorry @pauloue. Could you remove the `-c` support for now?

src/luarocks/cmd/install.lua
pauloue: `adjust_name_and_namespace` returns a name in the user/rock format. It might be better to keep the name and namespace separate to avoid splitting it again in e.g. `queries.new`. The argparse action callback could split the name and namespace, putting the name in `args.name` and the namespace in `args.namespace`.

daurnimator: Sounds like a good idea to me. But lets do that in another PR after this one is merged?

src/luarocks/manif/writer.lua
```diff
@@ -288,8 +288,6 @@ function writer.make_namespace_file(name, version, namespace)
    assert(type(name) == "string" and not name:match("/"))
    assert(type(version) == "string")
    assert(type(namespace) == "string" or not namespace)
-   name = util.adjust_name_and_namespace(name, { namespace = namespace })
-   name, namespace = util.split_namespace(name)
```
pauloue: Basad on https://github.com/luarocks/luarocks/commit/b6e96a3651db5ed27ec561a110610893e5421a59#diff-2286c8a0ccc4a1cf3148c26d4f2b7e71R287 I think the name argument isn't supposed to have a namespace.

src/luarocks/cmd.lua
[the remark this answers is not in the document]
pauloue: Dropped the commit that added it

hishamhm:
@pauloue I think this PR is good to go! @daurnimator looks like all suggested changes were addressed, can we dismiss the request for changes?

src/luarocks/argparse.lua
daurnimator: Tracked in #1067

src/luarocks/cmd/install.lua
daurnimator: Tracked in #1067

daurnimator:
@pauloue thank you for this PR and congratulations on completing your GSoC project!

teto:
You mentioned that argparse can generate shell completion files, how can I generate those for luarocks ?

pauloue:
See `luarocks help completion`. The argparse documentation is at https://argparse.readthedocs.io/en/stable/completions.html.

teto:
Thanks that worked perfectly. I've just added completion support to nixpkgs in https://github.com/NixOS/nixpkgs/pull/136507

Issue #1039, opened 2019-06-19
Cannot install packages with Visual Studio 2019


harrysummer:
I was using Visual Studio 2019 to build `luajit`. I successfully install LuaRocks. However, when I try to install luarocks packages, it shows:
```
  'mingw32-gcc' is not recognized as an internal or external command,
  operable program or batch file.
```

Looks like LuaRocks doesn't find any Visual Studio installation so it falls back to MinGW. After checking the code of LuaRocks, it seems getting Visual Studio path would fail for VS2019 because:
1. The runtime dll version no longer reflect the installed Visual Studio version.
2. The installation path of Visual Studio is no longer available in the registry.

I found [this blog](https://devblogs.microsoft.com/setup/vswhere-is-now-installed-with-visual-studio-2017/) from VS dev team. They now suggest to use `vswhere` as the tool to find the installation path of Visual Studio 2017 or later.

Here are some relevant log when installing LuaRocks:
```
  ========================
  == Checking system... ==
  ========================
...
  Headers found, checking runtime to use...
      D:/a/1/b/lua\luajit.exe uses VCRUNTIME140.DLL as runtime
  Runtime check completed.
  Looking for Microsoft toolchain matching runtime VCRUNTIME140 and architecture x86_64
      checking: HKLM\Software\Microsoft\VisualStudio\14.0\Setup\VC
      checking: HKLM\Software\Microsoft\VCExpress\14.0\Setup\VS
      Cannot auto-detect Windows SDK version from VCRUNTIME140
  
  ==========================
  == System check results ==
  ==========================
...
  Lua interpreter : D:/a/1/b/lua\luajit.exe
      binaries    : D:/a/1/b/lua
      libraries   : D:/a/1/b/lua
      includes    : D:/a/1/b/lua\include
      architecture: x86_64
      binary link : lua51.lib with runtime VCRUNTIME140.dll
  
  Compiler        : Microsoft (make sure it is in your path before using LuaRocks)
```

blueyed:
This is due to LuaRocks expecting to see `VCINSTALLDIR` in the environment: https://github.com/luarocks/luarocks/blob/3e12f3df50e4f6f0035f13b255d380fbf207a7c1/src/luarocks/core/cfg.lua#L608-L621 - changed in https://github.com/luarocks/luarocks/commit/760ad7c98a4a322b1f028ea9585cd17c5b74b2e2.

This looks confusing/wrong, and the value from `hardcoded.lua` should be used always probably.

[...]

PR #1040, opened 2019-06-19
Support VS tool chain 2017 or higher

harrysummer:
This PR solves #1039 by adding the logic of finding commandline tool of VS 2017 or higher using `vswhere` tool in function `get_msvc_env_setup_cmd`.

### DIFF
--- a/install.bat
+++ b/install.bat
@@ -549,7 +549,10 @@ local function get_msvc_env_setup_cmd()
 	local x64 = vars.UNAME_M=="x86_64"
 
 	-- 1. try visual studio command line tools of VS 2017 or higher
-	local vsdir = get_visual_studio_directory_from_vswhere()
+	local vsdir, err = get_visual_studio_directory_from_vswhere()
+	if err then
+		print("    Error when finding Visual Studio directory from vswhere: "..err)
+	end
 	if vsdir then
 		local vcvarsall = vsdir .. '\\VC\\Auxiliary\\Build\\vcvarsall.bat'
 		if exists(vcvarsall) then
--- a/install.bat
+++ b/install.bat
@@ -448,7 +448,7 @@ local function get_registry(key, value)
 	return nil
 end
 
-local function get_visual_studio_directory()
+local function get_visual_studio_directory_from_registry()
 	assert(type(vars.LUA_RUNTIME)=="string", "requires vars.LUA_RUNTIME to be set before calling this function.")
 	local major, minor = vars.LUA_RUNTIME:match('VCR%u*(%d+)(%d)$') -- MSVCR<x><y> or VCRUNTIME<x><y>
 	if not major then 
@@ -469,6 +469,47 @@ local function get_visual_studio_directory()
     end
 	end
 	return nil
+end
+
+local function get_visual_studio_directory_from_vswhere()
+	assert(type(vars.LUA_RUNTIME)=="string", "requires vars.LUA_RUNTIME to be set before calling this function.")
+	local major, minor = vars.LUA_RUNTIME:match('VCR%u*(%d+)(%d)$')
+	if not major then
+		print(S[[    Cannot auto-detect Visual Studio version from $LUA_RUNTIME]])
+		return nil
+	end
+	if tonumber(major) < 14 then
+		return nil
+	end
+	local program_dir = os.getenv('PROGRAMFILES(X86)')
+	if not program_dir then
+		return nil
+	end
+	local vswhere = program_dir.."\\Microsoft Visual Studio\\Installer\\vswhere.exe"
+	if not exists(vswhere) then
+		return nil
+	end
+	local f, msg = io.popen('"'..vswhere..'" -products * -property installationPath')
+	if not f then return nil, "failed to run vswhere: "..msg end
+	local vsdir = nil
+	while true do
+		local l, err = f:read()
+		if not l then
+			if err then
+				f:close()
+				return nil, err
+			else
+				break
+			end
+		end
+		vsdir = l
+	end
+	f:close()
+	if not vsdir then
+		return nil
+	end
+	print("    Visual Studio 2017 or higher found in: "..vsdir)
+	return vsdir
 end
 
 local function get_windows_sdk_directory()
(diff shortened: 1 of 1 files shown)


install.bat
```diff
@@ -471,6 +471,35 @@ local function get_visual_studio_directory()
...
 end
 
+local function get_visual_studio_directory_from_vswhere()
+	assert(type(vars.LUA_RUNTIME)=="string", "requires vars.LUA_RUNTIME to be set before calling this function.")
+	local major, minor = vars.LUA_RUNTIME:match('VCRUNTIME(%d+)(%d)$')
+	if not major or tonumber(major) < 14 then
+		print(S[[    Cannot auto-detect Visual Studio 2017 or higher from $LUA_RUNTIME]])
+		return nil
+	end
+	local vswhere_dir = os.getenv('PROGRAMFILES(X86)').."\\Microsoft Visual Studio\\Installer\\vswhere.exe"
```
daurnimator: Need to handle case that the env variable is not set.

install.bat
```diff
@@ -471,6 +471,35 @@ local function get_visual_studio_directory()
...
+	if not exists(vswhere_dir) then
+		return nil
+	end
+	local f = io.popen('"'..vswhere_dir..'" -products * -property installationPath')
+	if not f then return nil, "failed to run vswhere " end
+	local vsdir = nil
+	while true do
+		local l = f:read()
```
daurnimator: Should check for errors from `:read`

install.bat
daurnimator: It's not a `dir`, but the path to an executable, could you rename this variable?

harrysummer: Renamed, thanks!

harrysummer: handled when `os.getenv('PROGRAMFILES(X86)')` is nil

install.bat
harrysummer: What is the change you are propose here? `l` is already checked for nil on the following line of `read`.

daurnimator:
```lua
local l, err = f:read()
if not l then
    f:close()
    return nil, err
end
```

install.bat
```diff
@@ -507,8 +540,18 @@ local function get_msvc_env_setup_cmd()
...
 
-	-- 1. try visual studio command line tools
-	local vcdir = get_visual_studio_directory()
+	-- 1. try visual studio command line tools of VS 2017 or higher
+	local vsdir = get_visual_studio_directory_from_vswhere()
+	if vsdir then
+		local vcvarsall = vsdir .. '\\VC\\Auxiliary\\Build\\vcvarsall.bat'
+		if exists(vcvarsall) then
+			local vcvarsall_args = { x86 = "", x86_64 = " x64" }
+			return ('call "%s"%s'):format(vcvarsall, vcvarsall_args[vars.UNAME_M])
```
daurnimator: This line appears to have a windows line ending on it.

daurnimator: Could this be reachable with other architectures? (check if `vcvarsall_args[vars.UNAME_M]` is nil?)

install.bat
harrysummer: I'm afraid this might not be the correct logic. What I was trying to do, is going through the output of `vswhere` and set `vsdir` to the last line of the output (the higher version of Visual Studio appears at last). And when `l` becomes `nil`, it means eof is reached and the while loop should stop. It does not make sense to return nil just because of eof.

install.bat
harrysummer: I double-checked, the whole file has CRLF line ending. I guess this is by design because it is is bat file running on WIndows. Github shows `^M` on this line because it parses it wrong. You can see `^M` on line 536, 545, 719, 958 of master branch `install.bat` as well.

install.bat
daurnimator: Ah right, it doesn't always mean EOF; it can also mean that you're missing permissions to read the output for some reason; or some other weird error: it's useful to return that error message.
So it needs to me more like:
```lua
local l, err = f:read()
if not l then
    if err then
        f:close()
        return nil, err
    else
        break
    end
end
```
Also note that you need to `:close()` after the loop in the general case.

install.bat
daurnimator: Weird.
Should report a bug to github :smile:

harrysummer: fixed by adding an assert before the return.

harrysummer: Because it parses the file as bat....

install.bat
harrysummer: Fixed.

install.bat
```diff
@@ -471,6 +471,47 @@ local function get_visual_studio_directory()
...
+	local program_dir = os.getenv('PROGRAMFILES(X86)')
+	if not program_dir then
+		return nil
+	end
+	local vswhere = program_dir.."\\Microsoft Visual Studio\\Installer\\vswhere.exe"
+	if not exists(vswhere) then
+		return nil
+	end
+	local f = io.popen('"'..vswhere..'" -products * -property installationPath')
+	if not f then return nil, "failed to run vswhere " end
```
daurnimator: Can you append any error from `popen` here?

harrysummer: Sure. Code changed.

install.bat
```diff
@@ -471,6 +471,47 @@ local function get_visual_studio_directory()
...
 
+local function get_visual_studio_directory_from_vswhere()
+	assert(type(vars.LUA_RUNTIME)=="string", "requires vars.LUA_RUNTIME to be set before calling this function.")
+	local major, minor = vars.LUA_RUNTIME:match('VCR%u*(%d+)(%d)$')
+	if not major then
+		print(S[[    Cannot auto-detect Visual Studio version from $LUA_RUNTIME]])
+		return nil
+	end
+	if tonumber(major) < 14 then
+		return nil
```
daurnimator: Should these code paths have error messages?

install.bat
```diff
@@ -507,8 +548,19 @@ local function get_msvc_env_setup_cmd()
 	assert(type(vars.UNAME_M) == "string", "requires vars.UNAME_M to be set before calling this function.")
 	local x64 = vars.UNAME_M=="x86_64"
 
-	-- 1. try visual studio command line tools
-	local vcdir = get_visual_studio_directory()
+	-- 1. try visual studio command line tools of VS 2017 or higher
+	local vsdir = get_visual_studio_directory_from_vswhere()
```
daurnimator: Should report any error messages encountered by this function.

install.bat
harrysummer: Not sure but I prefer not. Every `return nil` could just be a normal case. For example `tonumber(major) < 14` could be true if user is using VS lower than 2015. `program_dir` could be empty if the system architecture is x86. `vswhere` could be not found if using VS2015 or an early version of VS2017. These are regular cases that luarocks might just want to fallback to other means of finding a VS installation.

install.bat
harrysummer: Added in commit https://github.com/luarocks/luarocks/pull/1040/commits/98f7f4d9b5287d5a9e76715f3039e1983fca1b23

install.bat
harrysummer: @daurnimator Do you suggest to keep the current code as-is, or we add additional error messages (that may bother pre-VS2017 user on Windows)?

hishamhm:
@daurnimator do you still have any concerns with this PR, can the request for changes be dismissed? I think we could merge it.

daurnimator:
> can the request for changes be dismissed?

done.

hishamhm:
@harrysummer This in now merged, thank you!! A new release should come shortly.

PR #1042, opened 2019-06-23
[RFC] deps: fix handling of luajit

blueyed:
The installer will only write `lua_interpreter` into the config, but not
`luajit_version`.

But `deps.check_lua` would use `cfg.luajit_version` before this patch to
handle luajit, resulting in:

  - not finding headers in the first place, and if providing them
    (`--with-lua-include=…/include/luajit-2.0`)
  - not finding the library (which cannot be specified then, since only
    the dir can be passed)

This patch checks `cfg.lua_interpreter == "luajit"` instead.

### DIFF
--- a/src/luarocks/deps.lua
+++ b/src/luarocks/deps.lua
@@ -545,7 +545,7 @@ function deps.check_lua(vars)
       "lua-" .. shortv,
       "lua",
    }
-   if cfg.lua_interpreter == "luajit" then
+   if cfg.luajit_version then
       table.insert(libnames, 1, "luajit-" .. cfg.lua_version)
    end
    local cache = {}
--- a/src/luarocks/deps.lua
+++ b/src/luarocks/deps.lua
@@ -545,7 +545,7 @@ function deps.check_lua(vars)
       "lua-" .. shortv,
       "lua",
    }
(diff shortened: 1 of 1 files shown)


blueyed:
See changes in the first commit: https://github.com/luarocks/luarocks/pull/1042/commits/fd46cb7680deebaa2d5dcca71b1622bd188209a5
(I've reverted it to test if CI is good)

The change by itself does not work for headers.
For this the luajit_version is required really (2.0 vs 2.1).
The best would be to save `luajit_version` into the config after all, but that involves doing so in the GNUMakefile and install.bat, so I'm skipping this until getting some feedback first.

blueyed:
As for GNUMakefile: it uses luarocks-init therein, which adds `luajit_version` to the config, but this then gets overwritten from the one generated in `build`.  See https://github.com/luarocks/luarocks/pull/1043.

blueyed:
Should be not necessary anymore with https://github.com/luarocks/luarocks/pull/1051.

blueyed:
But from https://github.com/luvit/luv/pull/344#issuecomment-512594969 it might be that "luarocks init" is still not working?!

PR #1043, opened 2019-06-23
[WIP/RFC] GNUmakefile: do not overwrite config from luarocks-init

blueyed:
This would cause the luajit_version stored therein to be overwritten.


blueyed:
For Windows the `install.bat` does not use `luarocks-init` , and does not even write `lua_version` to the config (https://github.com/blueyed/luarocks/blob/c40631c607f623e426609d0d3a158fc6a359de01/install.bat#L1061-L1103).  There it likely comes via `hardcoded.lua` then or something similar.

blueyed:
Closing in favor of https://github.com/luarocks/luarocks/pull/1051.

PR #1051, opened 2019-07-12
detect LuaJIT dynamically

hishamhm:
This reduces the complexity of the interaction between build-time configuration, run-time auto-detection, and CLI flag overrides (which, all together, make for a pretty confusing support matrix). The LuaJIT version is now always auto-detected at run-time based on the Lua interpreter currently configured, based on the values of configuration options `variables.LUA_BINDIR` and `lua_interpreter`.

This is proposed as an alternative to #1043, #1042, #883. 

@blueyed @fperrad Please let me know if this approach works for you!

### DIFF
--- a/src/luarocks/util.lua
+++ b/src/luarocks/util.lua
@@ -588,14 +588,25 @@ do
       if luaver and luaver ~= lv then
          return nil
       end
+      return lv
+   end
+
+   function util.get_luajit_version()
+      local cfg = require("luarocks.core.cfg")
+      if cfg.cache.luajit_version_checked then
+         return cfg.cache.luajit_version 
+      end
+      cfg.cache.luajit_version_checked = true
+
       local ljv
-      if lv == "5.1" then
-         ljv = util.popen_read(Q(lua_exe) .. ' -e "io.write(tostring(jit and jit.version:sub(8)))"')
+      if cfg.lua_version == "5.1" then
+         ljv = util.popen_read(Q(cfg.variables["LUA_BINDIR"] .. "/" .. cfg.lua_interpreter) .. ' -e "io.write(tostring(jit and jit.version:sub(8)))"')
          if ljv == "nil" then
             ljv = nil
          end
       end
-      return lv, ljv
+      cfg.cache.luajit_version = ljv
+      return ljv
    end
 
    local find_lua_bindir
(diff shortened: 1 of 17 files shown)


blueyed:
Thanks!
Will give this a try.

blueyed:
I can confirm that it fixes `make bootstrap` (on Linux).
btw: is `make bootstrap` better than `make install`?  (we use it in Neovim for bundled deps, and `make install` appears to work as good when I've used it as a workaround (it was using `make bootstrap` before).  Has `bootstrap` the benefit of allowing for easier uninstall?

src/luarocks/util.lua
```diff
@@ -587,14 +587,25 @@ do
...
+      if cfg.cache.luajit_version_checked then
+         return cfg.cache.luajit_version 
+      end
+      cfg.cache.luajit_version_checked = true
+
       local ljv
-      if lv == "5.1" then
-         ljv = util.popen_read(Q(lua_exe) .. ' -e "io.write(tostring(jit and jit.version:sub(8)))"')
+      if cfg.lua_version == "5.1" then
+         ljv = util.popen_read(Q(cfg.variables["LUA_BINDIR"] .. "/" .. cfg.lua_interpreter) .. ' -e "io.write(tostring(jit and jit.version:sub(8)))"')
```
blueyed: Shouldn't/couldn't this be done always (regardless of `lua_version`)?

blueyed:
btw: would be nice to have this rebased when you update it to have the install.bat fix in it also.

blueyed:
Can you check https://github.com/luvit/luv/pull/344#issuecomment-512594969 ?  (this PR/branch was tested there)
It might be that "luarocks init" is still not working properly.

src/luarocks/util.lua
hishamhm: LuaRocks generally assumes that LuaJIT implies Lua 5.1, to avoid trouble with different LuaJITs compiled with different levels of Lua 5.x compatibility.

hishamhm:
> Can you check luvit/luv#344 (comment) ?  (this PR/branch was tested there) It might be that "luarocks init" is still not working properly.

Thank you for testing! I will take a look at this tomorrow.

blueyed:
Thanks!

Can you shed some light on "make bootstrap" vs "make install"?

> btw: is `make bootstrap` better than `make install`? (we use it in Neovim for bundled deps, and `make install` appears to work as good when I've used it as a workaround (it was using `make bootstrap` before). Has `bootstrap` the benefit of allowing for easier uninstall?

hishamhm:
Some people expected LuaRocks to be able to upgrade itself, as in `luarocks
install luarocks`. For those users, "make bootstrap" installs LuaRocks and
registers itself as a rock, which can then be later upgraded like any other
rock (save for fundamental breaking changes like the switch from LuaRocks 2
to 3).

For installations of LuaRocks done via a package manager, it should _not_
be able to self-upgrade: the system package management has ownership and
control over the files that constitute LuaRocks, so LuaRocks should not
modify them. Hence, the regular "make; make install" installs LuaRocks but
doesn't register LuaRocks itself as a rock.

>

PR #1054, opened 2019-07-15
Don't use project dir or default_lua_version with --force-config

### LINKED ISSUE #1045: ./configure --force-config: should ignore project config/dirs?
I am using `./configure --prefix=… --force-config`.

However, in 3.1.3, even though `--force-config` is used, it will use another tree, and especially interpreter - apparently through the new project mode / config handling.

(With 3.1.3 I am using `--lua-interpreter=luajit`, with 2.4.4 `--suffix=jit` (so here the missing `luajit_version` in the config might be an issue also, but likely not (https://github.com/luarocks/luarocks/pull/1042)).)

LuaRocks 2.4.4 displays with `luarocks` then:
```
CONFIGURATION
Lua version: 5.1                                                                                                                                                Configuration files:
System: …/Vcs/neovim/.deps/usr/etc/luarocks/config-5.1.lua (ok)
User  : disabled in this LuaRocks installation.
```

LuaRocks 3.1.3 however:
```
CONFIGURATION
        Lua version: 5.3

        Configuration files:
                System  : …/Vcs/neovim/.deps/usr/etc/luarocks/config-5.3.lua (not found)
                User    : /home/user/.luarocks/config-5.3.lua (not found)
                Project : …/Vcs/neovim/./.luarocks/config-5.3.lua (ok)

        Rocks trees in use:
                …/Vcs/neovim/./lua_modules ("project")
                /home/user/.luarocks ("user")
```

This seems to be caused by a `.luarocks` dir in the current dir, but also used when inside `.deps:
```
% grep -r . .luarocks
.luarocks/default-lua-version.lua:return "5.3"

.luarocks/config-5.3.lua:lua_interpreter = "lua5.3"
.luarocks/config-5.3.lua:variables = {
.luarocks/config-5.3.lua:   LUA_BINDIR = "/usr/bin",
.luarocks/config-5.3.lua:   LUA_DIR = "/usr",
.luarocks/config-5.3.lua:   LUA_INCDIR = "/usr/include",
.luarocks/config-5.3.lua:   LUA_LIBDIR = "/usr/lib"
.luarocks/config-5.3.lua:}
```

The use case here is to have an isolated rock tree to be used with a specific interpreter only.
So any user config / project dirs should be ignored.

I've tried passing in `--tree` explicitly with `luarocks build`, but that also requires `--with-lua` then, and would not support an explicit interpreter AFAICS.
So the `luarocks` wrapper that is being called should just honor the configured interpreter/config.

`.deps/usr/bin/luarocks`:
```sh
% head .deps/usr/bin/luarocks
#!/bin/sh

LUAROCKS_SYSCONFDIR='…/Vcs/neovim/.deps/usr/etc/luarocks' exec '/usr/bin/lua5.1' -e 'package.path="…/Vcs/neovim/.deps/usr/share/lua/5.1/?.lua;…/Vcs/neovim/.deps/usr/share/lua/5.1/?/init.lua;"..package.path;package.cpath="…/Vcs/neovim/.deps/usr/lib/lua/5.1/?.so;"..package.cpath;local k,l,_=pcall(require,"luarocks.loader") _=k and l.add_context("luarocks","3.1.3-1")' '…/Vcs/neovim/.deps/usr/lib/luarocks/rocks-5.1/luarocks/3.1.3-1/bin/luarocks' "$@"
```

Due to using `Lua version: 5.3`, it even says `System  : …/Vcs/neovim/.deps/usr/etc/luarocks/config-5.3.lua (not found)` then (although there is `…/Vcs/neovim/.deps/usr/etc/luarocks/config-5.1.lua`.

(the `.luarocks` dir was created not intentionally, by running `luarocks init` manually IIRC).

hishamhm:
Fixes #1045.

### DIFF
--- a/src/luarocks/cmd.lua
+++ b/src/luarocks/cmd.lua
@@ -13,6 +13,11 @@ local dir = require("luarocks.dir")
 local fun = require("luarocks.fun")
 local fs = require("luarocks.fs")
 
+local hc_ok, hardcoded = pcall(require, "luarocks.core.hardcoded")
+if not hc_ok then
+   hardcoded = {}
+end
+
 local program = util.this_program("luarocks")
 
 cmd.errorcodes = {
@@ -189,6 +194,10 @@ do
       end
    
       local function find_default_lua_version(flags, project_dir)
+         if hardcoded.FORCE_CONFIG then
+            return nil
+         end
+
          local dirs = {}
          if project_dir then
             table.insert(dirs, dir.path(project_dir, ".luarocks"))
--- a/src/luarocks/core/cfg.lua
+++ b/src/luarocks/core/cfg.lua
@@ -561,7 +561,7 @@ function cfg.init(detected, warning)
    local lua_incdir = detected.lua_incdir or hardcoded.LUA_INCDIR
    local lua_libdir = detected.lua_libdir or hardcoded.LUA_LIBDIR
    local lua_dir = detected.lua_dir or hardcoded.LUA_DIR or (lua_bindir and lua_bindir:gsub("[\\/]bin$", ""))
-   local project_dir = detected.project_dir
+   local project_dir = (not hardcoded.FORCE_CONFIG) and detected.project_dir
    
    local init = cfg.init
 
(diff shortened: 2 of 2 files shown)


blueyed:
Thanks for addressing it.
Might be good to have a test for this?

hishamhm:
> Might be good to have a test for this?

The test suite doesn't currently have integration tests that test for different configure-time flag combinations. :-\ It would be a bit of an undertaking to add that...

Issue #1049, opened 2019-07-03
Error: "touch: invalid date format '198000000000.00'"


rosik:
## Reproducing

1. Provide `zlib` lua module, so luarocks can use it
1. Make sure `luafilesystem` isn't installed 
1. Get any rock and pack it to `.all.rock` (or platform-specific, it doesn't matter)
1. Install it: `luarocks install mymodule.all.rock`

You will see errors in console:
```txt
touch: invalid date format '198000000000.00'
```

Because month and day can't be zero.

## Source of problems

During `luarocks pack`:

1.  [luarocks.tools.zip](https://github.com/luarocks/luarocks/blob/master/src/luarocks/tools/zip.lua#L93-L108) generates `timestamp = 0`
```
   lfh.last_mod_file_time = 0 -- TODO
   lfh.last_mod_file_date = 0 -- TODO
```

During `luarocks install`

1. The rock is unzipped
2. function [process_central_dir](https://github.com/luarocks/luarocks/blob/master/src/luarocks/tools/zip.lua#L413) reads `0` from zip headers,
3. Function [ziptime_to_luatime](https://github.com/luarocks/luarocks/blob/master/src/luarocks/tools/zip.lua#L308-L317) converts it to `{year = 1980, month = 0, day = 0, ...}`
4. In absence of lfs module unzip [calls](https://github.com/luarocks/luarocks/blob/master/src/luarocks/tools/zip.lua#L499) `touch -t 19800000...` 

That's it.

daurnimator:
Is the underlying issue that `.day` and `.month` need to have `1` added?

rosik:
Ziptime already has that offset. Only `0 -> 1` should be replaced.

rosik:
What do you think about fix in our fork? https://github.com/tarantool/luarocks/pull/8/files 

If it's appropriate, we can cherry-pick it and open PR in upstream too. Or maybe there is a better way to fix the problem? Also, should it be covered with tests (is it ever possible)?

daurnimator:
Perhaps a better solution might be to set the date to invalid/the epoch if any of the values are out of range?
Looking at various zip files on my system, most of them use the file stamp extension (0x5544)

hishamhm:
> If it's appropriate, we can cherry-pick it and open PR in upstream too.

@rosik Please do!

[...]

PR #1056, opened 2019-07-16
Fix zero values in ziptime

vanyarock01:
If the rock was packed with luarocks then the time in it is set to zeros
and this is not the case with the unix format can be removed when date given is added to zipwriter_open_new_file_in_zip


vanyarock01:
Resolve #1049

hishamhm:
@vanyarock01 Merged, thank you!!

PR #1066, opened 2019-08-21
support RaptorJIT

fperrad:
RaptorJIT is a fork of LuaJIT, where jit.version = 'RaptorJIT 1.0.0'

this fix avoids a failure in find_lua_incdir()
    luajitver and prefix .. "/include/luajit-" .. luajitver:match("^(%d+%.%d+)"),
    --> attempt to concatenate a nil value


hishamhm:
@fperrad Thanks!

PR #1068, opened 2019-08-26
Fix build --no-doc

pauloue:
I missed this in #1035.
Note that `build --pack-binary-rock --no-doc` doesn't work (and it never did). I'll submit a fix for that soon.

### DIFF
--- a/src/luarocks/cmd/build.lua
+++ b/src/luarocks/cmd/build.lua
@@ -30,6 +30,7 @@ function cmd_build.add_to_parser(parser)
       :args("?")
 
    cmd:flag("--only-deps", "Installs only the dependencies of the rock.")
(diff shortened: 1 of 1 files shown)


src/luarocks/cmd/build.lua
```diff
@@ -30,6 +30,7 @@ function cmd_build.add_to_parser(parser)
       :args("?")
 
    cmd:flag("--only-deps", "Installs only the dependencies of the rock.")
+   cmd:flag("--no-doc", "Installs the rock without its documentation.")
```
daurnimator: Should this be part of `make.cmd_options`?

pauloue: `make` didn't previously support `--no-doc` but it probably should. I'll have to move `remove_doc_dir` to util.

PR #1079, opened 2019-09-02
build: fix --branch flag

### LINKED ISSUE #1077: build/make --branch option doesn't work
In `buid.opts` the branch is a boolean: https://github.com/luarocks/luarocks/blob/master/src/luarocks/build.lua#L21. I tried making it a string, but I don't think that fixes it.

hishamhm:
The `--branch` flag is optional and takes a string argument. This fixes the propagation of the string value.

The `--branch` flag does not make sense for `luarocks make` because it does not fetch sources, it builds/installs based on whatever is in the current directory.

This also adds tests that verify the behavior, but these don't run in Travis CI because of issues running a Git daemon there. They were verified locally.

Fixes #1077.

### DIFF
--- a/src/luarocks/cmd/install.lua
+++ b/src/luarocks/cmd/install.lua
@@ -42,7 +42,7 @@ function install.add_to_parser(parser)
    util.deps_mode_option(cmd)
    -- luarocks build options
    parser:flag("--pack-binary-rock"):hidden(true)
-   parser:flag("--branch"):hidden(true)
+   parser:option("--branch"):hidden(true)
    parser:flag("--sign"):hidden(true)
 end
 
--- a/spec/build_spec.lua
+++ b/spec/build_spec.lua
@@ -4,6 +4,7 @@ local get_tmp_path = test_env.get_tmp_path
 local run = test_env.run
 local testing_paths = test_env.testing_paths
 local write_file = test_env.write_file
+local git_repo = require("spec.util.git_repo")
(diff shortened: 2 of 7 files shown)


pauloue:
Tested with
```
luarocks build https://raw.githubusercontent.com/luarocks/luarocks/master/luarocks-dev-1.rockspec --branch 3.2.0-fixes
```
It works.

src/luarocks/cmd/install.lua
```diff
@@ -42,7 +42,7 @@ function install.add_to_parser(parser)
    util.deps_mode_option(cmd)
    -- luarocks build options
    parser:flag("--pack-binary-rock"):hidden(true)
-   parser:flag("--branch"):hidden(true)
+   parser:option("--branch"):argname("<name>"):hidden(true)
```
pauloue: The argname doesn't matter since it's hidden.

pauloue:
There's also a `--branch` test here: https://github.com/luarocks/luarocks/blob/master/spec/build_spec.lua#L112.
I don't think it makes sense in that context since that command will not use git.

src/luarocks/cmd/install.lua
hishamhm: :+1: !

hishamhm:
@pauloue good catch on the dummy test there, I removed it since we have new tests to cover that now.

PR #1082, opened 2019-09-03
Fix `luarocks build --pack-binary-rock --no-doc`

pauloue:
Use the name and version returned by `do_build` instead of the name and version passed on the command line (the name might be namespaced and the version argument is not required).


hishamhm:
Merged! I noticed we don't have tests for `luarocks build --pack-binary-rock` at all — we should probably add something there to avoid a regression like this.

Issue #1067, opened 2019-08-24
Refact adjust_name_and_namespace 


daurnimator:
`adjust_name_and_namespace` returns a name in the user/rock format. It might be better to keep the name and namespace separate to avoid splitting it again in e.g. `queries.new`. The argparse action callback could split the name and namespace, putting the name in `args.rock` and the namespace in `args.namespace`.

_Originally posted by @pauloue in https://github.com/luarocks/luarocks/pull/1035_

PR #1091, opened 2019-09-13
Refactor command line handling of namespaces

pauloue:
https://github.com/luarocks/luarocks/issues/1067

Rock name and namespace are always passed around separately to avoid splitting them multiple times.

### DIFF
--- a/src/luarocks/cmd/write_rockspec.lua
+++ b/src/luarocks/cmd/write_rockspec.lua
@@ -263,9 +263,7 @@ local function rockspec_cleanup(rockspec)
 end
 
 function write_rockspec.command(args)
-
-   local name = util.adjust_name_and_namespace(args.name, args)
-   local version = args.version
+   local name, version = args.name, args.version
    local location = args.location
 
    if not name then
--- a/src/luarocks/util.lua
+++ b/src/luarocks/util.lua
@@ -215,6 +215,10 @@ function util.this_program(default)
    return prog
 end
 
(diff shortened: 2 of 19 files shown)


src/luarocks/cmd/write_rockspec.lua
```diff
@@ -67,6 +67,7 @@ rockspec, and is not guaranteed to be complete or correct. ]], util.see_also())
 
    cmd:argument("name", "Name of the rock.")
       :args("?")
+      :action(util.namespaced_name_action)
```
pauloue: Does write_rockspec even need to accept a namespace?

src/luarocks/util.lua
```diff
@@ -373,6 +353,21 @@ function util.split_namespace(ns_name)
    return ns_name
 end
 
+--- Argparse action callback for namespaced rock arguments.
+function util.namespaced_name_action(args, target, ns_name)
```
daurnimator: This looks a lot like `adjust_name_and_namespace`: could you modify the behaviour of that function?
Also e.g. keep the `assert`s?

src/luarocks/cmd/write_rockspec.lua
hishamhm: I don't think so, I think that was just blanket-added to every command that used a `name`...

hishamhm:
@pauloue I fixed the conflicts and merged manually. Thank you for this!

PR #1092, opened 2019-09-13
Support make --no-doc

pauloue:
Same as build --no-doc. Also removes some code duplication.


hishamhm:
@pauloue Could you add an entry in `make_spec.lua` similar to the one we have in `install_spec.lua` to cover `make --no-doc` in tests? Thanks!

hishamhm:
@pauloue Merged!!

PR #1113, opened 2019-11-13
Fix commands with a dash

### LINKED ISSUE #1103: luarocks-admin --local-tree argument
```
$ luarocks-admin make-manifest 
Making manifest for /usr/lib/luarocks/rocks-5.3
Warning: This looks like a local rocks tree, but you did not pass --local-tree.

Error: Cannot create file at /usr/lib/luarocks/rocks-5.3/manifest.tmp
```
```
$ luarocks-admin make-manifest --local-tree
Usage: luarocks-admin make-manifest [-h]

Error: unknown option '--local-tree'
```
```
$ luarocks-admin --local-tree make-manifest 
Usage: luarocks-admin [-h] [--version] [--dev] [--server <server>]
       [--only-server <server>] [--only-sources <url>]
       [--namespace <namespace>] [--lua-dir <prefix>]
       [--lua-version <ver>] [--tree <tree>] [--local] [--global]
       [--verbose] [--timeout <seconds>] [<command>] ...

Error: unknown option '--local-tree'
```
```
$ luarocks-admin make-manifest -h
Usage: luarocks-admin make-manifest [-h]

Options:
   -h, --help            Show this help message and exit.
```

pauloue:
Fixes #1103 
It's a bit hacky but if I used aliases it would show both versions in help messages.

### DIFF
--- a/src/luarocks/admin/cmd/make_manifest.lua
+++ b/src/luarocks/admin/cmd/make_manifest.lua
@@ -13,14 +13,17 @@ local dir = require("luarocks.dir")
 
 function make_manifest.add_to_parser(parser)
    local cmd = parser:command("make_manifest", "Compile a manifest file for a repository.", util.see_also())
-   parser:command("make-manifest"):hidden(true):action(function(args) args.command = "make_manifest" end)
+   local cmd2 = parser:command("make-manifest"):hidden(true)
+      :action(function(args) args.command = "make_manifest" end)
 
-   cmd:argument("repository", "Local repository pathname.")
-      :args("?")
+   for _, cmd in ipairs({cmd, cmd2}) do
+      cmd:argument("repository", "Local repository pathname.")
+         :args("?")
 
-   cmd:flag("--local-tree", "If given, do not write versioned versions of the manifest file.\n"..
-      "Use this when rebuilding the manifest of a local rocks tree.")
-   util.deps_mode_option(cmd)
+      cmd:flag("--local-tree", "If given, do not write versioned versions of the manifest file.\n"..
+         "Use this when rebuilding the manifest of a local rocks tree.")
+      util.deps_mode_option(cmd)
+   end
 end
 
 --- Driver function for "make_manifest" command.
--- a/src/luarocks/admin/cmd/refresh_cache.lua
+++ b/src/luarocks/admin/cmd/refresh_cache.lua
@@ -8,11 +8,14 @@ local cache = require("luarocks.admin.cache")
 
 function refresh_cache.add_to_parser(parser)
    local cmd = parser:command("refresh_cache", "Refresh local cache of a remote rocks server.", util.see_also())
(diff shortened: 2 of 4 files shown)


daurnimator:
Would this make more sense to be fixed in argparse?

pauloue:
That would mean adding support for hidden aliases. Maybe an alias could be marked as hidden by surrounding with with parentheses: `parser:command "new_version (new-version)"`.

daurnimator:
> That would mean adding support for hidden aliases. Maybe an alias could be marked as hidden by surrounding with with parentheses: `parser:command "new_version (new-version)"`.

I was more thinking that anything with `-` or `_` in the middle could be happy to take either (considering that we already convert subcommands/options with `-` into `_` form keys inside of the arguments table)

hishamhm:
I agree that making argparse support both form sout-of-the-box would make the most sense. I know it's a bit opinionated in principle, but a CLI command where subcommands `foo-bar` and `foo_bar` mean different things would be a super footgun.

hishamhm:
@pauloue the argparse change could be a setting to the parser as a whole (`parser:accept_dashed_commands()` or something to that effect). How does that sound to you?

@daurnimator if this doesn't make it as an argparse feature, I'm inclined to merge this PR, hacky as it is, to get the issue fixed on the LuaRocks side at least.

pauloue:
To make this work correctly in argparse for both parsing and the argument to the `help` command I can't think of any reasonable way other than by supporting hidden aliases. An alias could be added automatically by argparse, or there could be some way of manually adding a hidden alias like I suggested before. This could also be useful for e.g. deprecated command names.

hishamhm:
> An alias could be added automatically by argparse, or there could be some way of manually adding a hidden alias like I suggested before. This could also be useful for e.g. deprecated command names.

Either solution sounds good to me!

hishamhm:
Wouldn't this be sufficient or is it too much of a hack? Just make commands always accept `-` instead of `_`:

```
diff --git a/src/luarocks/argparse.lua b/src/luarocks/argparse.lua
index dc6cdb0d..48c7b1dd 100644
--- a/src/luarocks/argparse.lua
+++ b/src/luarocks/argparse.lua
@@ -1849,7 +1849,7 @@ function ParseState:get_option(name)
 end
 
 function ParseState:get_command(name)
-   local command = self.commands[name]
+   local command = self.commands[name:gsub("%-", "_")]
 
    if not command then
       if #self.commands > 0 then
```

pauloue:
A similar hack would be needed in `add_help_command` to make the argument to the help command work if a `-` is used when the command has a `_`.
I opened https://github.com/luarocks/argparse/pull/13 which allows adding an hidden alias with e.g. `command:hidden_name("new-version")`.

hishamhm:
I just tested vendoring https://github.com/luarocks/argparse/pull/13 into LuaRocks instead and it worked fine! Let's push that instead into argparse and LuaRocks and close this one. Thank you @pauloue for the nice fix!

PR #1115, opened 2019-11-15
Add --check-lua-versions flag

hishamhm:
Only check for availability of a rock for other Lua versions if the user explicitly requests so.

### DIFF
--- a/src/luarocks/search.lua
+++ b/src/luarocks/search.lua
@@ -205,17 +205,14 @@ end
 
 -- Find out which other Lua versions provide rock versions matching a query,
 -- @param query table: a query object.
--- @param cli boolean: print status messages as it works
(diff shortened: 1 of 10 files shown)


src/luarocks/search.lua
```diff
@@ -225,71 +222,84 @@ local function supported_lua_versions(query, cli)
...
+      local add
+      if check_lua_versions then
+         util.printout(query.name .. " not found for Lua " .. cfg.lua_version .. ".")
+         util.printout("Checking if available for other Lua versions...")
       
          -- Check if constraints are satisfiable with other Lua versions.
-         local lua_versions = supported_lua_versions(query, cli)
-   
+         local lua_versions = supported_lua_versions(query)
+
```
daurnimator: whitespace here.

PR #1116, opened 2019-11-19
support moonjit

fperrad:
moonjit is a fork of LuaJIT, where jit.version = 'moonjit 2.1.1'

### DIFF
--- a/src/luarocks/util.lua
+++ b/src/luarocks/util.lua
@@ -454,7 +454,7 @@ do
 
       local ljv
       if cfg.lua_version == "5.1" then
-         ljv = util.popen_read(Q(cfg.variables["LUA_BINDIR"] .. "/" .. cfg.lua_interpreter) .. ' -e "io.write(tostring(jit and jit.version:gsub([[^%w-JIT ]], [[]])))"')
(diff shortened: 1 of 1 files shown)


src/luarocks/util.lua
```diff
@@ -454,7 +454,7 @@ do
 
       local ljv
       if cfg.lua_version == "5.1" then
-         ljv = util.popen_read(Q(cfg.variables["LUA_BINDIR"] .. "/" .. cfg.lua_interpreter) .. ' -e "io.write(tostring(jit and jit.version:gsub([[^%w-JIT ]], [[]])))"')
+         ljv = util.popen_read(Q(cfg.variables["LUA_BINDIR"] .. "/" .. cfg.lua_interpreter) .. ' -e "io.write(tostring(jit and jit.version:gsub([[^%w+ ]], [[]])))"')
```
daurnimator: Does this work? `%w` is alphanumeric and won't include `-`, `_` and any other punctuation we may expect here

hishamhm: It will for now, while we don't have another LuaJIT fork with dashes :) But `%S` should work too.

(I clicked "resolved" by accident from the phone and apparently Github's mobile site doesn't include "unresolve", sorry about that)

fperrad:
reworked with `%S`

hishamhm:
@fperrad Merged, thank you!

PR #1122, opened 2019-12-14
git: add support for shallow recommendations

tammela:
In more recent versions of git, the .gitmodules file can
be configured to recommend if a submodule should be cloned as
shallow or not. Add support for this feature since it gives the
rock maintainer more control over which submodules should be shallow
cloned.

The rock maintainer may set the boolean `shallow` to true or false in
.gitmodules. If omitted it is treated as false.

Signed-off-by: Pedro Tammela <pctammela@gmail.com>

### DIFF
--- a/src/luarocks/fetch/git.lua
+++ b/src/luarocks/fetch/git.lua
@@ -50,6 +50,13 @@ local function git_supports_shallow_submodules(git_cmd)
    return git_is_at_least(git_cmd, "1.8.4")
 end
 
+--- Git >= 2.10 can fetch submodules shallowly according to .gitmodules configuration, allowing more granularity.
+-- @param git_cmd string: name of git command.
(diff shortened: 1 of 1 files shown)


daurnimator:
Don't we always want shallow though? Why would we want to follow the recommendation?

tammela:
> Don't we always want shallow though? Why would we want to follow the recommendation?

Enforcing shallow clones on submodules may break the rock if it relays on old versions. Increasing the depth of the shallow clone is not a good solution since it is enforced in all submodules, also it does not solve the problem I've mentioned, just minimizes it.

Therefore it should be the git's repository decision which submodules should be shallow cloned.

hishamhm:
I think always shallow cloning should be sufficient for our purposes, but following the repo's recommendation doesn't hurt.

hishamhm:
Thanks @tammela, merged!!

PR #1125, opened 2019-12-31
Windows on ARM support

soapdog:
This PR adds support for using Luarocks from a Windows on ARM machine. These machines use Visual Studio running under their built-in x86-32 emulation to cross-compile into ARM64 native binaries.

I've patched `install.bat` to detect when it is running under ARM64 and select the appropriate cross-compiler.

I've successfully used it to install some rocks such as LuaFileSystem, busted, lpeg, so I think it is working. There were some failures but mostly due to C code that assumes UNIX and tries to include header files which are not present on standard Visual Studio installs.

### DIFF
--- a/install.bat
+++ b/install.bat
@@ -399,6 +399,8 @@ end
 local function get_runtime()
 	local f
 	vars.LUA_RUNTIME, f = pe.msvcrt(vars.LUA_BINDIR.."\\"..vars.LUA_INTERPRETER)
+	print("msvcrt: " .. f)
+	print("vars.LUA_RUNTIME" .. vars.LUA_RUNTIME)
 	if type(vars.LUA_RUNTIME) ~= "string" then
 		-- analysis failed, issue a warning
 		vars.LUA_RUNTIME = "MSVCR80"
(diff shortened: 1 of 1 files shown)


install.bat
```diff
@@ -571,9 +574,12 @@ local function get_msvc_env_setup_cmd()
 			x86_64 = {
 				"bin\\amd64\\vcvars64.bat", -- prefers native compiler
 				"bin\\x86_amd64\\vcvarsx86_amd64.bat" -- then cross compiler
+			},
+			arm64 = {
+				"bin\\x86_arm64\\vcvarsx86_arm64.bat" -- need to use cross compiler"
```
daurnimator: mixed line ending here

install.bat
```diff
@@ -415,9 +415,12 @@ local function get_architecture()
 	if not proc then
 		die("Could not detect processor architecture used in "..vars.LUA_INTERPRETER)
 	end
+	print("arch: " .. proc .. " -> " .. pe.const.Machine[proc] )
```
daurnimator: trailing whitespace

install.bat
soapdog: oh shit, I forgot I am on windows... sorry. Will fix.

install.bat
soapdog: fixed.

install.bat
soapdog: fixed.

hishamhm:
@soapdog Merged manually! (I ran `unix2dos` so that the diff wasn't the whole file)
Thanks @daurnimator for the review!

Issue #1135, opened 2020-01-21
Testing new compiler/toolchain on Windows 10


RussellHaley:
Hi,

I've found a mad Finn that has created a [clang/mingw](https://github.com/mstorsjo/llvm-mingw) hybrid based on the clang toolchian. I should be able to get the entire thing down to about 440 Mb and it doesn't require nasty vsvars.bat or runtime detection.  I've been able to compile a small `sleep` module and use it with Lua. 

I'd like to add support to Luarocks but I'm not sure where to start? Can I just specify my toolchain in a config file? If not, I don't mind working with the standard package instead of the all-in-one for my testing if there are some files that need to be manipulated?

Thanks!
Russell

RussellHaley:
Okay, so I dug into src/luarocks/build/builtin.lua and found the variables I needed to replace the compiler and linker values in the configuration file. It appears if I just don't specify `MSVCRT = 'VCRUNTIME140'` in the config file, then the cfg thinks I am running mingw (which I am) - that's good because clang mimics the gcc compiler options that are implied in "mingw32". I tried running `luarocks build chronos` and the object built but I'm having a linker problem.

From what I understand, the gnu ld linker allows linking against a DLL even though "your not supposed to" (Don't know the mechanics behind it yet). The linker from LLVM - `ld.lld.exe` - does not allow that so I need to specify a lib file. I specified 'lua53.lib' in my config:

```lua
rocks_trees = {
    { name = [[user]],
         root    = home..[[/luarocks]],
    },
    { name = [[system]],
         root    = [[c:\program files (x86)\winlua\lua\5.3\]],
    },
}
variables = {
    --~ MSVCRT = 'VCRUNTIME140',
    CC = [[i686-w64-mingw32-clang.exe]],
    LD = [[i686-w64-mingw32-clang.exe]],
    LUA_LIBDIR = [[C:\Program Files (x86)\WinLua\Lua\5.3\bin]],
    LUA_INCDIR = [[C:\Program Files (x86)\WinLua\Lua\5.3\include]],
    LUALIB = 'lua53.lib',
}
verbose = true   -- set to 'true' to enable verbose output
```

I can see where variables.LUALIB is added in builtin.lua on line 183, but it's not my config value that I'm getting back:

```
i686-w64-mingw32-clang.exe -O2 -c -o src/chronos.o -IC:\Program Files (x86)\WinLua\Lua\5.3\include src/chronos.c
[...]

PR #1138, opened 2020-01-24
Adjust search mingw32 lib search to support llvm-mingw32 compiler.

RussellHaley:
Hello, 

As per #1135, I was able to get the llvm-mingw32 toolchain to work with luarocks. I'm not sure why the mingw32 internal build script doesn't respect variables.LUALIB from the config file, but the fix is minor. I simply adjusted the search path as provided by cfg.lua. I moved the `?.lib` search string to the front of the table definition. I don't expect this to affect mingw32-gcc, but that's a guess (based on an old faded memory) on my part. I suspect it either doesn't generate lib files (so won't find it in a search string) or is agnostic of their use. I'm indifferent to leaving the comment in (e.g. take it out if you feel). I thought it appropriate given there was an opposing comment there before.


src/luarocks/core/cfg.lua
```diff
@@ -347,8 +347,10 @@ local function make_defaults(lua_version, target_cpu, platforms, home)
       defaults.external_deps_patterns = {
          bin = { "?.exe", "?.bat" },
          -- mingw lookup list from http://stackoverflow.com/a/15853231/1793220
-         -- ...should we keep ?.lib at the end? It's not in the above list.
-         lib = { "lib?.dll.a", "?.dll.a", "lib?.a", "cyg?.dll", "lib?.dll", "?.dll", "?.lib" },
+         --~ 2020-01-23: putting ?.lib first allows use of the llvm-mingw32 compiler toolchain
+         --~ https://github.com/mstorsjo/llvm-mingw
+         --~ The LLVM linker doesn't support linking to a dll (GNU ld does).         
+         lib = { "?.lib", "lib?.dll.a", "?.dll.a", "lib?.a", "cyg?.dll", "lib?.dll", "?.dll" },
```
hishamhm: Is there a way to check that this change does not affect non-LLVM mingw32 in some way? Not sure if the testsuite (and the CI run for `spec/build_spec.lua` in Appveyor) covers it.

hishamhm: > Not sure if the testsuite (and the CI run for spec/build_spec.lua in Appveyor) covers it.

I think there may be a chance it does, in the very latest code of `master`. You might want to rebase this PR over the latest `origin/master` and force-push it so we get a new CI run.

hishamhm: On deeper inspection, I don't think the test suite actually checks that the build of a module with an external dependency works correctly (i.e. that it produces a module that the Lua interpreter is really able to load and use). I'm reluctant about merging this before further testing, I can try my hand at extending the tests (probably next week).

RussellHaley: I got a build working last night so I can move forward and test; no rush on up-streaming it. What platform are you concerned about? Windows, Linux?

hishamhm:
The other platform that would be affected by this change would be mingw
with gcc.

I recently added a test that builds a dummy external library on the spot
(look for "fixturedep") and uses it as an external dependency, but the test
doesn't go as far as verifying that the linkage was correct (i.e. that the
Lua module is actually able to use the C library successfully).

PR #1143, opened 2020-02-01
Add OpenBSD specific dependency paths.

jan-schreib:
This commit enables luarocks to be configured and build on OpenBSD.
The lua header files reside in /usr/local/include/lua-5.3/ and this changes tell the configure script and deps.lua about it.

### DIFF
--- a/configure
+++ b/configure
@@ -435,6 +435,7 @@ check_incdir() {
       "$incdir/lua$LUA_VERSION/lua.h" \
       "$incdir/lua$(echo "$LUA_VERSION" | tr -d .)/lua.h" \
       "$incdir/lua.h" \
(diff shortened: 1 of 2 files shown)


configure
```diff
@@ -435,6 +435,7 @@ check_incdir() {
       "$incdir/lua$LUA_VERSION/lua.h" \
       "$incdir/lua$(echo "$LUA_VERSION" | tr -d .)/lua.h" \
       "$incdir/lua.h" \
+      "$incdir/lua-$LUA_VERSION/lua.h" \
```
daurnimator: I think lua.h is meant to be checked last. Could you move this up a line or two?

daurnimator:
Merged in 7c70a320

PR #1145, opened 2020-02-02
Contribution Guidelines

aashimgarg:
The Contribution guidelines pr with all the changes mentioned by @hishamhm


hishamhm:
This line mentioning "front end change" (which makes no sense for LuaRocks) shows that this is copy-pasted from somewhere else.

Sorry, you seem to be pushing some generic Git instructions here that are not specific to LuaRocks. People can find this info elsewhere. I don't think we need this in the main README.

aashimgarg:
i thought that was missing from the documentation. I agree to you people can find these instructions about Git online. 
Thanks @hishamhm

PR #1151, opened 2020-02-05
fix downgrades of rocks containing directories

### LINKED ISSUE #1141: LuaRocks leaves empty files (at least on OSX)
This has been bothering me quite long. It seems like `downgrade/upgrade` leaves empty 0 bytes files behind, which originally have been `directories`.

This is how I can reproduce this:
```
git clone https://github.com/Kong/kong.git
cd kong
git fetch
git checkout next
make dev
git checkout 1.4.3
make dev
git checkout next
make dev
Error: .luaver/luarocks/3.3.0_5.1/share/lua/5.1/busted is not a directory
Error: .luaver/luarocks/3.3.0_5.1/share/lua/5.1/luacheck is not a directory
cd luaver/luarocks/3.3.0_5.1/share/lua/5.1
ls -l
-rwxr-xr-x   1 bungle  staff     0B Jan 30 13:58 busted*
-rwxr-xr-x   1 bungle  staff     0B Jan 30 13:59 luacheck*
```

hishamhm:
Fix #1141

(...hopefully! I was not really able to trigger the error in the test suite, but I was able to reproduce it by hand :\ and with this tweaks it doesn't happen anymore. I'm hoping the CI will tell me if this fix breaks something in some platform.)

### DIFF
--- a/src/luarocks/repos.lua
+++ b/src/luarocks/repos.lua
@@ -272,7 +272,7 @@ local function backup_existing(should_backup, target)
       until not fs.exists(backup) -- Slight race condition here, but shouldn't be a problem.
    
       util.warning(target.." is not tracked by this installation of LuaRocks. Moving it to "..backup)
-      local move_ok, move_err = fs.move(target, backup)
+      local move_ok, move_err = os.rename(target, backup)
       if not move_ok then
          return nil, move_err
       end
@@ -331,7 +331,7 @@ end
 local function rollback_install(op)
    fs.delete(op.dst)
    if op.backup_file then
-      fs.move(op.backup_file, op.dst)
+      os.rename(op.backup_file, op.dst)
    end
    fs.remove_dir_tree_if_empty(dir.dir_name(op.dst))
    return true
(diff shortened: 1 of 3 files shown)


src/luarocks/repos.lua
```diff
@@ -347,7 +347,7 @@ local function op_rename(op)
    if fs.exists(op.src) then
       fs.make_dir(dir.dir_name(op.dst))
       fs.delete(op.dst)
-      local ok, err = fs.move(op.src, op.dst)
+      local ok, err = os.rename(op.src, op.dst)
```
daurnimator: Could this one end up trying to rename across file systems?

hishamhm: no, this is used to really _rename_ things within a LuaRocks tree, and not move files around. For actual deployment of files ,`fs.move` is still used to move files (by copying and deleting them) one by one.

bungle:
@hishamhm 

I manually applied the patch to `3.3.0` and yes, it fixes the issue. Thank you!

PR #1182, opened 2020-04-15
Fix to avoid undefined global reference

elliottslaughter:
This patch is required to support for [Terra](http://terralang.org/), a language embedded in Lua/LuaJIT with support for JITing high-performance code.

The only change here is that Terra uses a strict mode by default, so the attempt to access an undefined global has be wrapped in pcall to make it safe.


src/luarocks/loader.lua
```diff
@@ -29,7 +29,7 @@ local require = nil
 
 -- Workaround for wrappers produced by older versions of LuaRocks
 local temporary_global = false
-if luarocks then
+if pcall(function() return luarocks end) then
```
hishamhm, line 32: that is incorrect: `pcall` succeeds even though `luarocks` is nil. Need to test the second return value instead.

elliottslaughter: I believe the updated patch does this, though it is still failing CI.

elliottslaughter: Er, maybe I was looking at the outdated CI results. It seems to have passed now. I think this resolved, perhaps modulo variable name preferences.

hishamhm:
@elliottslaughter Thanks for the patch! It's now merged and will be in the next release!

PR #1207, opened 2020-07-16
Update tools.lua - use --help flag versus -h

cvega:
If you're using busybox, luarocks fails despite unzip being installed. Using the --help flag works in GNU, Busybox, and OSX

### DIFF
--- a/src/luarocks/fs/unix/tools.lua
+++ b/src/luarocks/fs/unix/tools.lua
@@ -141,7 +141,7 @@ end
 -- @return boolean: true on success, nil and error message on failure.
 function tools.unzip(zipfile)
    assert(zipfile)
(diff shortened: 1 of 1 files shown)


cvega:
On second look. It appears `luarocks` determines if unzip is available at time of `configure` using the standard method `command`. Is there any reason `luarocks` wouldn't want to standardize these checks? Does `luacheck` have a need to use a specific version of `unzip`?

hishamhm:
Thanks for the PR!

> Is there any reason luarocks wouldn't want to standardize these checks?

That's a fair question. LuaRocks is generally moving to runtime checks to make the stand-alone binary version self-sufficient. The `configure` time checks are still there for the time being for the sake of Linux distros.

PR #1213, opened 2020-08-01
Add the `--no-install` option to the `make` command

dwenegar:
Add the --no-install option to the make command to build the rock w/o installing it.

This should improve the workflows using `luarocks make` to build a rock for testing purposes.
The rock will be compiled and _deployed_ in the source folder but not installed.

### DIFF
--- a/src/luarocks/build.lua
+++ b/src/luarocks/build.lua
@@ -23,6 +23,7 @@ build.opts = util.opts_table("build.opts", {
    verify = "boolean",
    check_lua_versions = "boolean",
    pin = "boolean",
+   no_install = "boolean"
 })
 
 do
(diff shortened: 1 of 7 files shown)


src/luarocks/build.lua
```diff
@@ -385,16 +386,20 @@ function build.build_rockspec(rockspec, opts)
       return name, version
    end   
 
-   if repos.is_installed(name, version) then
-      repos.delete_version(name, version, opts.deps_mode)
+   local dirs, err
+   if not opts.no_install then
+      if repos.is_installed(name, version) then
+         repos.delete_version(name, version, opts.deps_mode)
```
hishamhm, line 392: Does this mean that `--no-install` actually uninstalls a previously installed copy?

I think I understand why that is (so that the installed one doesn't override the one built in the local folder). Is that the idea?

dwenegar: It's `not opts.no_install`. the old rocks is not uninstalled.

I added this option to help my testing workflow. 

Lua-only rocks can be tested w/o `luarocks make`; mixed rocks (C and lua) cannot and need to be "made" first. `--no-install` stops everything after compiling and linking the rock.

hishamhm: @dwenegar Ah, got it! Thanks for the clarification!

src/luarocks/build.lua
```diff
@@ -385,16 +386,20 @@ function build.build_rockspec(rockspec, opts)
       return name, version
    end   
 
-   if repos.is_installed(name, version) then
-      repos.delete_version(name, version, opts.deps_mode)
+   local dirs, err
```
hishamhm, line 389: You need to move the declaration of `rollback` up here too

hishamhm:
@dwenegar I believe that if you fix the `rollback` variable as described above, the tests will pass and this PR will be mergeable

PR #1217, opened 2020-08-14
add CLI variable DEPS_DIR

fperrad:
This patch adds a 3rd way to declare a directory path for external dependency.

```luarocks build variables.DEPS_DIR /some/where foo```

When today, the variable name depends of the library required by the rock 

```luarocks build variables.LIBFOO_DIR /some/where foo```

And currently, as `external_deps_dirs` is a table, it cannot be modified by a command `luarocks config ...`.

It could be very convenient in the context of [Buildroot](http://buildroot.org).


hishamhm:
@fperrad sounds good, thanks! Would you mind adding a note in the wiki about this, marked as "(since 3.4.0)"?

hishamhm:
@fperrad Thank you! :heart:

PR #1218, opened 2020-08-24
Do not assume Lua 5.3 includes bit32

### LINKED ISSUE #1204: Luarocks 3.3 claims things are "provided by VM" that are not and fails to install rocks as requested
I just tried to upgrade some CI jobs that were using `luarocks` 2.4.2 to 3.3.1 and all the Lua 5.3 jobs in the matrix started failing. I traced it to a fiasco with the *bit32* module. The new luarocks is refusing to install it as a dependency claiming:

> sile dev-1 depends on bit32 (5.3-1 provided by VM)

The problem is it **isn't** provided by the VM. That would only be the case if Lua 5.3 was compiled with 5.2 compatibility flags enabled. These may be default, but for the purpose of a CI testing matrix it makes sense to compile without. Why it is assuming it is there when it may or may not be without even checking I don't understand.

hishamhm:
We can't autodetect this either because the interpreter running LuaRocks may not be the same one running Lua.

Fixes #1204.

### DIFF
--- a/src/luarocks/util.lua
+++ b/src/luarocks/util.lua
@@ -600,7 +600,7 @@ function util.get_rocks_provided(rockspec)
 
    rocks_provided["lua"] = lv.."-1"
 
-   if lv == "5.2" or lv == "5.3" then
+   if lv == "5.2" then
       rocks_provided["bit32"] = lv.."-1"
    end
 


daurnimator:
How can an installation say that its lua 5.3 *does* provide bit32?

hishamhm:
Using rocks_provided in the config file.

>

PR #1220, opened 2020-09-13
silence the MSVC env setup script

dwenegar:
The batch files used to initialize the environment when using the MSVC compiler print a banner that makes the use of `luarocks path` difficult in a CI context.

This PR make it so that the output of the environment initialization scripts is redirected to NUL.


hishamhm:
@dwenegar Thanks for the patch! Looks like your patch changed all of the Windows line-endings. (The Github diff shows all lines as modified) — could you double-check that? Thanks!!

dwenegar:
> 
> 
> @dwenegar Thanks for the patch! Looks like your patch changed all of the Windows line-endings. (The Github diff shows all lines as modified) — could you double-check that? Thanks!!

It did indeed change all the line endings.
It actually *fixed* them, since `.editorconfig` has `end_of_line = lf` for all files.

I updated the `.editrconfig` and fixed the line-endings.

hishamhm:
@dwenegar Thank you for going above and beyond! :) 

Merging!!

PR #1223, opened 2020-09-19
Keep the old url if the md5 didn't change.

dwenegar:
`check_url_and_update_md5` does not return an error on an invalid URL, and this causes the invalid URL to end up in the new rockspec file.

This PR add an optional parameter to `check_url_and_update_md5` called `invalid_is_error`. When this parameter is `true` the function will treat invalid URLs as errors.

When updating the url, if `check_url_and_update_md5` returns an error, the old URL is restored.

### DIFF
--- a/src/luarocks/cmd/new_version.lua
+++ b/src/luarocks/cmd/new_version.lua
@@ -26,8 +26,9 @@ revision number of the given (or downloaded) rockspec.
 
 If a URL is given, it replaces the one from the old rockspec with the given URL.
 If a URL is not given and a new version is given, it tries to guess the new URL
-by replacing occurrences of the version number in the URL or tag. It also tries
-to download the new URL to determine the new MD5 checksum.
+by replacing occurrences of the version number in the URL or tag; if the guessed
+URL is invalid, the old URL is restored. It also tries to download the new URL
+to determine the new MD5 checksum.
 
 If a tag is given, it replaces the one from the old rockspec. If there is an old
 tag but no new one passed, it is guessed in the same way URL is.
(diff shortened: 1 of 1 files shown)


src/luarocks/cmd/new_version.lua
```diff
@@ -118,8 +121,14 @@ local function update_source_section(out_rs, url, tag, old_ver, new_ver)
    if out_rs.source.file then
       try_replace(out_rs.source, "file", old_ver, new_ver)
    end
+
+   local old_url = out_rs.source.url
    if try_replace(out_rs.source, "url", old_ver, new_ver) then
-      return check_url_and_update_md5(out_rs)
+      local ok, err_or_changed = check_url_and_update_md5(out_rs, true)
```
daurnimator, line 127: I think this is the only call-site of `check_url_and_update_md5`; I think you shouldn't change the signature of the function, but instead change what it does.

daurnimator (changes requested):

src/luarocks/cmd/new_version.lua
dwenegar: I thought about that, but I preferred to change the signature and preserve the current behaviour to avoid ripple effects.

dwenegar:
Something like this? @daurnimator

daurnimator (dismissed):

daurnimator:
Yeah I guess that's okay. I'm not familiar with this section of the codebase so will wait for an additional review.

dwenegar:
I think it's best to change the behaviour of `new_version` only in the case the guessed URL is invalid.
I updated the command help accordingly.

hishamhm:
Behavior change looks okay to me.

hishamhm:
(Sorry, clicked the wrong button and closed it by mistake!)

hishamhm (approved):

PR #1224, opened 2020-09-19
fix(win32): generate proper temp filename

dwenegar:
`fs.tmpname()` is currently broken on windows.
On Windows, `os.tmpname()` uses `tmpnam()` that already uses the `TMP` environment variable, if defined (see [ _tempnam, _wtempnam, tmpnam, _wtmpnam](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/tempnam-wtempnam-tmpnam-wtmpnam?view=vs-2019)).

### DIFF
--- a/src/luarocks/fs/win32.lua
+++ b/src/luarocks/fs/win32.lua
@@ -336,7 +336,7 @@ function win32.make_temp_dir(name_pattern)
 end
 
 function win32.tmpname()
-   return os.getenv("TMP")..os.tmpname()
+   return os.tmpname()
 end
 
 function win32.current_user()
--- a/src/luarocks/fs/win32.lua
+++ b/src/luarocks/fs/win32.lua
@@ -336,7 +336,11 @@ function win32.make_temp_dir(name_pattern)
 end
 
 function win32.tmpname()
-   return os.tmpname()
+   local tmpname = os.tmpname()
+   if tmpname:match("^\\") then
+      tmpname = os.getenv("TMP") .. tmpname
+   end
+   return tmpname
 end
 
 function win32.current_user()


Tieske:
That does not look like it is backward compatible, this was a more recent change in Windows iirc.

Tieske:
see also: https://github.com/lunarmodules/Penlight/blob/master/lua/pl/path.lua#L465-L476

dwenegar:
> 
> 
> see also: [https://github.com/lunarmodules/Penlight/blob/master/lua/pl/path.lua#L465-L476](https://github.com/lunarmodules/Penlight/blob/master/lua/pl/path.lua?rgh-link-date=2020-09-19T15%3A49%3A25Z#L465-L476)

Tieske:
@dwenegar your code is checking for an initial "\", where as the Penlight code is checking for a ":". Why the difference?

(asking because I do not have the Windows setup and am wondering if we need to update Penlight as well?)

dwenegar:
> 
> 
> @dwenegar your code is checking for an initial "", where as the Penlight code is checking for a ":". Why the difference?
> 
> (asking because I do not have the Windows setup and am wondering if we need to update Penlight as well?)
Microsoft's documumentation says:

> In previous versions, the tmpnam and tmpnam_s functions generated file names in the root of the drive (such as \sd3c).

and I thought that checking for a strating `\` was the Correct™ thing to do.

src/luarocks/fs/win32.lua
```diff
@@ -336,7 +336,11 @@ function win32.make_temp_dir(name_pattern)
 end
 
 function win32.tmpname()
-   return os.getenv("TMP")..os.tmpname()
+   local tmpname = os.tmpname()
+   if tmpname:match("^\\") then
```
Tieske, line 340:
```suggestion
   if tmpname:match("\\") then
```
I think the temp name can also be prefixed with a drive letter, eg. `C:\something` in which case this would fail. The old Windows behaviour was anyway that it would only return a filename, and never include a path. So imo checking for a `\` should suffice, just not anchored at the first position.

dwenegar: Most of the references I found around mentions the temporary files created in the root directory, so the `\` anchored at the beginning seems the best choice.

Alternatively, another solution is checking if the path returned by `os.tmpname()` starts with `os.getenv('TMP')` and modifying it accordingly.

dwenegar: Most of the references I found around mentions the temporary files created in the root directory, so the `\` anchored at the beginning seems the best choice.

Alternatively, another solution is checking if the path returned by `os.tmpname()` starts with `os.getenv('TMP')` and modifying it accordingly.

hishamhm:
@Tieske @dwenegar folks, I'm planning to push a new LuaRocks release this week and wanted to check with you if this needs to be merged first: did you reach a conclusion with regard to this PR?

Tieske:
I wanted to give it a test on a windows machine to check exact behaviour, but haven't been able to yet. I'd also need to know this to update Penlight if necessary. (probably have Win machine back next week)

Tieske:
if you want to go ahead with it, then maybe switch to the way Penlight does the detection, since that has been around for a while now. We can then always update it later.

hishamhm:
@Tieske Penlight is using `TEMP`, unlike LuaRocks `master` and this PR, which use `TMP`. Which one is correct?

hishamhm:
Looks like the Microsoft docs linked above say `TMP`, so I went with @dwenegar 's suggestion [here](https://github.com/luarocks/luarocks/pull/1224#discussion_r493011698) and implemented the prefix check.

PR #1225, opened 2020-09-22
feat(persist): add configurable table indentation

dwenegar:
`luarocks new_version` writes the new rockspec using `persist.write_table()`. This function uses a default indentation of three spaces per level that is not configurable and might conflict with a project's coding style.

This PR makes the indentation used by `persist.write_table()` configurable.

The default behaviour is unchanged.


dol:
Why was this closed? I'm in favour of this change.

PR #1229, opened 2020-09-28
Use table.pack so that nil varags aren't dropped

daurnimator:
Replaces #1228


hishamhm:
@daurnimator Looks like it's failing CI, could you check? Thanks!

src/luarocks/fs/lua.lua
```diff
@@ -54,7 +56,9 @@ end
 
 local function quote_args(command, ...)
    local out = { command }
-   for _, arg in ipairs({...}) do
+   local args = pack(...)
+   for i=1, args.n do
+      local i = args[i]
```
siffiejoe, line 61: Should probably be:
```lua
local arg = args[i]
```

PR #1251, opened 2020-12-19
Add DragonFly BSD support and a fix for OpenBSD

sunpoet:


src/luarocks/core/cfg.lua
```diff
@@ -492,8 +494,17 @@ local function make_defaults(lua_version, target_cpu, platforms, home)
...
+      defaults.arch = "dragonfly-"..target_cpu
+      defaults.variables.CC = os.getenv("CC") or "cc"
+      defaults.variables.CFLAGS = os.getenv("CFLAGS") or defaults.variables.CFLAGS
+      defaults.variables.LD = defaults.variables.CC
+      defaults.variables.LIBFLAG = (os.getenv("LDFLAGS") or "").." -shared"
+   end
+
    if platforms.openbsd then
       defaults.arch = "openbsd-"..target_cpu
+      defaults.gcc_rpath = false
```
daurnimator, line 507: why turn this off?

hishamhm: probably because OpenBSD has switched to clang instead of gcc? Back in 2014 I made a similar change for FreeBSD because of this, at least: b66bc076af602d5afe1eb6f2bf0275d3e9bc2422

omar-polo:
Thanks for the OpenBSD bit, but it needs more than disabling gcc, I've opened #1352 for OpenBSD support.

hishamhm:
Addressed slightly differently via 36f6fd122162c49fd1737344c1a9de7cc7bf3758! Thanks for the PR!

PR #1256, opened 2021-01-07
Fix. Do not revoke permission from the current user

moteus:
*set_permissions* function revokes permissions from all users returned by *get_system_users*.

### DIFF
--- a/src/luarocks/fs/win32/tools.lua
+++ b/src/luarocks/fs/win32/tools.lua
@@ -1,4 +1,4 @@
-
+/
 --- fs operations implemented with third-party tools for Windows platform abstractions.
 -- Download http://unxutils.sourceforge.net/ for Windows GNU utilities
 -- used by this module.
@@ -253,7 +253,7 @@ function tools.set_permissions(filename, mode, scope)
       -- Finally, remove all the other users from the ACL in order to deny them access to the file
       for _, user in pairs(get_system_users()) do
          if username ~= user then
-            local ok = fs.execute_quiet(fs.Q(vars.ICACLS) .. " " .. fs.Q(filename) .. " /remove " .. fs.Q(user))
+            local ok = fs.execute_quiet(vars.ICACLS .. " " .. fs.Q(filename) .. " /remove " .. fs.Q(user))
             if not ok then
                return false, "Failed setting permission " .. mode .. " for " .. scope
             end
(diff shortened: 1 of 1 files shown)


src/luarocks/fs/win32/tools.lua
```diff
@@ -238,16 +244,19 @@ function tools.set_permissions(filename, mode, scope)
       if not ok then
          return false, "Could not take ownership of the given file"
       end
+      local username = os.getenv('USERNAME')
       -- Grant the current user the proper rights
-      ok = fs.execute_quiet(vars.ICACLS .. " " .. fs.Q(filename) .. " /inheritance:d /grant:r \"%USERNAME%\":" .. perms)
+      ok = fs.execute_quiet(vars.ICACLS .. " " .. fs.Q(filename) .. " /inheritance:d /grant:r " .. fs.Q(username) .. ":" .. perms)
```
daurnimator, line 249: Why do you need this change?

src/luarocks/fs/win32/tools.lua
```diff
@@ -204,11 +204,17 @@ end
 --- Helper function for fs.set_permissions
 -- @return table: an array of all system users
 local function get_system_users()
+   local exclude = {
+      [""]              = true,
+      ["Name"]          = true,
+      ["\128\164\172\168\173\168\225\226\224\160\226\174\224"] = true, -- Administrator in cp866
```
daurnimator, line 210: do we potentially need to add other locales in here?
and what if the person has that username but is not in codepage 866?

src/luarocks/fs/win32/tools.lua
daurnimator, line 249: do you need `fs.Q(vars.ICACLS)`?

daurnimator (commented):
What if we ran `wmic` with the codepage set to a known value (with e.g. `chcp`)?

src/luarocks/fs/win32/tools.lua
moteus: This more for explicitly use the same value in this command and in the if condition later.

src/luarocks/fs/win32/tools.lua
moteus: > do we potentially need to add other locales in here?

I think yes if you whant to exclude administrator user. Windows creates such users by default.

> and what if the person has that username but is not in codepage 866?

src/luarocks/fs/win32/tools.lua
moteus: This is not my change. This was remove somwhere between 3.0.4 and 3.5.0

hishamhm: We keep the command name variables unquoted so that a user can override them adding custom flags, e.g. `ICACLS = "icacls /mycustomflag"`

src/luarocks/fs/win32/tools.lua
```diff
@@ -262,12 +271,12 @@ function tools.set_permissions(filename, mode, scope)
 
       local ok
       -- Grant permissions available to all users
-      ok = fs.execute_quiet(vars.ICACLS .. " " .. fs.Q(filename) .. " /inheritance:d /grant:r *S-1-1-0:" .. others_perms)
+      ok = fs.execute_quiet(fs.Q(vars.ICACLS) .. " " .. fs.Q(filename) .. " /inheritance:d /grant:r Everyone:" .. others_perms)
```
hishamhm, line 274: could you remove `fs.Q` here and the other one below as well? (as per my other comment above)

src/luarocks/fs/win32/tools.lua
```diff
@@ -1,4 +1,4 @@
-
+/
```
hishamhm, line 1: stray `/` here, causing CI to fail

PR #1264, opened 2021-01-14
Fix auto-detecting cygwin: src/luarocks/core/sysdetect.lua

lemniscati:
Searching `cyggcc` doesn't work in a recent cygwin distribution.
Add a search for `cygwin` in a `local function detect_pe(fd)`.

### DIFF
--- a/src/luarocks/core/sysdetect.lua
+++ b/src/luarocks/core/sysdetect.lua
@@ -329,7 +329,7 @@ local function detect_pe(fd)
       rdata_pos = bytes2number(rdata_pos, LITTLE)
       fd:seek("set", rdata_pos)
       local data = fd:read(512)
-      if data:match("cyggcc") then
+      if data:match("cygwin") or data:match("cyggcc") then
          system = "cygwin"
       end
    end


daurnimator (commented):
Seems reasonable to me; though I can't test it myself.
Can you explain why this change is needed (what changed in cygwin upstream?)

lemniscati:
@daurnimator Thank you, for reply.

Though I don't know about the structure of PE files...

In a recent cygwin enviroment, I've checked that when `local function detect_pe(fd)` looks into  `/bin/sh`, `data` will not match `cyggcc`, but will match `cygwin`, and that in cases for `/proc/self/exe` and `/usr/bin/lua`, `data` will not match either of `cyggcc` and `cygwin`. 

* `detect_pe ()` will be called from function `sysdetect.detect_file()`
* `sysdetect.detect_file()` will be called from `sysdetect.detect()`: for files `/bin/sh`, `/proc/self/exe`, and `/usr/bin/lua` (`/proc/self/exe` and `/usr/bin/lua` will be a binary which luarocks will invoke,  and  the binary differ depending on configuration).
* `sysdetect.detect()` will be called from `cfg.init()`

I don't know about older cygwin environments...

daurnimator:
Could some other windows-using person please test? @Tieske maybe?

hishamhm:
In any case, since this keeps existing cases matching, this is super unlikely to cause issues as a non-Cygwin PE file is not likely to have `cygwin` at that position anyway. I say let's merge it.

Issue #1032, opened 2019-06-08
LuaRocks 3.1.3 - FreeBSD unzip returns `nil exit 1` when using the unzip -h option


RussellHaley:
Hi LuaRocks Team (Hisham),

Referencing prior issue here: https://github.com/luarocks/luarocks/issues/1022

I was very pleased to see the attempt to better support FreeBSD. Unfortunately the version of unzip that comes with FreeBSD does not support the -n option as used in LuaRocks. Note the error I get when I build and run luarocks from github on the v3.1.3 tag:

```
russellh@g1 ~/G/luarocks> luarocks

LuaRocks 3.1.3, the Lua package manager

<SNIP>

CONFIGURATION
	Lua version: 5.4

	Configuration files:
		System  : /usr/local/etc/luarocks/config-5.4.lua (ok)
		User    : /home/russellh/.luarocks/config-5.4.lua (not found)
		Project : /usr/home/russellh/Git/luarocks/./.luarocks/config-5.4.lua (ok)

	Rocks trees in use: 
		/usr/home/russellh/Git/luarocks/./lua_modules ("project")
		/home/russellh/.luarocks ("user")
		/usr/local ("system")

russellh@g1 ~/G/luarocks> luarocks list | less

Rocks installed for Lua 5.4
---------------------------

russellh@g1 ~/G/luarocks> luarocks install http
Installing https://luarocks.org/http-0.3-0.src.rock

Error: Failed unpacking rock file: /tmp/luarocks_luarocks-rock-http-0.3-0-zX4ddm/http-0.3-0.src.rock: 'unzip -n' program not found. Make sure unzip is installed and is available in your PATH (or you may want to edit the 'variables.UNZIP' value in file './.luarocks/config-5.4.lua')
```
Unfortunately the Lua test for `unzip -n` fails:

```

russellh@jailbird:~/freebsd/ports/devel/lua-luarocks/work/luarocks-3.1.0/src/luarocks/fs> lua53
Lua 5.3.5  Copyright (C) 1994-2018 Lua.org, PUC-Rio
> print(os.execute('unzip -n'))
Usage: unzip [-aCcfjLlnopqtuvyZ1] [-d dir] [-x pattern] zipfile
nil	exit	1

```
[...]

Issue #1022, opened 2019-05-31
Unzip not found on FreeBSD


kdbeall:
When `unzip` is installed through `pkg` on FreeBSD, luarocks incorrectly detects `unzip` as not installed. A workaround is described [here](http://lua.2524044.n2.nabble.com/LuaRocks-3-0-4-on-FreeBSD-unzip-not-found-td7684719.html).

kdbeall:
Steps to reproduce:

    vagrant init freebsd/FreeBSD-11.2-RELEASE
    vagrant up
    vagrant ssh

hishamhm:
@kdbeall As stated in that link, `unzip` in FreeBSD does not offer any way for us to test-run the command and make it return success (`0`) (help returns error (`1`), no version flag). 

We do the test run to provide users with a nicer error message when `unzip` is not available (and to do some baseline checking that the program we're calling works like the one we are expecting to call — in that regard, the code is doing what it's supposed to, as it detected that the `unzip` there is not the one we tested LuaRocks with).

Is `unzip` always preinstalled in FreeBSD base systems? If so we could just skip that test on FreeBSD.

kdbeall:
@hishamhm unzip has been always preinstalled in FreeBSD base systems since 8. So, it should be safe to simply skip the unzip.

siffiejoe:
Or we could ship a minimal zip-compressed string, dump it into a temporary file, and run unzip on it.

hishamhm:
@siffiejoe that could work but sounds like overkill, especially at runtime.

kdbeall:
@hishamhm it would probably be easiest to simply check if the platform is FreeBSD

kdbeall:
I'd put in a PR request myself, but I'm unfortunately not super familiar with Lua :(

kdbeall:
Thanks @hishamhm!

PR #1267, opened 2021-01-17
Add filesystem abstraction for NetBSD

lawrencehoffman:
* Add fs abstraction file for NetBSD
* Add init function with verification that the unzip binary is
  available
  
Note that this fixes an issue in NetBSD similar to issues #1022 and #1032


hishamhm:
@lawrencehoffman merged, thank you!

PR #1277, opened 2021-01-28
Dynamically detect libdir on Linux

michel-slm:
Some Linux distributions (e.g. Fedora, CentOS) put 64-bit libraries
in `/usr/lib64` rather than `/usr/lib`. On such systems `luarocks`
should use `lib64` rather than `lib`.

PS the implementation here is really suboptimal; ideally I can set this when `LUA_LIBDIR` is populated:

Signed-off-by: Michel Alexandre Salim <michel@michel-slm.name>

---
This change is [Reviewable](https://reviewable.io/reviews/luarocks/luarocks/1277)

### DIFF
--- a/src/luarocks/core/cfg.lua
+++ b/src/luarocks/core/cfg.lua
@@ -397,6 +397,18 @@ local function make_defaults(lua_version, target_cpu, platforms, home)
          defaults.variables.CFLAGS = defaults.variables.CFLAGS.." -fPIC"
       end
       defaults.web_browser = "xdg-open"
+      if platforms.linux then
+         -- inline code from fs/linux.lua since
+         -- luarocks.fs can't be required here
+         -- (circular dependencies)
+         local fd, _, code = io.open("/usr/lib64", "r")
+         if code ~= 2 then
+            defaults.lib_modules_path = "/lib64/lua/"..lua_version
+         end
+         if fd then
+            fd:close()
+         end
+      end
    end
 
    if platforms.cygwin then


src/luarocks/core/cfg.lua
```diff
@@ -397,6 +397,18 @@ local function make_defaults(lua_version, target_cpu, platforms, home)
          defaults.variables.CFLAGS = defaults.variables.CFLAGS.." -fPIC"
       end
       defaults.web_browser = "xdg-open"
+      if platforms.linux then
+         -- inline code from fs/linux.lua since
+         -- luarocks.fs can't be required here
+         -- (circular dependencies)
+         local fd, _, code = io.open("/usr/lib64", "r")
+         if code ~= 2 then
+            defaults.lib_modules_path = "/lib64/lua/"..lua_version
```
daurnimator, line 406: was this meant to be `/usr/lib64/lua`?

daurnimator (commented):
Just because `/usr/lib64` exists doesn't mean it should be used: the whole idea of that directory was so that you could have one harddrive/system image used by multiple machines/architectures.
e.g. you boot your machine in 32bit mode, it would use /usr/lib32; but then you reboot into 64bit mode and it would use /usr/lib64.

src/luarocks/core/cfg.lua
michel-slm: `/usr/lib64/lua` would also work, yes. I'm just trying to distinguish between 32-bit Fedora/RHEL (/usr/lib is used) and 64-bit Fedora/RHEL (libraries goes in /usr/lib64)

michel-slm:
> Just because `/usr/lib64` exists doesn't mean it should be used: the whole idea of that directory was so that you could have one harddrive/system image used by multiple machines/architectures.
> e.g. you boot your machine in 32bit mode, it would use /usr/lib32; but then you reboot into 64bit mode and it would use /usr/lib64.

In case of Fedora, the multilib support is more limited - it's really just meant so 32-bit compatibility libraries can be installed next to 64-bit libraries (in contrast to Debian, for example, where files go to /usr/lib/<archtriplet> - see https://wiki.debian.org/Multiarch/Implementation)

So I do realize this quirk does not generalize that well - if there's a way to override this behavior in a system configuration file that would be great. Can such configuration contain logic though? I want to build luarocks once on Fedora, regardless of what architecture it's installed on, but want it to dynamically decide where the compiled modules should be installed (pure Lua modules go to /usr/share/lua on Fedora)

src/luarocks/core/cfg.lua
hishamhm: @daurnimator the `lib_modules_path` variabe is a path fragment, not an absolute path.

daurnimator: > the `lib_modules_path` variable is a path fragment, not an absolute path.

Then shouldn't we `io.open(prefix .. "/lib64")` or similar?

hishamhm:
No, because it is testing the existence of absolute path as an OS feature
to see if the path fragment should be used. But you raise a valid question:
just because lib64 directories exist does it mean we should use them? IIRC
there was code somewhere to detect arch specific lib directories (for
Debian?) which probed gcc, but I'm not at the computer right now to double
check, I think it was in configure, not sure if it was moved into luarocks
itself.

hishamhm:
Recent LuaRocks already does some dynamic lib checking and includes lib64 support:
https://github.com/luarocks/luarocks/blob/master/src/luarocks/core/cfg.lua#L520-L528
so I think this PR is outdated and can be closed as the wanted functionality has already landed by other means!

Issue #1233, opened 2020-10-13
The command syntax is incorrect


Nazadup:
So i'm trying to install Luarocks in a Windows 10, and i did it multiple times using multiple methods but whenever i try the command: luarocks , i get that answer from the prompt: The command syntax is incorrect. I've tried the command on cmd, PowerShell, Visual Studio and Visual Studio Code, with all of them giving the same response. The directory path seems to be correct because when i get it wrong the prompts shows another error.

My Windows is in portuguese, but i'll post some prints anyway.
![luarockspowershell](https://user-images.githubusercontent.com/72818938/95919470-04110300-0d84-11eb-9421-ac5b55560072.PNG)
![luarocksstudiocode](https://user-images.githubusercontent.com/72818938/95919475-05dac680-0d84-11eb-837b-cb27ea194012.PNG)

Tieske:
If you want to try Lua on Windows, you could try [luawinmulti](https://github.com/tieske/luawinmulti) or [hererocks](https://github.com/luarocks/hererocks)

hishamhm:
Is this the same as #1232 ?

Nazadup:
> Is this the same as #1232 ?

I'm very new into programming, but it seems to be the same. I'm not sure tho.

Nazadup:
> If you want to try Lua on Windows, you could try [luawinmulti](https://github.com/tieske/luawinmulti) or [hererocks](https://github.com/luarocks/hererocks)

Thanks a lot, i'll give it a try right away. Do you think it's ok to a newbie in Lua use it on Windows, or do you think it would be easier for me to try it on Linux?

devrrior:
I have the same problem, when I type "luarocks" in cmd, it returns "The command syntax is incorrect". I need laurock for a neovim plugin. I dont know what to do. 
![image](https://user-images.githubusercontent.com/69869135/95922626-4c76f380-0d79-11eb-8c17-b5d4559f88ed.png)

dirkstff:
> 
> 
> Is this the same as #1232 ?

We would have to assume so. Symptoms are identical. And #1232 will hit mostly inexperienced users, because the more experienced ones will have some compiler installed, so that `vars.COMPILER_ENV_CMD != ""`.

dirkstff:
> 
> 
> I dont know what to do.

[...]

Issue #1232, opened 2020-10-13
bug in install.bat, produces invalid luarocks.bat and luarocks-admin.bat under certain conditions


dirkstff:
The bug was introduced by commit 04513372d7d5859ea865252095c3b5ce0762228f. Now, if `vars.COMPILER_ENV_CMD == ""`, which it can very well be, the line `$COMPILER_ENV_CMD >NUL` resolves to ` >NUL`, which is a cmd syntax error. Accordingly, luarocks.bat and luarocks-admin.bat fail.

hishamhm:
@dwenegar Looks like the issue was introduced by your PR — could you take a look at this, please? Thanks!

PR #1288, opened 2021-03-15
Fix commands broken by vcvarsall.bat's output

dwenegar:
When LuaRocks is installed without specifying a compiler, a call to `vcvarsall.bat` is inserted at the beginning of `luarocks.bat`.

This causes the output of `vcvarsall.bat` to mix with the output of LuaRocks; in turn this breaks the invocation of certain commands (for exmaple, `luarocks path --lr-path').

This pull-request solve this problem by redirecting  `vcvarsall.bat` to `NUL`.

### DIFF
--- a/install.bat
+++ b/install.bat
@@ -559,7 +559,7 @@ local function get_msvc_env_setup_cmd()
 		if exists(vcvarsall) then
 			local vcvarsall_args = { x86 = "", x86_64 = " x64", arm64 = " x86_arm64" }
 			assert(vcvarsall_args[vars.UNAME_M], "vars.UNAME_M: only x86, x86_64 and arm64 are supported")
-			return ('call "%s"%s'):format(vcvarsall, vcvarsall_args[vars.UNAME_M])
+			return ('call "%s"%s > NUL'):format(vcvarsall, vcvarsall_args[vars.UNAME_M])
 		end
 	end
 


install.bat
```diff
@@ -559,7 +559,7 @@ local function get_msvc_env_setup_cmd()
 		if exists(vcvarsall) then
 			local vcvarsall_args = { x86 = "", x86_64 = " x64", arm64 = " x86_arm64" }
 			assert(vcvarsall_args[vars.UNAME_M], "vars.UNAME_M: only x86, x86_64 and arm64 are supported")
-			return ('call "%s"%s'):format(vcvarsall, vcvarsall_args[vars.UNAME_M])
```
Tieske, line 562: could this be a Windows line-end?

hishamhm:
@dwenegar Hello! This looks like a similar change to the one you added in https://github.com/luarocks/luarocks/pull/1220 (adding a `> NUL` to the MSVC env setup script).

I have no way to currently test MSVC on Windows, so I merged that one trusting it would be a minor change.

However, multiple users reported issues with the following release that included it (see #1232 and #1233).

Since it looked like it was an issue caused by the PR you sent, I [tried contacting you](https://github.com/luarocks/luarocks/issues/1232#issuecomment-707791630) about it, but got no response.

In that situation, all I could do to address the users issues was reverting your change from #1220, in commit ffbb66dada1bd50acc71c20c26227753c2e7ffa5 and making a new LuaRocks release.

This PR looks like another take on the same topic, but again I have no way to test this directly, and I'm worried that the same thing may happen again (take the change, users report issue, no feedback from change author, revert change).

To give us more confidence in taking this patch, could you expand a bit explaining to us why #1220 caused problems and how is this a safer version than that one? Thank you!

dwenegar:
@hishamhm 

> This looks like a similar change to the one you added in #1220

oh, I had forgotten about that change. Looking at it now I can see that it was indeed a bad change.

> Since it looked like it was an issue caused by the PR you sent, I tried contacting you about it, but got no response.

I definitely missed this. It was not my intention to ignore you. Sorry.

> In that situation, all I could do to address the users issues was reverting your change from #1220, in commit ffbb66d and making a new LuaRocks release

Ah, that's why my CI stopped working 😄 

> To give us more confidence in taking this patch, could you expand a bit explaining to us why #1220 caused problems and how is this a safer version than that one? Thank you!

The reason is the same as #1220:

> The batch files used to initialize the environment when using the MSVC compiler print a banner that makes the > > use of luarocks path difficult in a CI context.

Not redirecting to NUL causes this, for example:

```
.\luarocks.bat path --lr-path
**********************************************************************
** Visual Studio 2019 Developer Command Prompt v16.9.1
** Copyright (c) 2021 Microsoft Corporation
**********************************************************************
[vcvarsall.bat] Environment initialized for: 'x64'
SET LUA_PATH=..\LR\lua\?.lua;..\LR\lua\?\init.lua;...
```

Making it impossible to capture and use the output of the command.

#1220 made the mistake to add `> NUL` to the `install.bat` script:

> `$COMPILER_ENV_CMD >NUL`

this PR adds it to `$COMPILER_ENV_CMD` so you don't end up with a line containing only `> NUL` in your `luarocks.bat` and `luarocks-admiun.bat` files, when `COMPILER_ENV_CMD` is the empty string.

install.bat
dwenegar: Yes, it is a `crlf`

install.bat
```diff
@@ -559,7 +559,7 @@ local function get_msvc_env_setup_cmd()
 		if exists(vcvarsall) then
 			local vcvarsall_args = { x86 = "", x86_64 = " x64", arm64 = " x86_arm64" }
 			assert(vcvarsall_args[vars.UNAME_M], "vars.UNAME_M: only x86, x86_64 and arm64 are supported")
-			return ('call "%s"%s'):format(vcvarsall, vcvarsall_args[vars.UNAME_M])
+			return ('call "%s"%s > NUL'):format(vcvarsall, vcvarsall_args[vars.UNAME_M])
```
hishamhm, line 562: There are three other `call`s below (lines 586, 595, 604). Do we know that ` > NUL`s are not necessary in those, or do they need to be added as well?

hishamhm:
@Tieske do you happen to currently have a Windows setup that would be able to do a test-run of this PR? :grimacing:

hishamhm:
@dwenegar Thank you for the clarifications! And no worries, in this day and age of notification overload, I end up letting a lot of stuff slide myself...

At first glance the change makes sense, especially since it appears to be added on a `call` line, so I guess there's always an argument anyway? (my memory of MS-DOS batch files is super hazy, so I have no idea how exactly `> NUL` interacts with `CALL` and arguments and whatnot, but I'll take your word for it :) )

install.bat
dwenegar: Probably the redirection is needed there too, but I have no way to test it. For my use-case, setting upLua and LuaRocks with a GitHub action, the fix in this PR is sufficient.

dwenegar:
> 
> 
> @dwenegar Thank you for the clarifications! And no worries, in this day and age of notification overload, I end up letting a lot of stuff slide myself...
> 
> At first glance the change makes sense, especially since it appears to be added on a `call` line, so I guess there's always an argument anyway? (my memory of MS-DOS batch files is super hazy, so I have no idea how exactly `> NUL` interacts with `CALL` and arguments and whatnot, but I'll take your word for it :) )

I installed LuaRocks with the `/MSVC` switch thus setting `COMPILER_ENV_CMD` to the empty string: the resulting `luarocks.bat` ran without problems.

hishamhm:
All right, I'm merging this, and I'll add the `> NUL` redirect to the other calls.

PR #1290, opened 2021-03-19
pack: check that directory inside archive actually exists

hishamhm:
cc @Tieske


spec/pack_spec.lua
```diff
@@ -80,6 +81,65 @@ describe("luarocks pack #integration", function()
...
+               }
+            ]], finally)
+
+            assert.is.falsy(run.luarocks_bool("pack test-1.0-1.rockspec"))
+            assert.is.falsy(lfs.attributes("test-1.0-1.src.rock"))
+         end, finally)
+      end)
+
+
+      it("fails packing a rockspec into a .src.rock dir if doesn't exist", function()
```
Tieske, line 137:
```suggestion
      it("fails packing a rockspec into a .src.rock if dir doesn't exist", function()
```

Tieske (approved):
typo, other than that lgtm!

spec/pack_spec.lua
hishamhm: thanks!

Issue #1293, opened 2021-03-31
[Win] Luarocks 3.5.0 does not respect default-lua-version


gustavo-nramires:
Hi, I have luarocks configured with `luarocks config lua_version 5.4` which 
(1) generates `default-lua-version.lua` at `%appdata%/luarocks` as well as 
(2) write `lua_version = "5.4"` to my `config-5.4.lua` file
However, when I run `luarocks`, it still shows the errors of `config-5.3.lua` (not found). It would be great if it autodetected the version without needing to set --lua-version 5.4 every time. Thank you

gustavo-nramires:
It appears we could insert the home directory in the check of default-lua-version, by inserting `table.insert(dirs, cfg.homeconfdir)` into line 238 of `/src/luarocks/cmd.lua`

Taswen:
You can execute command `luarocks config --scope system  lua_version 5.4` in cmd under administrator rights to set default version of lua. It  will write `lua_version = "5.4"` to `C:/Program Files/luarocks/config-5.4.lua` that actually decides which version of lua to use.

hishamhm:
@gustavo-nramires fixed via a modified version of your PR! thank you!

PR #1294, opened 2021-04-02
configure: check home dir for default-lua-version

gustavo-nramires:
Automatically detect default lua version by checking user home config directory.
https://github.com/luarocks/luarocks/issues/1293

### DIFF
--- a/src/luarocks/cmd.lua
+++ b/src/luarocks/cmd.lua
@@ -235,6 +235,7 @@ do
             table.insert(dirs, dir.path(cfg.home_tree, ".luarocks"))
          end
          table.insert(dirs, cfg.sysconfdir)
+         table.insert(dirs, cfg.homeconfdir)
          for _, d in ipairs(dirs) do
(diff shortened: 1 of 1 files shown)


src/luarocks/cmd.lua
```diff
@@ -235,6 +235,7 @@ do
             table.insert(dirs, dir.path(cfg.home_tree, ".luarocks"))
          end
          table.insert(dirs, cfg.sysconfdir)
+         table.insert(dirs, cfg.homeconfdir)
```
hishamhm, line 238: Oh, this is what line 235 was supposed to be doing! There's a mistake there, with `cfg.home_tree` instead of `cfg.home` — but indeed, I think a more correct fix would be to change that to `table.insert(dirs, cfg.homeconfdir)`.

hishamhm:
@gustavo-nramires I made the small change I mentioned above and merged it manually! Thank you for hunting down this issue!

PR #1296, opened 2021-04-03
Add an option to keep the existing URL to the `new_version` command.

dwenegar:
Add a flag to keep the existing URL in a rockspec if a new one has not been specified when invoking the `new_version` command.

If the URL contains the old version (for example `dev`) the guessed URL might be wrong.
Instead of fixing the way the URL is guessed, add an option to keep the old URL instead.

### DIFF
--- a/src/luarocks/cmd/new_version.lua
+++ b/src/luarocks/cmd/new_version.lua
@@ -48,6 +48,7 @@ if it already exists.]], util.see_also())
 
    cmd:option("--dir", "Output directory for the new rockspec.")
    cmd:option("--tag", "New SCM tag.")
+   cmd:flag("--keep-url", "Keep the old url if a new one is not specified.")
 end
 
 
@@ -200,6 +201,10 @@ function new_version.command(args)
    local out_rs, err = persist.load_into_table(filename)
    local out_name = out_rs.package:lower()
    out_rs.version = new_rockver.."-"..new_rev
+
+   if not args.new_url and args.keep_url then
+      args.new_url = out_rs.source.url
+   end
 
    local ok, err = update_source_section(out_rs, args.new_url, args.tag, old_ver, new_ver)
    if not ok then return nil, err end


src/luarocks/cmd/new_version.lua
```diff
@@ -48,6 +48,7 @@ if it already exists.]], util.see_also())
 
    cmd:option("--dir", "Output directory for the new rockspec.")
    cmd:option("--tag", "New SCM tag.")
+   cmd:flag("--keep-url", "Keep the old url if a new one is not specified.")
```
daurnimator, line 51: Perhaps invert the flag?
i.e. have a `--guess-url` that defaults to `true` and then you can do `--guess-url=false`

src/luarocks/cmd/new_version.lua
```diff
@@ -201,6 +202,10 @@ function new_version.command(args)
    local out_name = out_rs.package:lower()
    out_rs.version = new_rockver.."-"..new_rev
 
+   if not args.new_url and args.keep_url then
+      args.new_url = out_rs.source.url
```
daurnimator, line 206: I don't think we modify `args` anywhere; consider factoring out into a new local `new_url`.

src/luarocks/cmd/new_version.lua
dwenegar: I am not particularly fond of that style, it sort of pollutes the command line. What about `--no-guess-url`  ?

src/luarocks/cmd/new_version.lua
dwenegar: you do at line 179
```
if args.tag and not args.new_version then
      args.new_version = args.tag:gsub("^v", "")
end
```

hishamhm:
> If the URL contains the old version (for example dev) the guessed URL might be wrong.
> Instead of fixing the way the URL is guessed, add an option to keep the old URL instead.

I still don't understand the use-case. I understand that a guessed URL may be incorrect, but surely keeping the old URL is incorrect as well? In the cases where you know that the guessing will get it wrong and you know what the new correct URL is, then you can just explicitly pass the URL as an argument instead.

dwenegar:
> I still don't understand the use-case. 

My use case is updating the `source.tag` field and skipping guessing a new URL.

> I understand that a guessed URL may be incorrect, but surely keeping the old URL is incorrect as well?

Keeping the old URL might be incorrect depending on the URL. The switch makes keeping the old URL _opt-in_: the current behaviour is not changed.

> In the cases where you know that the guessing will get it wrong and you know what the new correct URL is, then you can just explicitly pass the URL as an argument instead.

Sure. Just trying to save typing that.

hishamhm:
On Mon, Apr 5, 2021, 01:26 dwenegar ***@***.***> wrote:

> I still don't understand the use-case.
>
> My use case is updating the source.tag field and skipping guessing a new
> URL.
>
In such cases, the tag should be updated but the repository URL should not
contain the version number, so the version guessing shouldn't be changing
anything anyway. Did you hit a bug perhaps? What is the concrete example
that is causing you trouble?


I understand that a guessed URL may be incorrect, but surely keeping the
> old URL is incorrect as well?
>
> Keeping the old URL might be incorrect depending on the URL. The switch
> makes keeping the old URL *opt-in*: the current behaviour is not changed.
>
What I don't like about the proposed feature is that I don't want to have
is situations where people "keep the URL" and then overwrite the tarball in
their servers, thus breaking rockspecs for old versions.

I've been bitten too many times by open source maintainers keeping
unversioned URLs that break old dependencies using things like
latest.tar.gz.


In the cases where you know that the guessing will get it wrong and you
> know what the new correct URL is, then you can just explicitly pass the URL
> as an argument instead.
>
> Sure. Just trying to save typing that.
>

My concern is that this feature also facilitates the unfortunately common
antipattern I described above, so I'm not sure the gains outweighs the
risks. A crucial thing in UX design is steering users in the right
direction.

dwenegar:
@hishamhm Your concerns are valid; thanks for the feedback.

PR #1297, opened 2021-04-03
Update the source's MD5 only if the rockspec contains the field.

dwenegar:
If the source doesn't contain the `md5` field, do not try to update it and consider it unchanged.


daurnimator:
why would you not want to fill it in?

dwenegar:
> why would you not want to fill it in?

That's what the code does at present (see line 92~).
I just shortcut the process.

dwenegar:
I'll rethink it

PR #1298, opened 2021-04-05
XDG Base Directory

### LINKED ISSUE #1086: luarocks and config file
With the new release of luarocks, I've updated my fork and how nixpkgs converts luarocks packages into nix packages.
One thing I realized is that you can't override the configuration file from the command line.
You have to export LUAROCKS_CONFIG. I would prefer a flag since that makes the usage of a specific config more explicit.
Also, right now user config is in $HOME/.luarocks. Supporting the XDG spec would be cleaner. aka installing config in $XDG_CONFIG_HOME/luarocks (that I could share via my dotfiles) and the cache in $XDG_CACHE_HOME/luarocks (that I can prune easily).

thanks for your work on luarocks.

### LINKED ISSUE #1219: How to change the location of the user config file
Sorry if this is the wrong place for questions.
I would like to change location of the user config file from the default $HOME/.luarocks/config-5.3.lua to $XDG_CONFIG_HOME/luarocks/config-5.3.lua. I can't find any helpful information about how to achieve this from the manual pages and help outputs. How can I change the location of the user config file?
Thanks in advance.

carlsmedstad:
Hello again :wave:

This PR makes LuaRocks follow the XDG Base Directory specification. Info about this can be found here:
* https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html
* https://wiki.archlinux.org/index.php/XDG_Base_Directory

The jist of it is mainly that LuaRocks respects the environment variables `XDG_CACHE_HOME` and `XDG_CONFIG_HOME` and uses those directories accordingly. This is done in an backwards compatible manner, so for users that have their config file `~/.luarocks/` this will not have an affect at all. This does however allow users to place their config in `XDG_CONFIG_HOME/luarocks/` instead, should the wish to.

I've not yet added tests or written any documentation, but I'll be sure to do that if this PR gets traction.

Resolves #1086.
Resolves #1219.

### DIFF
--- a/src/luarocks/cmd.lua
+++ b/src/luarocks/cmd.lua
@@ -85,7 +85,7 @@ do
             replace_tree(args, root_dir)
          end
       elseif args["local"] then
-         if not cfg.home_tree then
+         if os.getenv("USER") == "root" then
             return nil, "The --local flag is meant for operating in a user's home directory.\n"..
                "You are running as a superuser, which is intended for system-wide operation.\n"..
                "To force using the superuser's home, use --tree explicitly."
@@ -642,7 +642,7 @@ function cmd.run_command(description, commands, external_namespace, ...)
    end
 
    -- if running as superuser, use system cache dir
-   if not cfg.home_tree then
+   if os.getenv("USER") == "root" then
       cfg.local_cache = dir.path(fs.system_cache_dir(), "luarocks")
    end
 
--- a/src/luarocks/cmd.lua
+++ b/src/luarocks/cmd.lua
@@ -85,7 +85,7 @@ do
             replace_tree(args, root_dir)
          end
       elseif args["local"] then
-         if os.getenv("USER") == "root" then
+         if fs.is_superuser() then
             return nil, "The --local flag is meant for operating in a user's home directory.\n"..
                "You are running as a superuser, which is intended for system-wide operation.\n"..
                "To force using the superuser's home, use --tree explicitly."
@@ -642,7 +642,7 @@ function cmd.run_command(description, commands, external_namespace, ...)
    end
 
    -- if running as superuser, use system cache dir
-   if os.getenv("USER") == "root" then
+   if fs.is_superuser() then
       cfg.local_cache = dir.path(fs.system_cache_dir(), "luarocks")
    end
 
--- a/src/luarocks/core/cfg.lua
+++ b/src/luarocks/core/cfg.lua
@@ -385,7 +385,8 @@ local function make_defaults(lua_version, target_cpu, platforms, home)
       }
       defaults.export_path_separator = ":"
       defaults.wrapper_suffix = ""
-      defaults.local_cache = home.."/.cache/luarocks"
+      local xdg_cache_home = os.getenv("XDG_CACHE_HOME") or home.."/.cache"
+      defaults.local_cache = xdg_cache_home.."/luarocks"
       if not defaults.variables.CFLAGS:match("-fPIC") then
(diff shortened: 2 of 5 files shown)


src/luarocks/cmd.lua
```diff
@@ -85,7 +85,7 @@ do
             replace_tree(args, root_dir)
          end
       elseif args["local"] then
-         if not cfg.home_tree then
+         if os.getenv("USER") == "root" then
```
hishamhm, line 88: I agree the previous test was done in a roundabout way and being explicit about it is better, but this solution above is Unix-specific, so it should be abstracted away. Maybe a `fs.is_superuser` function? Or `cfg.is_superuser` in case this is too early and `fs` platforms are not configured yet.

I understand the condition would only hit in a Unix system, but for future maintenance is better to not add platform-dependent code on the platform-independent modules.

src/luarocks/cmd.lua
```diff
@@ -642,7 +642,7 @@ function cmd.run_command(description, commands, external_namespace, ...)
    end
 
    -- if running as superuser, use system cache dir
-   if not cfg.home_tree then
+   if os.getenv("USER") == "root" then
```
hishamhm, line 645: same as the previous comment!

hishamhm:
Changes overall look good! I left a comment about portability there but from what I can tell cfg changes look nice and backward-compatible for people with existing installations.

I see that the tree of rocks itself continues to live at `~/.luarocks`. If you look inside you'll see it looks like a Unix prefix, with `bin`, `lib`, `share`. Is there an "XDG place" for that as well, such as `~/.local`?

carlsmedstad:
Thanks, for the quick feedback!

> I see that the tree of rocks itself continues to live at ~/.luarocks. If you look inside you'll see it looks like a Unix prefix, with bin, lib, share. Is there an "XDG place" for that as well, such as ~/.local?

Not from my understanding of the spec. There is `XDG_DATA_HOME` (which defaults to `~/.local/share/`) that can be used for more general application "data" but I think this is a bad fit for our use-case. As such I consider the default location of the home tree to be unrelated to the XDG Base DIrectory spec compliance.

That said, I think it would be nice to have `~/.local` as the default home tree location. This is where pip places local packages, for example. However, I feel making this change while staying backwards compatible would be slightly harder problem to solve. Maybe something for a future PR :)

src/luarocks/cmd.lua
carlsmedstad: Good suggestion! I'll try implementing such a function.

carlsmedstad:
I started an attempt to make an integration test that ensures that config files in `XDG_CONFIG_HOME/luarocks` are being read but `busted` is not my strong side and this turned out to be fairly complex. One has to create the directory structure, set `XDG_CONFIG_HOME` and create a config file with a name matching the version that is being tested.

Would you be able to help with this, or do you think this could be merged as is?

hishamhm:
Yes, I think we can just merge this — from eyeballing the changes, the PR looks good to me! I'll give it a shot at writing a test after merging. Could you update the wiki with some docs about this functionality? Thanks!

carlsmedstad:
Great :) Sure thing, I'll do it right away. Thank you too!

hishamhm:
I've pushed the test — there's a ton of non-obvious boilerplate that currently goes into the testsuite which makes it a pain to work with :\ , but I knew which other similar tests to copy :)
https://github.com/luarocks/luarocks/commit/81b28e34e02c3464dc2357bf3b6958f82d3ae823

PR #1311, opened 2021-05-10
deps: remove redundant params

Gerrard-YNWA:
Should remove the second argument `true` due to the definition in [`find_suitable_rock`](https://github.com/luarocks/luarocks/blob/master/src/luarocks/search.lua#L230).


hishamhm:
@Gerrard-YNWA thanks!!

PR #1318, opened 2021-07-07
The "--tree" parameter affects path and cpath

wangling12:
Without these, the function of the "--tree" parameter is very limited.


hishamhm:
This should have been addressed by f4b1deab270a94778e1bd8903ff48387b0ed9fbc

PR #1321, opened 2021-08-10
upload: support uploading pre-packaged .src.rock files

hishamhm:


src/luarocks/cmd/upload.lua
```diff
@@ -14,6 +14,8 @@ function upload.add_to_parser(parser)
       :summary("Upload a rockspec to the public rocks repository.")
 
    cmd:argument("rockspec", "Rockspec for the rock to upload.")
+   cmd:argument("src_rock", "A corresponding .src.rock file; if not given it will be generated.")
```
daurnimator, line 17: generally we use kebab-case for arguments
```suggestion
   cmd:argument("src-rock", "A corresponding .src.rock file; if not given it will be generated.")
```

src/luarocks/cmd/upload.lua
```diff
@@ -73,19 +75,22 @@ function upload.command(args)
    end
 
    local rock_fname
-   if not args.skip_pack and not is_dev_version(rockspec.version) then
+   if args.src_rock then
+      rock_fname = args.src_rock
```
daurnimator, line 79: Worth adding a test for when there is no file at the path?

hishamhm:
@daurnimator thanks for the review :)

PR #1352, opened 2021-09-11
fix configuration for OpenBSD

omar-polo:
Hello,

I'd like to port luarocks to OpenBSD.  With this PR I'm able to install and use various libraries (tried with luasockets, dkjson, lua-lsp, luaffifb) -- I'm just a lua beginner thought.

I've seen that for other OS there is an implementation of filesystem and platform abstractions, is this needed for all OSes?  Or there are other interfaces that are better be implemented?

Thanks!

### DIFF
--- a/src/luarocks/core/cfg.lua
+++ b/src/luarocks/core/cfg.lua
@@ -480,6 +480,11 @@ local function make_defaults(lua_version, target_cpu, platforms, home)
 
    if platforms.openbsd then
       defaults.arch = "openbsd-"..target_cpu
+      defaults.gcc_rpath = false
+      defaults.variables.CC = os.getenv("CC") or "cc"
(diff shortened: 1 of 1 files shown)


omar-polo:
Oh, I forgot to explain why the changes.  I'm disabling `gcc` and using `cc` instead because on (most?) architectures `cc` is clang but on those not supported by LLVM `cc` is actually an old `gcc`.  GCC is not shipped anymore in base on amd64 and (if wanted) needs to be installed via packages (but then it's called `egcc`, for "external" I presume)

hishamhm:
Thanks for the PR! Looks like the changes mirror those for FreeBSD, so it seems safe to apply.

> I've seen that for other OS there is an implementation of filesystem and platform abstractions, is this needed for all OSes?

Not unless you hit compatibility issues, but if you were able to build rocks successfully, then they shouldn't be necessary.

PR #1353, opened 2021-09-11
Prevent creation of directory named '-p' on Windows

stevenwdv:
`tools.make_dir` used `mkdir -p <name>` while `-p` [is not an existing switch](https://docs.microsoft.com/windows-server/administration/windows-commands/mkdir) on Windows. The switch is also not necessary because plain `mkdir` on Windows also creates intermediate directories.

### DIFF
--- a/src/luarocks/fs/win32/tools.lua
+++ b/src/luarocks/fs/win32/tools.lua
@@ -36,7 +36,7 @@ end
 function tools.make_dir(directory)
    assert(directory)
    directory = dir.normalize(directory)
-   fs.execute_quiet(vars.MKDIR.." -p ", directory)
+   fs.execute_quiet(vars.MKDIR, directory)
    if not fs.is_dir(directory) then
       return false, "failed making directory "..directory
    end


hishamhm:
@stevenwdv Interesting, how did you hit this issue?

When deployed via the legacy zip file and installed with `INSTALL.BAT`, LuaRocks ships a `mkdir.exe` binary, which mimics the Unix behavior.

When deployed as a single compiled `luarocks.exe` binary, that binary bundles LuaFileSystem, which is used by `luarocks/fs/lua.lua`, so that particular implementation using `vars.MKDIR` shouldn't be used.

We are phasing out the tool-based implementation in favor of the LuaFileSystem-based implementation, so this problem should go away in a future release.

stevenwdv:
@hishamhm I installed it [via MSYS2](https://packages.msys2.org/package/mingw-w64-x86_64-lua-luarocks?repo=mingw64), apparently it uses `cmd` there but does not have the utility.
Glad to hear you're ditching executing commands in the shell, as it can indeed be fragile.

PR #1355, opened 2021-09-21
fix check_external_dependency_at

erw7:
Fix a problem where `files` were being sorted even though they should have been inspected in the order of `external_deps_patterns`.


hishamhm:
Thanks for the patch!

Rereading the code, the original intent of that `table.sort` was to apply exact matches (not including `*`) ahead of matches including `*`.

Perhaps instead of removing the `table.sort`, we should instead produce a new table, filtering the non-`*`-entries first and then the `*`-entries?

hishamhm:
Thank you @erw7 !

PR #1362, opened 2021-11-02
always use https for cloning git repository

ElvinEfendi:
Github deprecating the usage of `git:` (among other things) https://github.blog/2021-09-01-improving-git-protocol-security-github/

I suspect this is not the best solution, but hopefully good enough to start the discussion. Or should this even not be concern of Luarocks and be handled in  individual rock files?

### DIFF
--- a/src/luarocks/fetch/git.lua
+++ b/src/luarocks/fetch/git.lua
@@ -112,7 +112,8 @@ function git.get_sources(rockspec, extract, dest_dir, depth)
    local ok, err = fs.change_dir(store_dir)
    if not ok then return nil, err end
 
-   local command = {fs.Q(git_cmd), "clone", depth or "--depth=1", rockspec.source.url, module}
+   local https_source_url = "https://" .. rockspec.source.url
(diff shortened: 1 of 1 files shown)


src/luarocks/fetch/git.lua
```diff
@@ -112,7 +112,8 @@ function git.get_sources(rockspec, extract, dest_dir, depth)
    local ok, err = fs.change_dir(store_dir)
    if not ok then return nil, err end
 
-   local command = {fs.Q(git_cmd), "clone", depth or "--depth=1", rockspec.source.url, module}
+   local https_source_url = "https://" .. rockspec.source.url
```
daurnimator, line 115: Just putting https at the front isn't going to be correct. e.g. the url in the rockspec might already have `https` in the front.
Alternatively it could be `ssh://`, or numerous other options.

hishamhm:
Thanks for the PR! I implemented an alternative solution in 9ff512e35455939f02eaec2318e3acc77782fdeb

PR #1370, opened 2021-11-22
Improve performance of reading zip file by not processing header

arichard4:
lfh is never accessed non-circularly. The two circular accesses are to let us know how long the file_name and extra_field are, to skip over them.

We can improve the performance of reading individual zip files by skipping over processing the unused local file header. Note that we pass in the already-processed central directory metadata to this function, and rely solely on it.

### DIFF
--- a/src/luarocks/tools/zip.lua
+++ b/src/luarocks/tools/zip.lua
@@ -329,19 +329,13 @@ local function read_file_in_zip(zh, cdr)
       return nil, "failed reading Local File Header signature"
    end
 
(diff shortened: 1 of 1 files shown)


hishamhm:
Interesting! Did this change give you a noticeable performance increase?

arichard4:
No, not a significant improvement. I didn't find this via performance analysis, I found this via static analysis.

(lfh is referenced circularly, for the last two reads; but most of its fields are never read.)

PR #1390, opened 2022-01-20
[fix] The misleading `--pin` option

### LINKED ISSUE #1389: The `--pin` option documentation seems inconsistent
### Please read the following before submitting:
- Please do NOT submit bug reports for questions. Ask questions in the [Gitter
  chat-room](https://gitter.im/luarocks/luarocks).
- Please do NOT submit duplicate bug reports. Look through the [issue
  tracker](https://github.com/luarocks/luarocks/issues) before submitting.

### Please fill out the following:
- **Platform:**
  - Linux (but irrelevant)

- **LuaRocks version:**
  - `luarocks --version` /usr/bin/luarocks 3.8.0

- **Configuration file:**
  - Please try to produce with the default configuration.
  - If you cannot reproduce with the default configuration, please try to find
    the minimal configuration to reproduce.
  - Upload the config to a pastebin such as gist.github.com.

Reproducible from current master after `./configure && make`

  - Use the `--verbose` flag.

- **Description:**
  - The steps you took in plain English to reproduce the problem.

Tried to use the `--pin` option.

The main luarocks help message includes a mention of this option, but it is inusable : 

```sh
$ luarocks --help
Usage: luarocks [-h] [--version] [--dev] [--server <server>]
       [--only-server <server>] [--only-sources <url>]
       [--namespace <namespace>] [--lua-dir <prefix>]
       [--lua-version <ver>] [--tree <tree>] [--local] [--global]
       [--verbose] [--timeout <seconds>] [--pin] [<command>] ...

LuaRocks 3.8.0, the Lua package manager

/usr/bin/luarocks - LuaRocks main command-line interface

Options:
[...]
   --pin                 Create a luarocks.lock file listing the exact versions
                         of each dependency found for this rock (recursively),
                         and store it in the rock's directory. Ignores any
                         existing luarocks.lock file in the rock's sources.

Commands:
[...]
```

```sh
$ luarocks --pin

[the output of `luarock` called without option or command]
```

After further investigation, this option is supposed to be used with `build` and `make` commands.
I think it should be removed from the "main help" message.

Reading #1127, the `--pin` option can also be used with the `install` command (if the installed rock is a Lua module, I'm not sure to understand what it means, looking at the code, I guess it's because when calling the `install` command on a `.rockspec` file "redirects" to the `build` command).
But this option is not part of the `luarocks install --help` message. I think it should be explicitly documented.

Thank you for your work!

Aire-One:
Hello,

This is my attempt at fixing the issue I reported earlier #1389.

Thank you, @daurnimator, for your input on the build command part.

I also tried to document the ability to use this option with the `install` command. I for example used it to generate the `luarocks.lock` file of my module after adding dependencies in the `.rockspec` file : `./luarocks install --deps-only --pin testmodule-1-1.rockspec`.

However, I don't know if the description I wrote is correct. Any feedback is welcomed! :)

Fixes #1389

### DIFF
--- a/src/luarocks/cmd/build.lua
+++ b/src/luarocks/cmd/build.lua
@@ -36,7 +36,7 @@ function cmd_build.add_to_parser(parser)
       "rockspec. Allows to specify a different branch to fetch. Particularly "..
       'for "dev" rocks.')
       :argname("<name>")
-   parser:flag("--pin", "Create a luarocks.lock file listing the exact "..
+   cmd:flag("--pin", "Create a luarocks.lock file listing the exact "..
       "versions of each dependency found for this rock (recursively), "..
       "and store it in the rock's directory. "..
       "Ignores any existing luarocks.lock file in the rock's sources.")
--- a/src/luarocks/cmd/install.lua
+++ b/src/luarocks/cmd/install.lua
@@ -43,6 +43,10 @@ function install.add_to_parser(parser)
       "and report if it is available for another Lua version.")
    util.deps_mode_option(cmd)
    cmd:flag("--no-manifest", "Skip creating/updating the manifest")
+   cmd:flag("--pin", "If the installed rock is a Lua module, create a "..
+      "luarocks.lock file listing the exact versions of each dependency found for "..
+      "this rock (recursively), and store it in the rock's directory. "..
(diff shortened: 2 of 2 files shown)


hishamhm:
@Aire-One Thank you for the fix!! Merged!

PR #1398, opened 2022-03-07
Test tweaks

hishamhm:


src/luarocks/core/cfg.lua
```diff
@@ -358,7 +361,16 @@ local function make_defaults(lua_version, target_cpu, platforms, home)
       defaults.external_lib_extension = "so"
       defaults.obj_extension = "o"
       defaults.external_deps_dirs = { "/usr/local", "/usr", "/" }
-      defaults.variables.CFLAGS = "-O2"
+
+      defaults.variables.CFLAGS = os.getenv("CFLAGS") or "-O2"
+      -- we pass -fPIC via CFLAGS because of old Makefile-based Lua projects
+      -- which didn't have -fPIC in their Makefiles but which honor CFLAGS
```
daurnimator, line 367: Then shouldn't this be in the `Makefile` build backend?

hishamhm:
I think this has fixed other backends in the past as well, plus it's easier
to keep track of platform differences here.

PR #1411, opened 2022-04-18
configure: Use shell '#' and '%' magic instead of sed

hugomg:
Mainly a style preference... Also has the benefit of not calling an external process for this.

### DIFF
--- a/configure
+++ b/configure
@@ -211,8 +211,8 @@ search_interpreter() {
 
 while [ -n "$1" ]
 do
-   value="$(echo "$1" | sed 's/[^=]*.\(.*\)/\1/')"
-   key="$(echo "$1" | sed 's/=.*//')"
+   key=${1%%=*}
+   value=${1#*=}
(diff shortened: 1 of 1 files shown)


hishamhm:
is this change POSIX-shell compatible? This script is `#!/bin/sh`, not bash, and I've had fixes sent in in the past for stuff such as "fix for Solaris", so I'm always wary of using advanced stuff

hugomg:
AFAIK this is perfectly kosher. (See the link below,  ctrl-f for "largest suffix")

https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02

That said, in the end it boils down to a style decision. I will trust your judgement on this one.

hishamhm:
I've bitten too many times by surprising incompatibilities in shell scripts, so I'll go with "if it ain't broke don't fix it" on this one :\ thanks for taking the time and making the PR, though!

PR #1433, opened 2022-08-22
cfg: follow standard enviromental variables

Artturin:
distributions use these to use the correct compiler and other tools

### DIFF
--- a/src/luarocks/core/cfg.lua
+++ b/src/luarocks/core/cfg.lua
@@ -215,11 +215,11 @@ local function make_defaults(lua_version, target_cpu, platforms, home)
       connection_timeout = 30,  -- 0 = no timeout
 
       variables = {
-         MAKE = "make",
-         CC = "cc",
-         LD = "ld",
-         AR = "ar",
-         RANLIB = "ranlib",
+         MAKE = os.getenv("MAKE") or "make",
+         CC = os.getenv("CC") or "cc",
+         LD = os.getenv("CC") or "ld",
+         AR = os.getenv("AR") or "ar",
+         RANLIB = os.getenv("RANLIB") or "ranlib",
 
          CVS = "cvs",
          GIT = "git",
@@ -287,12 +287,12 @@ local function make_defaults(lua_version, target_cpu, platforms, home)
 
       defaults.makefile = "Makefile.win"
       defaults.variables.PWD = "echo %cd%"
-      defaults.variables.MAKE = "nmake"
-      defaults.variables.CC = "cl"
-      defaults.variables.RC = "rc"
+      defaults.variables.MAKE = os.getenv("MAKE") or "nmake"
+      defaults.variables.CC = os.getenv("CC") or "cl"
+      defaults.variables.RC = os.getenv("WINDRES") or "rc"
       defaults.variables.LD = "link"
-      defaults.variables.MT = "mt"
-      defaults.variables.AR = "lib"
+      defaults.variables.MT = os.getenv("MT") or "mt"
+      defaults.variables.AR = os.getenv("AR") or "lib"
       defaults.variables.LUALIB = "lua"..lua_version..".lib"
       defaults.variables.CFLAGS = os.getenv("CFLAGS") or "/nologo /MD /O2"
       defaults.variables.LDFLAGS = os.getenv("LDFLAGS")
@@ -330,12 +330,12 @@ local function make_defaults(lua_version, target_cpu, platforms, home)
       defaults.static_lib_extension = "a"
       defaults.external_deps_dirs = { "c:/external/", "c:/mingw", "c:/windows/system32" }
       defaults.cmake_generator = "MinGW Makefiles"
-      defaults.variables.MAKE = "mingw32-make"
-      defaults.variables.CC = "mingw32-gcc"
-      defaults.variables.RC = "windres"
-      defaults.variables.LD = "mingw32-gcc"
-      defaults.variables.AR = "ar"
-      defaults.variables.RANLIB = "ranlib"
+      defaults.variables.MAKE = os.getenv("MAKE") or "mingw32-make"
+      defaults.variables.CC = os.getenv("CC") or "mingw32-gcc"
+      defaults.variables.RC = os.getenv("WINDRES") or "windres"
+      defaults.variables.LD = os.getenv("CC") or "mingw32-gcc"
+      defaults.variables.AR = os.getenv("AR") or "ar"
+      defaults.variables.RANLIB = os.getenv("RANLIB") or "ranlib"
       defaults.variables.CFLAGS = os.getenv("CFLAGS") or "-O2"
       defaults.variables.LDFLAGS = os.getenv("LDFLAGS")
       defaults.variables.LIBFLAG = "-shared"
@@ -372,8 +372,8 @@ local function make_defaults(lua_version, target_cpu, platforms, home)
       defaults.variables.LDFLAGS = os.getenv("LDFLAGS")
 
       defaults.cmake_generator = "Unix Makefiles"
-      defaults.variables.CC = "gcc"
-      defaults.variables.LD = "gcc"
+      defaults.variables.CC = os.getenv("CC") or "gcc"
+      defaults.variables.LD = os.getenv("CC") or "gcc"
       defaults.gcc_rpath = true
       defaults.variables.LIBFLAG = "-shared"
       defaults.variables.TEST = "test"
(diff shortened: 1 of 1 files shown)


src/luarocks/core/cfg.lua
```diff
@@ -399,8 +399,8 @@ local function make_defaults(lua_version, target_cpu, platforms, home)
       defaults.lib_extension = "so" -- can be overridden in the config file for mingw builds
       defaults.arch = "cygwin-"..target_cpu
       defaults.cmake_generator = "Unix Makefiles"
-      defaults.variables.CC = "echo -llua | xargs gcc"
-      defaults.variables.LD = "echo -llua | xargs gcc"
+      defaults.variables.CC = os.getenv("CC") or "echo -llua | xargs gcc"
+      defaults.variables.LD = os.getenv("LD") or "echo -llua | xargs gcc"
```
daurnimator, line 403: This one is a bit weird

alerque: Seriously, not that this PR added them, but why `echo -llua | xargs gcc` instead of `gcc -llua`?

siffiejoe: My guess is that CC and LD are used to build more complex command lines using string concatenation, and the xargs part adds `-llua` at the end of that command line, which is required by some linkers.

siffiejoe: Which also means that this is probably incorrect, because you can replace the `gcc` with `os.getenv("CC")`, but you still need the `echo` and `xargs` stuff.

alerque: I don't see how that's possible. The `echo foo | xargs bar` construct would not allow for being appended to anything else (it would supplant it and kill off any prior piping), and anything else being append to this construct could just as easily be appended to `bar`. Searching the source code in fact I don't see any usage where a value of just `gcc` wouldn't function the same. It is just concatenating `CC .. " " .. CFLAGS`.

siffiejoe: I assume that LuaRocks does something like `os.execute(CC .. " " .. options_and_files)` but cygwin requires `os.execute(CC .. " " .. options_and_files .. " -llua")`. This is what the `echo` and `xargs` stuff is supposed to facilitate.

alerque: Oh interesting, I do see now. Thanks for the explanation. That's probably to tricky a stunt to pull without a comment about why it is that way, but yes it does get that job done.

```console
$ echo foo | xargs echo bar baz
bar baz foo
```

siffiejoe: How about:

```lua
defaults.variables.CC = "echo -llua | xargs " .. (os.getenv("CC") or "gcc")
```

alerque: What if the `CC` being supplied is `clang` or `cc` or something else that doesn't accept the `-llua` argument or wants it in a different order?

siffiejoe: The linking requirement of the Lua (shared) library is actually a Windows thing. Windows doesn't support exporting symbols from the executable to shared extension libraries like Linux does. The `-l` flag is pretty standard too. The lookup order of symbols in libraries/objects on the command line is specified by POSIX -- some linkers are order-independent as an extension. `clang` and `gcc` (including MinGW) should be fine, and I don't know what other C compilers we are likely to encounter in a cygwin environment.

siffiejoe: And you can still modify `variables.CC` in your LuaRocks config.

src/luarocks/core/cfg.lua
```diff
@@ -372,8 +372,8 @@ local function make_defaults(lua_version, target_cpu, platforms, home)
       defaults.variables.LDFLAGS = os.getenv("LDFLAGS")
 
       defaults.cmake_generator = "Unix Makefiles"
-      defaults.variables.CC = "gcc"
-      defaults.variables.LD = "gcc"
+      defaults.variables.CC = os.getenv("CC") or "gcc"
+      defaults.variables.LD = os.getenv("LD") or "gcc"
```
Artturin, line 376: with this patch the build of luasystem fails

LD=ld
```
ld  -shared -o system/core.so src/core.o src/compat.o src/time.o -L/nix/store/6f66prpgx1qx4n6k450sxs3d157ia1ps-glibc-2.35-163/lib -Wl,-rpath,/nix/store/6f66prpgx1qx4n6k450sxs3d157ia1ps-glibc-2.35-163/lib -lrt
/nix/store/vhf2cr6immz4qdxd83y025fnxim8mmfg-binutils-2.38/bin/ld: unrecognized option '-Wl,-rpath,/nix/store/6f66prpgx1qx4n6k450sxs3d157ia1ps-glibc-2.35-163/lib'
/nix/store/vhf2cr6immz4qdxd83y025fnxim8mmfg-binutils-2.38/bin/ld: use the --help option for usage information

Error: Build error: Failed compiling module system/core.so
```

LD=gcc
```
gcc  -shared -o system/core.so src/core.o src/compat.o src/time.o -L/nix/store/...-glibc-2.35-163/lib -Wl,-rpath,/nix/store/...-glibc-2.35-163/lib -lrt
```

`-Wl,rpath... -lrt` are added by https://github.com/luarocks/luarocks/blob/ecb63347bf7d44c68210befc9d03c013350c7831/src/luarocks/build/builtin.lua#L249-L250

siffiejoe: The patch is not really the problem. The issue here is that LuaRocks expects a gcc-compatible linker frontend in `variables.LD`. You can try something like: LD=`gcc -fuse-ld=gold`. The linker you choose still needs to support the `-rpath` command line option though if  `cfg.gcc_rpath` is set to true.

hishamhm: @siffiejoe Indeed, but at this point I think changing this behavior in LuaRocks 3.x would cause more compatibility issues than improvements. We could switch to

```lua
      defaults.variables.LD = os.getenv("CC") or "gcc"
```

and make it a  gradual and less-breaking improvement.

src/luarocks/core/cfg.lua
hishamhm: I like this: 
```
defaults.variables.CC = "echo -llua | xargs " .. (os.getenv("CC") or "gcc")
```

src/luarocks/core/cfg.lua
```diff
@@ -518,7 +518,7 @@ local function make_defaults(lua_version, target_cpu, platforms, home)
       defaults.arch = "netbsd-"..target_cpu
    elseif platforms.solaris then
       defaults.arch = "solaris-"..target_cpu
-      defaults.variables.MAKE = "gmake"
+      defaults.variables.MAKE = os.getenv("MAKE") or "gmake"
```
hishamhm, line 521: I'm concerned about the changes to the BSD-ish platforms. We are defaulting to `gmake` because Makefile-based rockspecs are mostly written by Linux users who don't stick to "standard POSIX make" limitations. Defaulting to `gmake` in those platforms has proven to be a less bumpy experience for end-users.

hishamhm (changes requested):
I like the general direction of this PR (some platforms like FreeBSD already used `os.getenv` to read from this system, this standardizes it further). I think this will be good to merge after changes  in the three points that were raised in review comments:

* keeping the C compiler as the default for `LD`
* keeping the `xargs` trick working on Cygwin
* not overriding `gmake` in platforms where we made this the default

src/luarocks/core/cfg.lua
alerque: Do most BSD environments where this will run have a `MAKE` var just floating around that would commonly override this, or would it only be set when set deliberately?

alerque: If this is ever run from *inside* a Makefile then yes `MAKE` will be set to the make that ran it, which is probably not good.

Artturin: removed the getenv for systems which use gmake

src/luarocks/core/cfg.lua
Artturin: > ```lua
>       defaults.variables.LD = os.getenv("CC") or "gcc"
> ```


applied

src/luarocks/core/cfg.lua
Artturin: > I like this:
> 
> ```
> defaults.variables.CC = "echo -llua | xargs " .. (os.getenv("CC") or "gcc")
> ```

applied thanks

src/luarocks/core/cfg.lua
```diff
@@ -287,12 +287,12 @@ local function make_defaults(lua_version, target_cpu, platforms, home)
...
-      defaults.variables.MAKE = "nmake"
-      defaults.variables.CC = "cl"
-      defaults.variables.RC = "rc"
-      defaults.variables.LD = "link"
-      defaults.variables.MT = "mt"
-      defaults.variables.AR = "lib"
+      defaults.variables.MAKE = os.getenv("MAKE") or "nmake"
+      defaults.variables.CC = os.getenv("CC") or "cl"
+      defaults.variables.RC = os.getenv("WINDRES") or "rc"
+      defaults.variables.LD = os.getenv("CC") or "link"
```
Artturin, line 293: should this not use the env var

hishamhm: ah, good catch! — I think for platforms where you do get a distinct linker by default it should be fine to keep the platform's default expected linker variable (is it usually `LD` or `LINK` on Windows with MSVC? I don't know)

Artturin: no idea, i've removed the getenv for it

hishamhm: From what I can quickly guess from the web, looks like the most likely variable in that case would be "LINK". I'll merge this PR and add that!

hishamhm:
Merged manually! Thank you for your contribution and for addressing the reviews!

PR #1435, opened 2022-08-25
tests: Use upstream provided GH Action for Luacheck

alerque:
This will run faster and has less moving parts that need occasional attention.


hishamhm:
cool, thx!

PR #1445, opened 2022-09-27
WIP: Detect and config for SerenityOS

dlemos:
SerenityOS is a Unix-like operating system built from scratch. LuaRocks build and run on the build directory but I still can't get it to run after installing it.

### DIFF
--- a/src/luarocks/core/sysdetect.lua
+++ b/src/luarocks/core/sysdetect.lua
@@ -192,6 +192,11 @@ local function detect_elf_system(fd, hdr, sections)
             return "linux"
          end
       end
+
+      local uname_s = io.popen("uname -s"):read("*l")
+      if uname_s == "SerenityOS" then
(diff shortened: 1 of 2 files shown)


src/luarocks/core/sysdetect.lua
```diff
@@ -192,6 +192,11 @@ local function detect_elf_system(fd, hdr, sections)
             return "linux"
          end
       end
+
+      local uname_s = io.popen("uname -s"):read("*l")
```
hishamhm, line 196: The goal of the sysdetect module is to implement OS and architecture detection without forking processes. There should be a way to detect the OS and architecture uniquely by verifying some fixed location in the filesystem or the identifying headers of well-known executables.

hishamhm:
@dlemos Thank you for the initiative! I gave the PR a first look and left you a review comment.

PR #1484, opened 2023-01-10
WIP: Add Unix manual pages

### LINKED ISSUE #1483: Add Unix manual page
- **Description:**
  - Luarocks should include a Unix manual page that has most information one would need to know about using the `luarocks` command-line utility. This could then be viewed on demand by running `man luarocks`.

jtbx:
This pull request adds Unix manual pages for each `luarocks` command.
A large amount of the content in these manuals are derived from the output of the `luarocks help` command.

I have a few questions:
 - Temporarily, I have put the manual pages in /man. What directory *should* they be in?
 - In the luarocks.1 manual page, I ordered the commands by how useful I think they are, and how much they will be used. Is this okay?

I used the OpenBSD [mdoc](http://man.openbsd.org/mdoc) format over groff, as it is easier to write for manuals and requires no preprocessing.

Closes #1483.

### DIFF
--- a/man/luarocks-build.1
+++ b/man/luarocks-build.1
@@ -1,55 +1,91 @@
 .Dd $Mdocdate: January 11 2023 $
-.Dt LUAROCKS-INSTALL 1
+.Dt LUAROCKS-BUILD 1
 .Os
 .Sh NAME
-.Nm luarocks-install
-.Nd install a rock
+.Nm luarocks-build
+.Nd build and install a rock
 
 .Sh SYNOPSIS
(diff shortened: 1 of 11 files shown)


daurnimator:
> A large amount of the content in these manuals are derived from the output of the `luarocks help` command.

`luarocks --help` is derived using argparse  (https://github.com/luarocks/argparse).
Should that library perhaps gain some sort of manpage/groff output?

jtbx:
@daurnimator I don't know much about that library, but I think it sounds like a good idea.

PR #1490, opened 2023-02-26
upload.api: using CURLNOCERTFLAG in curl_cmd instead of hardcoded -k

imolein:


hishamhm:
@imolein Thanks for the patch!

Issue #1443, opened 2022-09-18
"Failed producing checksum: no MD5 checker tool available, please install md5sum, openssl or md5 in your system" but a md5 checker is installed.


MiguelEXE:
### Please fill out the following:
- **Platform:**
  - Windows

- **LuaRocks version:**
  - C:\Program Files (x86)\LuaRocks\luarocks.lua 3.9.1

- **Configuration file:**
  - https://pastebin.com/9hrFyMTv

- **LuaRocks output from when the issue occurred:**
  - https://pastebin.com/XhTiv9K6

- **Description:**
  - Open command prompt in administrator mode, enter in the %userprofile% (home directory) and run `luarocks install argparse`

Video: (problem)
https://user-images.githubusercontent.com/52385139/190881133-b405dd93-1cb7-45c2-a414-477873429e88.mp4

Video: (verbose)

https://user-images.githubusercontent.com/52385139/190881280-4217dbfe-7d7e-4964-91a3-b635b68920d4.mp4

Image: (tools)
![image](https://user-images.githubusercontent.com/52385139/190881317-119d69f0-0492-4475-bd57-93b8cccfad2f.png)

aflyingdoge:
i also meet this problem, do u fix it?

vaanwd:
There is no fix for this?

NickGrease:
Same problem.

alerque:
Possibly a duplicate of #1001.

benpinpop:
Encountering the same issue, even though there's an md5 checker. I'm attempting to install luasocket. Is there any information I can provide about my system to help fix this?
![image](https://user-images.githubusercontent.com/103378040/236219129-17ae0f7c-f032-4324-bfa8-6a991b13ca50.png)

pyorokun7:
So, this issue has been around for nearly a year, is still present in 3.9.2, but seems fixed.

Is there a workaround I can do as an user, should I downgrade to 3.8.0 or is there a timeline for a release that include the fix?

roman-orekhov:
@pyorokun7 
Replace this line 

> https://github.com/luarocks/luarocks/blob/9a3c5a879849f4f411a96cf1bdc0c4c7e26ade42/src/luarocks/fs/lua.lua#L129

with these
```Lua
         local tool_cmd_no_args_normalized = dir.path(tool_cmd_no_args)
         local fd = io.open(tool_cmd_no_args_normalized, "r")
```

(I don't remember where I got this fix)

[...]

PR #1498, opened 2023-03-29
Strip quotes for tools in paths on Windows

dundargoc:
As mentioned by @erw7 in
https://github.com/luarocks/luarocks/issues/1443#issuecomment-1483816481,
quotes is required when using io.popen but causes problems when using
io.iopen. This makes luarocks unable to find its own md5sum.exe it is
shipped with.

Fixes https://github.com/luarocks/luarocks/issues/1443
Fixes https://github.com/neovim/neovim/issues/22752

Co-authored-by: erw7 <erw7.github@gmail.com>

### DIFF
--- a/src/luarocks/fs/lua.lua
+++ b/src/luarocks/fs/lua.lua
@@ -126,7 +126,8 @@ function fs_lua.is_tool_available(tool_cmd, tool_name)
 
       -- if it looks like the tool has a pathname, try that first
       if tool_cmd_no_args:match("[/\\]") then
-         local fd = io.open(tool_cmd_no_args, "r")
+         local tool_cmd_no_args_normalized = dir.path(tool_cmd_no_args)
+         local fd = io.open(tool_cmd_no_args_normalized, "r")
          if fd then
             fd:close()
             ok = true


erw7:
This change causes problems on UNIX because it is not quoted as follows

https://github.com/luarocks/luarocks/blob/9a3c5a879849f4f411a96cf1bdc0c4c7e26ade42/src/luarocks/core/cfg.lua#L254

If you want to convert on one line `:gsub('^(["\']?)(.-)%1$', '%2')` can be used. However, this would be difficult to understand when viewed at a later date, and therefore difficult for maintenance personnel to accept.

[dir.path()](https://github.com/luarocks/luarocks/blob/9a3c5a879849f4f411a96cf1bdc0c4c7e26ade42/src/luarocks/core/dir.lua#L25-L34) calls [unquote()](https://github.com/luarocks/luarocks/blob/9a3c5a879849f4f411a96cf1bdc0c4c7e26ade42/src/luarocks/core/dir.lua#L7-L14) internally, so you may want to use this. Alternatively, [unquote()](https://github.com/luarocks/luarocks/blob/9a3c5a879849f4f411a96cf1bdc0c4c7e26ade42/src/luarocks/core/dir.lua#L7-L14) could be changed to a public function. Which is better depends on the opinion of the maintenance staff.

justinmk:
@hishamhm anything blocking this, are there concerns or questions?

hishamhm:
Thank you @dundargoc for the PR and @erw7 for the review!

PR #1510, opened 2023-05-27
Make instructions in script match documents

Huell-Howitzer:
This makes it clearer, and more specific to the user that they should **run** `make` and `sudo make install`.


hishamhm:
This PR is not fully correct: `sudo` is not necessarily needed, if a custom prefix with proper permissions is used.

Honestly, I'm not keen on including this PR because if a user doesn't know that `Type: make` means running the `make` command, that is going to be the least of their problems using the `luarocks` CLI tool. It does assume a good level of familiarity with the command-line.

(Also, two lines below there's another `Type:` in there which went unchanged, so even if we were to adopt this change, this PR would be incomplete.)

Still, thank you for taking the time to make a PR!

PR #1517, opened 2023-06-28
update domain name of mirror (luafr.org -> loadk.com)

catwell:
It is the same server, but due to registrar-related changes I probably won't renew the luafr.org domain name. OTOH loadk.com is paid through 2026 and almost certainly not going anywhere after than. :)

### DIFF
--- a/src/luarocks/core/cfg.lua
+++ b/src/luarocks/core/cfg.lua
@@ -200,7 +200,7 @@ local function make_defaults(lua_version, target_cpu, platforms, home)
          {
            "https://luarocks.org",
            "https://raw.githubusercontent.com/rocks-moonscript-org/moonrocks-mirror/master/",
-           "https://luafr.org/luarocks/",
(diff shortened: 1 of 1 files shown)


alerque (approved):
Confirmed DNS points to the same server, but a 3rd party could easily do the same. WHOIS data is blocked on both so it's a little hard to confirm anything there. Presumably pointing to the same server combined with the sites themselves both pointing to @catwell's domain make this pretty legit, but if the sever happened to be a shared hosting provider it would be possible for this to be a spoof ;-) Given the trusted GitHub user info here I think we're good, just commenting with my OpSec hat on.

catwell:
:wave: Can someone please merge this before the next release?

I have [updated the actual luafr page](https://luafr.org) in case that helps to prove this is not malicious. If that isn't enough @hishamhm can shoot me an email to confirm. :)

catwell:
Given that there was no release, I have renewed the domain for one more year.

PR #1519, opened 2023-07-05
luajit: ignore extra version info in jit.version

wojas:
Ignore any extra version info in jit.version when separated by a space.

A normal LuaJIT jit.version string looks like "LuaJIT 2.1.0-beta3". Since official LuaJIT releases have all but stopped, various forks continue to use the same version for all forks.

This change allows forks and patched rebuilds to add additional version information at the end of the string without breaking luarocks version detection, e.g. "LuaJIT 2.1.0-beta3 some-extra-version-info".

### DIFF
--- a/src/luarocks/util.lua
+++ b/src/luarocks/util.lua
@@ -457,7 +457,8 @@ do
 
       local ljv
       if cfg.lua_version == "5.1" then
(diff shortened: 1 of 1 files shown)


hishamhm:
@wojas thank you!! merged

PR #1536, opened 2023-09-25
admin: remove now works with the file protocol

0x501D:
`admin add` can add rockspec using the file protocol without any problems: ./bin/luarocks-admin add testapp-scm-1.rockspec --server '/tmp/rocks/orig'

But deletion only works using the rsync protocol.
This patch adds deletion via file protocol.

### DIFF
--- a/src/luarocks/admin/cmd/remove.lua
+++ b/src/luarocks/admin/cmd/remove.lua
@@ -37,9 +37,6 @@ local function remove_files_from_server(refresh, rockfiles, server, upload_serve
    if not local_cache then
       return nil, protocol
    end
-   if protocol ~= "rsync" then
-      return nil, "This command requires 'rsync', check your configuration."
-   end
 
    local ok, err = fs.change_dir(at)
    if not ok then return nil, err end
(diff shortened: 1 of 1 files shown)


hishamhm:
@0x501D Thank you, merged!

src/luarocks/admin/cmd/remove.lua
```diff
@@ -68,6 +65,17 @@ local function remove_files_from_server(refresh, rockfiles, server, upload_serve
    util.printout("Updating index.html...")
    index.make_index(local_cache)
 
+   if protocol == "file" then
+       local cmd = cfg.variables.RSYNC.." "..cfg.variables.RSYNCFLAGS.." --delete "..local_cache.."/ ".. server_path.."/"
```
hishamhm, line 69: @0x501D  After merging I realized that the `file` protocol mode still uses the RSYNC command. Would it be possible to just do this using `fs.delete`?...

0x501D: I think I can do this

PR #1541, opened 2023-10-30
Fix problems in path normalisation algorithm

AlgebraicWolf:
Current implementation of path normalisation contains various flaws:

- Trailing `.` and `..` at the end of path don't get normalised,
- Path `/` turns into an empty string after normalisation.

This patch changes implementation of normalisation to fix these.


hishamhm:
Thank you, @AlgebraicWolf ! Merged!

Issue #1549, opened 2023-12-04
Crash when using --tree on musl


leso-kn:
Hello :)

There seems to be a bug when working with `--tree`s on musl:

Luarocks crashes for any package with sub-dependencies (specifically in `fulfill_dependency()`) and after lots of digging, the issue appears to be, that the value passed to `--tree` is never added to `cfg.rocks_trees`.

**Context and solution**
<details>

</br>

Trying this on several different platforms (Ubuntu 22.04 with luarocks 3.9.0, Msys2 with luarocks 3.9.2) shows the expected behaviour without any issue, so the problem seems to be musl specific – though telling by the [related code](https://github.com/luarocks/luarocks/blob/4cfcf9d8df1abd7cca9f2c3590402ada543df327/src/luarocks/cmd.lua#L84), I'm not sure how the specified tree would be added to `cfg.rocks_trees` on any platform.  
(which should be required for fulfill_dependency() to find installed dependencies – [core/path.lua:143](https://github.com/luarocks/luarocks/blob/4cfcf9d8df1abd7cca9f2c3590402ada543df327/src/luarocks/core/path.lua#L143C13-L143C13))

Adding a `table.insert(cfg.rocks_trees, ..)` for the value passed in `--tree` fixes the crash, so I will submit a PR – in case this is the intended behaviour.
</details>

**Steps to reproduce**  
(in docker or on a machine running a musl based system e.g. Alpine Linux):

```bash
> docker run --rm -ti alpine:edge
> apk add luarocks5.1 lua5.1-dev build-base  # happens on other lua versions as well (e.g. lua5.3)
> luarocks-5.1 --tree=custom/luatree install luacheck

# crash
```

**Environment**
- **Platform:**
  - Linux/Alpine Linux amd64 (musl)

- **LuaRocks version:**
  - 3.9.2

- **Configuration file:**
  - none (clean docker container)

- **LuaRocks output from when the issue occurred:**
  <details>

  ```log
  Error: LuaRocks 3.9.2 bug (please report at https://github.com/luarocks/luarocks/issues).
  Arch.: linux-x86_64
  /usr/share/lua/5.3/luarocks/deps.lua:243: assertion failed!
  stack traceback:
  	[C]: in function 'assert'
  	/usr/share/lua/5.3/luarocks/deps.lua:243: in function 'luarocks.deps.fulfill_dependency'
[...]

PR #1550, opened 2023-12-04
Fix crash when using --tree on musl

leso-kn:
This PR addresses the crash described in #1549 and adds the value passed to the `--tree` argument to `cfg.rocks_trees`.
Tested and working on `alpine:edge` running `luarocks-5.1` v3.9.2.

### DIFF
--- a/src/luarocks/cmd.lua
+++ b/src/luarocks/cmd.lua
@@ -83,6 +83,9 @@ do
          if not named then
             local root_dir = fs.absolute_name(args.tree)
             replace_tree(args, root_dir)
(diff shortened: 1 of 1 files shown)


hishamhm:
The combination of all possible configurations and flags added over the year to tend for numerous scenarios and use-cases often end up leading to some surprising behaviors... A big cleanup of the tree management code would do it good, but given backward compatibility constraints, I prefer keeping changes to this logic minimal for now.

Since that you reported this fixes a crash and the test suite seems to happy about it, I am going to merge your change, thanks!

PR #1563, opened 2024-01-24
cmd: 'server' option change

better0fdead:
Changed 'server' option, so it now can take several URLs.
e.g '--server 'URL1 URL2 URL3''.

### DIFF
--- a/src/luarocks/cmd.lua
+++ b/src/luarocks/cmd.lua
@@ -126,8 +126,12 @@ end
 
 local function process_server_args(args)
    if args.server then
-      local protocol, pathname = dir.split_url(args.server)
-      table.insert(cfg.rocks_servers, 1, protocol.."://"..pathname)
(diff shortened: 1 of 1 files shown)


src/luarocks/cmd.lua
```diff
@@ -455,9 +459,10 @@ Enabling completion for Fish:
       end)
    parser:flag("--dev", "Enable the sub-repositories in rocks servers for "..
       "rockspecs of in-development versions.")
-   parser:option("--server", "Fetch rocks/rockspecs from this server "..
+   parser:option("--server", "Fetch rocks/rockspecs from this servers "..
```
hishamhm, line 462: I think there is an method in argparse to allow an `option` to be passed multiple times (resulting in an array of values). I think that would be preferrable to relying on splitting by spaces.

PR #1569, opened 2024-02-16
fix: `build.install_command` doesn't execute on windows

### LINKED ISSUE #1568: `build.install_command = "..."` does not execute on windows
When installing a package, which uses the command build type, luarocks doesn't run the command specified in `install_command`.

- **Platform:**
  - Windows

- **LuaRocks version:**
  - `luarocks 3.9.2`

- **Configuration file:**
```lua
-- example-scm-0.rockspec`
...
 build = {
   type = "command",
   install_command = "echo test";
}
```
```
> luarocks build

echo test
No existing manifest. Attempting to rebuild...
example scm-0 is now installed in  C:/msys64/ucrt64 (license: *** please specify a license ***)
```
> **Observe:** "test" was not printed -> `echo` was not executed.

sewbacca:
This PR fixes #1568.

The issue was that `os.execute()` on windows does not support multiple commands seperated by newlines.
Instead it's required to use the `cmd.exe`-syntax: `command1 & command2`.

This is now implemented in `win32.execute_env()`.

### DIFF
--- a/src/luarocks/fs/lua.lua
+++ b/src/luarocks/fs/lua.lua
@@ -54,7 +54,7 @@ function fs_lua.is_writable(file)
    return result
 end
 
-local function quote_args(command, ...)
+function fs_lua.quote_args(command, ...)
    local out = { command }
    local args = pack(...)
    for i=1, args.n do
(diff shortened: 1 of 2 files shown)


hishamhm:
@sewbacca Thank you!

Issue #1001, opened 2019-04-19
Unable to reach md5sum.exe ?


DiegoAlpizar:
Can't install anything because of "No MD5 checking tool available"

___________
Windows 10.
Lua 5.1 and luarocks 3.0.4 installed with luarocks-3.0.4-win32 (`INSTALL /L`) from developer command prompt for Visual Studio 2019

Set my paths as the last output of the install script suggested:
```bat
LUA_CPATH=C:\Users\e\AppData\Roaming\LuaRocks\lib\lua\5.1\?.dll
LUA_PATH=C:\Users\e\AppData\Roaming\LuaRocks\share\lua\5.1\?.lua;C:\Users\e\AppData\Roaming\LuaRocks\share\lua\5.1\?\init.lua
[...]

PR #1593, opened 2024-02-19
vendor in a fallback pure-Lua md5 implementation

hishamhm:
Vendor in md5.lua from https://github.com/kikito/md5.lua as a fallback.

Specifically:
https://github.com/kikito/md5.lua/blob/fa8e8d53e2483af7029659d492af574965ffb808/md5.lua

See #1001.


hishamhm:
This cannot be merged as-is, since the pure Lua md5 is unbearably slow. It might still be a good idea to have it as a fallback, but it still must come after trying a local binary implementation.

PR #1618, opened 2024-02-27
fix: on windows environment variables are passed with trailing space

sewbacca:
This fix addresses a mistake I've made in the PR #1569. `cmd`s `SET key=value` syntax assigns to key everything behind `=` until the next `&`, even if it contains spaces:
```shell
> cmd /C type NUL && SET FOO=BAR & lua -e "print(string.format('%q', os.getenv 'FOO'))"                                                 
"BAR "
```
Either one quotes the whole argument:
```shell
> cmd /C type NUL && SET "FOO=BAR" & lua -e "print(string.format('%q', os.getenv 'FOO'))"
"BAR"
```
Or we remove the extra spaces:
```shell
cmd /C type NUL && SET FOO=BAR&lua -e "print(string.format('%q', os.getenv 'FOO'))"                                                   
"BAR"
```
I opted for the first option, which appears safer, considering that we quote the argument via `fs.Q`.

PR #1629, opened 2024-02-29
tests: use paths with spaces when testing on Windows

hishamhm:
This should force us to test our compatibility with paths with spaces all over.

See #1618.


hishamhm:
This would require too big of an overhaul of the test suite in order to work, more effort than I'm able to put into right now.
So I'm closing this now, but it would be great to be able to revisit this idea in the future.

PR #1686, opened 2024-06-10
configure: minor version check for Lua 5.4.7-rc3 onwards

### LINKED ISSUE #1685: Latest lua 5.4.7 RC breaks luarocks configure grep for installed headers
- **Platform:**
  - Linux

- **LuaRocks version:**
  - 3.11.1

- **Configuration file:**
  - cannot install

  - cannot install

- **Description:**

As [reported on the lua mailing list](https://groups.google.com/g/lua-l/c/xqy9tf9dASQ) by **sur-behoffski**, Lua 5.4.7 (as of the lua-5.4.7-rc3 release candidate) changes the format of the version `#defines` in `lua.h` in a way that breaks the simplistic `grep` in luarocks' `configure` script.

Previously, the `lua.h` header contained a line similar to:

```c
#define LUA_VERSION_NUM 504
```

which matched the 'configure' script's check:

https://github.com/luarocks/luarocks/blob/a51f0bfdfe940200fe77db43103e3065a1cfb030/configure#L431-L441

But `lua-5.4.7-rc3` changes that `#define` to:

```c
#define LUA_VERSION_NUM (LUA_VERSION_MAJOR_N * 100 + LUA_VERSION_MINOR_N)
```

which the `grep` fails to match.

hishamhm:
The previous check served us well for many years, for both Lua and LuaJIT.

A more correct solution would use the C preprocessor to read the value as suggested by @CDSoft, but the change in this commit is the simplest possible check to adapt for the new versions, while not requiring the C toolchain to be present at installation time.

Fixes #1685.

### DIFF
--- a/configure
+++ b/configure
@@ -438,6 +438,8 @@ check_incdir() {
    do
       if [ -f "$lua_h" ]
       then
+         lua_minor=$(echo "$LUA_VERSION" | cut -b 3-)
+         grep "LUA_VERSION_MINOR_N.*$lua_minor" "$lua_h" > /dev/null 2> /dev/null && return
          grep "LUA_VERSION_NUM.*$LUA_VERSION" "$lua_h" > /dev/null 2> /dev/null && return
       fi
       tried="$tried $lua_h"
--- a/src/luarocks/deps.lua
+++ b/src/luarocks/deps.lua
@@ -703,6 +703,9 @@ local function lua_h_exists(d, luaver)
       if data:match("LUA_VERSION_NUM%s*" .. tostring(luanum)) then
          return d
       end
+      if data:match("LUA_VERSION_MINOR_N%s*" .. tostring(minor)) then
+         return d
+      end
       return nil, "Lua header lua.h found at " .. d .. " does not match Lua version " .. luaver .. ". You can use `luarocks config variables.LUA_INCDIR <path>` to set the correct location.", "dependency", 2
    end
 


src/luarocks/deps.lua
```diff
@@ -703,6 +703,9 @@ local function lua_h_exists(d, luaver)
       if data:match("LUA_VERSION_NUM%s*" .. tostring(luanum)) then
          return d
       end
+      if data:match("LUA_VERSION_MINOR_N%s*" .. tostring(minor)) then
```
ferdnyc, line 706: Should this also check `LUA_VERSION_MAJOR_N`, so that it's both equivalent and Lua-6-ready?

configure
```diff
@@ -438,6 +438,8 @@ check_incdir() {
    do
       if [ -f "$lua_h" ]
       then
+         lua_minor=$(echo "$LUA_VERSION" | cut -b 3-)
+         grep "LUA_VERSION_MINOR_N.*$lua_minor" "$lua_h" > /dev/null 2> /dev/null && return
```
ferdnyc, line 442: Same question here.

src/luarocks/deps.lua
hishamhm: A *lot* of other things would break in LuaRocks with a hypothetical Lua 6. Actually, even for Lua 5.5 it would require manual intervention to make LuaRocks accept the new version.

Given Lua's evolution history, LuaRocks is not written using the assumption than a 5.(x+1) would ever work out-of-the-box. There were changes in past Lua versions which included standard library functions remaining syntactically compatible but semantically changed. For this reason, we actually prevent an unknown version from running, to avoid having it "almost work" by accident and then fail in insidious ways.

ferdnyc: @hishamhm Understood, but that's sort of my point.

In the incredibly unlikely event that Lua's release cadence massively accelerates, **in _theory_** if this check were to encounter a future Lua 6.4 release, it would accept it under the assumption that it's actually Lua 5.4.

ferdnyc: (When I say "Lua 6 ready", I don't mean "Lua 6 compatible" so much I mean ready to _reject_ Lua 6.x releases as **incompatible**. Right now, the new checks can't differentiate between Lua 6.1 and Lua 5.1, or Lua 7.2 and Lua 5.2.

hishamhm: Fair point!

hishamhm:
Lua 5.4.7-rc4 was released, which reverted the change in `lua.h`. (It was not just LuaRocks; apparently various tools reported issues with the change in `lua.h`.)

I am going to close this without applying, since I do not find it worth it to add logic just for a shortlived rc3.

ferdnyc:
Sanity prevails! Excellent news, thanks @hishamhm.

kevinkirkup:
Looks like this is still broken in the final `Lua 5.47` release.
Any updates?

PR #1688, opened 2024-06-13
fix: use realpath for find_program()

Freed-Wu:


Tieske:
I'm not sure this is safe. Until recently macOS didn't have realpath, had to be installed separately.

hishamhm:
hi @Freed-Wu, thanks for the PR! Can you give a bit more context on why you needed this?

Also, `realpath` is not very portable in non-Linux systems. We do have a `canonicalpath` function in the `configure` script that could be used in its stead.

Freed-Wu:
In some case, we have a symlink to real lua to allow system exists many lua in different position.

hishamhm:
@Freed-Wu could you please confirm that changing `realpath` to `canonicalpath` in that line would also work for your use-case? Thank you!

Freed-Wu:
No, It will give me 'lua.h for Lua 5.1 not found'

Freed-Wu:
Does macos have `readlink`?

`readlink -f` can get the path of symlink.

Issue #1662, opened 2024-04-08
luarocks.lock in cwd not respected by `luarocks install`


mrcjkb:
### Please fill out the following:
- **Platform:**
  - Linux/NixOS

- **LuaRocks version:**
  - 3.11.0

- **Configuration file:**
 - N/A

- **LuaRocks output from when the issue occurred:**
  - N/A

- **Description:**

From [this documentation](https://github.com/luarocks/luarocks/wiki/Pinning-versions-with-a-lock-file#using-pinned-dependencies-in-luarocks):

> When building a package with luarocks build, luarocks make (or via luarocks install if there is not prebuilt binary package), without using --pin, if the current directory contains a luarocks.lock file, it is used as the authoritative source [...]
>
> When building a package using a lock file, luarocks.lock is copied to the package's metadata directory (e.g. /usr/local/luarocks/rocks/5.3/name/version/luarocks.lock)

### Actual behaviour:

```bash
mkdir rocks
echo "return { dependencies = { ['nvim-web-devicons'] = '0-0', }, }" > luarocks.lock
luarocks install --tree rocks oil.nvim #depends on nvim-web-devicons
```

- nvim-web-devicons 0.99-1 is installed.

I had a look at the source for `install`

https://github.com/luarocks/luarocks/blob/15d71214638c6f7f808b838cfc4215ece0a53101/src/luarocks/cmd/install.lua?plain=1#L127

and it looks like it always passes the `install_dir` as the `deplock_dir` to `fulfill_dependencies`.

### Expected behaviour

Based on the documentation, I would expect luarocks ~to check if the `install_dir` contains a `luarocks.lock` file, and if not,~ to pass the `cwd` to `fulfill_dependencies`, and only pass the `install_dir` if the `cwd` doesn't contain a luarocks.lock file.

Maybe a nicer solution would be if `luarocks install` had a `--lockfile <path-to-luarocks.lock>` flag.

mrcjkb:
This is fixed with version 3.11.1 when using the package installed with gnumake.

```
Using dependencies pinned in lockfile: luarocks.lock
oil.nvim 2.7.0-1 is pinned to nvim-web-devicons 0.99-1 (not installed)
Installing https://luarocks.org/nvim-web-devicons-0.99-1.src.rock
No existing manifest. Attempting to rebuild...
nvim-web-devicons 0.99-1 is now installed in /tmp/rocks (license: MIT)
[...]

PR #1693, opened 2024-07-08
fix: use luarocks.lock in cwd for non-binary rocks too

mrcjkb:
This completes the fix for #1692.

See [this comment on the issue](https://github.com/luarocks/luarocks/issues/1662#issuecomment-2212547760).


hishamhm:
@mrcjkb Merged, thank you!

mrcjkb:
@hishamhm thanks for merging :smile: 

Do you have an ETA on the next luarocks release?

PR #1699, opened 2024-07-25
Update pe-parser to version v0.6

### LINKED ISSUE #1697: Windows install script sets incorrect runtime if lua was linked against UCRT
- **Platform:** Windows, **LuaRocks version:**  3.8.0

This is the relevant part of the install.bat output:
```
Headers found, checking runtime to use...
*** WARNING ***: could not analyse the runtime used, defaulting to MSVCR80
Runtime check completed.
```

- **Description:**
  - I ran into this error in a windows-2022 environment on github actions, while using hererocks (`hererocks lua_install -r^ --lua=5.4 --verbose`). Lua binaries that are built with VS 2022 seem to not be linked against any msvcrt.dll file, so the install script reverts to MSVCRT80: https://github.com/luarocks/luarocks/blob/f87568a7cba69e2c19411e7a2b73016c80a14a3b/install.bat#L406
  - This means that the config is set with `variables.MSVCRT = "MSVCR80"`, so this is the library that is linked against. I ran into a problem where a library would fail to load with `require` because of this dll being linked: https://github.com/lunarmodules/luafilesystem/issues/149
  - I did some debugging, these are the dlls that pe-parser.lua does see:
```
KERNEL32.dll
api-ms-win-crt-convert-l1-1-0.dll
api-ms-win-crt-environment-l1-1-0.dll
api-ms-win-crt-filesystem-l1-1-0.dll
api-ms-win-crt-heap-l1-1-0.dll
api-ms-win-crt-locale-l1-1-0.dll
api-ms-win-crt-math-l1-1-0.dll
api-ms-win-crt-private-l1-1-0.dll
api-ms-win-crt-runtime-l1-1-0.dll
api-ms-win-crt-stdio-l1-1-0.dll
api-ms-win-crt-string-l1-1-0.dll
api-ms-win-crt-time-l1-1-0.dll
lua53.dll
KERNEL32.dll
api-ms-win-crt-convert-l1-1-0.dll
api-ms-win-crt-environment-l1-1-0.dll
api-ms-win-crt-filesystem-l1-1-0.dll
api-ms-win-crt-heap-l1-1-0.dll
api-ms-win-crt-locale-l1-1-0.dll
api-ms-win-crt-math-l1-1-0.dll
api-ms-win-crt-private-l1-1-0.dll
api-ms-win-crt-runtime-l1-1-0.dll
api-ms-win-crt-stdio-l1-1-0.dll
api-ms-win-crt-string-l1-1-0.dll
api-ms-win-crt-time-l1-1-0.dll
api-ms-win-crt-utility-l1-1-0.dll
```

tobil4sk:
Closes #1697.

### DIFF
--- a/win32/pe-parser.lua
+++ b/win32/pe-parser.lua
@@ -5,7 +5,7 @@
 -- case of 64 bit fields (bit/flag fields). Pointer arithmetic is still done numerically, so for
 -- very large files this could lead to undefined results. Use with care!
 --
--- Version 0.4, [copyright (c) 2013-2015 Thijs Schreijer](http://www.thijsschreijer.nl)
+-- Version 0.5, [copyright (c) 2013-2018 Thijs Schreijer](http://www.thijsschreijer.nl)
 -- @name pe-parser
 -- @class module
 
@@ -16,7 +16,7 @@ local M = {}
 -- For flag fields the name is extended with `_flags`.
 -- @usage -- lookup descriptive name for the myobj.Magic value
 -- local desc = pe.const.Magic(myobj.Magic)
--- 
+--
 -- -- get list of flag names, indexed by flag values, for the Characteristics field
 -- local flag_list = pe.const.Characteristics_flags
 M.const = {
@@ -100,7 +100,7 @@ M.const = {
       ["800"] = "IMAGE_SCN_LNK_REMOVE",
       ["1000"] = "IMAGE_SCN_LNK_COMDAT",
       ["8000"] = "IMAGE_SCN_GPREL",
-      ["20000"] = "IMAGE_SCN_MEM_PURGEABLE",
+      ["10000"] = "IMAGE_SCN_MEM_PURGEABLE",
       ["20000"] = "IMAGE_SCN_MEM_16BIT",
       ["40000"] = "IMAGE_SCN_MEM_LOCKED",
       ["80000"] = "IMAGE_SCN_MEM_PRELOAD",
@@ -128,7 +128,7 @@ M.const = {
       ["80000000"] = "IMAGE_SCN_MEM_WRITE",
     },
   },
-  
+
 }
 
 
@@ -143,7 +143,7 @@ function M.toHex(IN, len)
         IN,D=math.floor(IN/B),math.fmod(IN,B)+1
         OUT=string.sub(K,D,D)..OUT
     end
-    len = len or string.len(OUT)
+    len = len or #OUT
     if len<1 then len = 1 end
     return (string.rep("0",len) .. OUT):sub(-len,-1)
 end
@@ -193,8 +193,8 @@ local function get_list(list, f, add_to)
   local r = add_to or {}
   for i, t in ipairs(list) do
     assert(r[t.name] == nil, "Value for '"..t.name.."' already set")
-    local val,err = f:read(t.size)  -- read specified size in bytes
-    val = val or "\0"    
+    local val = f:read(t.size)  -- read specified size in bytes
+    val = val or "\0"
     if t.is_str then   -- entry is marked as a string value, read as such
       for i = 1, #val do
         if val:sub(i,i) == "\0" then
@@ -248,25 +248,25 @@ end
 -- local obj = pe.parse("c:\lua\lua.exe")
 -- obj:dump()
 M.parse = function(target)
-  
-  local list = {    -- list of known architectures
-    [332]   = "x86",       -- IMAGE_FILE_MACHINE_I386
-    [512]   = "x86_64",    -- IMAGE_FILE_MACHINE_IA64
-    [34404] = "x86_64",    -- IMAGE_FILE_MACHINE_AMD64
-  }
-  
+
+  -- local list = {    -- list of known architectures
+  --   [332]   = "x86",       -- IMAGE_FILE_MACHINE_I386
+  --   [512]   = "x86_64",    -- IMAGE_FILE_MACHINE_IA64
+  --   [34404] = "x86_64",    -- IMAGE_FILE_MACHINE_AMD64
+  -- }
+
   local f, err = io.open(target, "rb")
   if not f then return nil, err end
-  
+
   local MZ = f:read(2)
   if MZ ~= "MZ" then
     f:close()
     return nil, "Not a valid image"
   end
-  
+
   f:seek("set", 60)                    -- position of PE header position
   local peoffset = get_int(f:read(4))  -- read position of PE header
-  
+
   f:seek("set", peoffset)              -- move to position of PE header
   local out = get_list({
         { size = 4,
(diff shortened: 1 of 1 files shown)


win32/pe-parser.lua
```diff
@@ -525,44 +525,39 @@ function M.msvcrt(infile)
...
-    if rt then 
+    if rt then
       return rt, ref  -- found it
     end
   end
 
   return nil, "No msvcrt found"
 end
 
-function M.get_architecture(program)
```
Tieske, line 553: is this correct to remove? was this maybe a luarocks specific function that was added? not sure though

tobil4sk: Hm, this was added in d90c3c918a2555eff8776e1acabab7a777f13c31, when pe-parser was in `src/luarocks/fs/`. There are no references currently to this function anywhere in the repository so maybe it's fine to remove it, but I'm struggling to figure out why it was added in the first place...

hishamhm: This was moved there because it was part of `install.bat` as part of an effort to move logic out of `install.bat` and into LuaRocks modules proper.

I still think it would be a good idea to move it into pe-parser, essentially as a high-level utility API (note that that version takes a `program` as an argument, versus the one in `install.bat` which hardcodes `vars.LUA`)

win32/pe-parser.lua
```diff
@@ -5,7 +5,7 @@
 -- case of 64 bit fields (bit/flag fields). Pointer arithmetic is still done numerically, so for
 -- very large files this could lead to undefined results. Use with care!
 --
--- Version 0.4, [copyright (c) 2013-2015 Thijs Schreijer](http://www.thijsschreijer.nl)
+-- Version 0.5, [copyright (c) 2013-2018 Thijs Schreijer](http://www.thijsschreijer.nl)
```
hishamhm, line 8: is this pe-parser 0.6? comment still says 0.5

tobil4sk: I took it from the 0.6 tag as is: https://github.com/Tieske/pe-parser/blob/version_0.6/src/pe-parser.lua#L8

win32/pe-parser.lua
Tieske: I don't think that'll work. PE-parser recognizes far more formats, hence the simple implementation for detecting 32/64 bits doesn't suffice. It might for LuaRocks, but not for the parser itself.

Here's the list of archs:
https://github.com/Tieske/pe-parser/blob/acbcd85a337dcbfbda2c7511b53c56666337d3f8/src/pe-parser.lua#L29-L52

As per ChatGPT:

Here are the common names for each of the machine architectures listed in your PortableExecutable introspection data:

- `IMAGE_FILE_MACHINE_UNKNOWN` (0): Unknown machine type
- `IMAGE_FILE_MACHINE_AM33` (1d3): Matsushita AM33
- `IMAGE_FILE_MACHINE_AMD64` (8664): x64 (AMD or Intel)
- `IMAGE_FILE_MACHINE_ARM` (1c0): ARM little endian
- `IMAGE_FILE_MACHINE_ARMNT` (1c4): ARM Thumb-2 little endian
- `IMAGE_FILE_MACHINE_ARM64` (aa64): ARM64 little endian
- `IMAGE_FILE_MACHINE_EBC` (ebc): EFI byte code
- `IMAGE_FILE_MACHINE_I386` (14c): Intel 386 or later processors and compatible processors
- `IMAGE_FILE_MACHINE_IA64` (200): Intel Itanium processor family
- `IMAGE_FILE_MACHINE_M32R` (9041): Mitsubishi M32R little endian
- `IMAGE_FILE_MACHINE_MIPS16` (266): MIPS16
- `IMAGE_FILE_MACHINE_MIPSFPU` (366): MIPS with FPU
- `IMAGE_FILE_MACHINE_MIPSFPU16` (466): MIPS16 with FPU
- `IMAGE_FILE_MACHINE_POWERPC` (1f0): Power PC little endian
- `IMAGE_FILE_MACHINE_POWERPCFP` (1f1): Power PC with floating point support
- `IMAGE_FILE_MACHINE_R4000` (166): MIPS little endian
- `IMAGE_FILE_MACHINE_SH3` (1a2): Hitachi SH3
- `IMAGE_FILE_MACHINE_SH3DSP` (1a3): Hitachi SH3 DSP
- `IMAGE_FILE_MACHINE_SH4` (1a6): Hitachi SH4
- `IMAGE_FILE_MACHINE_SH5` (1a8): Hitachi SH5
- `IMAGE_FILE_MACHINE_THUMB` (1c2): ARM or Thumb ("thumb")
- `IMAGE_FILE_MACHINE_WCEMIPSV2` (169): MIPS little-endian WCE v2

These names are more commonly used and recognizable when referring to these specific machine architectures in the context of Portable Executable files.

The PR only assumes `IMAGE_FILE_MACHINE_I386` to be `x86` (32 bit), and everything else to be `x86_64` (64 bit). And hence misses the ARM architectures, as well Itanium.

I do not see an easy way to add this to pe-parser. So imho this should probably stay within LuaRocks in a simple wrapper.

win32/pe-parser.lua
Tieske: thx @hishamhm fixed those in `master` of `pe-parser`. 

Since there are no code changes, I didn't push a new release. So you can either pick up current master, or stay with this code as it is.

win32/pe-parser.lua
hishamhm: > As per ChatGPT:

Please do not use ChatGPT as a source of technical information in any project I'm involved with. It constantly produces false data, and having to double-check everything it produces is equivalent to just skipping it altogether and going to a reliable source anyway.

This link illustrates another occasion where this has happened to me before, and the "reference" info provided by ChatGPT was a blatant lie:
https://mastodon.social/@hanshuebner/109952292956659419

hishamhm: > The PR only assumes IMAGE_FILE_MACHINE_I386 to be x86 (32 bit), and everything else to be x86_64 (64 bit). And hence misses the ARM architectures, as well Itanium.
> 
> I do not see an easy way to add this to pe-parser. So imho this should probably stay within LuaRocks in a simple wrapper.

Fair point! We can still move it out of `install.bat` in a separate PR, and not mix it with the vendored code of pe-parser.

hishamhm: Turns out that sysdetect (also vendored-in) already does something similar (covering at least the set of architectures our users seem to care about)

https://github.com/luarocks/sysdetect/blob/main/src/sysdetect.tl#L401-L408

hishamhm:
Thank you @tobil4sk for the PR and @Tieske for the review comments!

win32/pe-parser.lua
Tieske: > Please do not use ChatGPT as a source of technical information in any project I'm involved with. It constantly produces false data, and having to double-check everything it produces is equivalent to just skipping it altogether and going to a reliable source anyway.

I totally agree (and I know your stance), hence I specifically mentioned it as chatgpt generated. The only aim here was to list the more common names of the platforms, to proof the point. It wasn't about being technically accurate.

PR #1705, opened 2024-08-22
Convert LuaRocks to Teal

hishamhm:
Google Summer of Code 2024 project by @V1K1NGbg, mentored by @hishamhm 

This is the cleaned-up branch with the changes reorganized for merging into the LuaRocks main branch, based on the work done by @V1K1NGbg in his [working branch](https://github.com/luarocks/luarocks/tree/lua-to-teal-migration).


hishamhm (commented):
@V1K1NGbg I made this revised commit set which organizes the changes file-by-file. I'm going through them one-by-one, and I managed to get to about a third (I stopped at `luarocks.manif`, going by commit order. I will continue from there tomorrow.)

You can make any the changes I mention in the comments here, and push them directly as separate commits to this `teal` branch, then click "resolve conversation" on the comment. I'll then join the commits again in the end to make them ready to merge. (Just name the commit after the file name, to make it easier for me to sort them later!)

So far I must say I'm very happy with how things are turning out! The review changes are all minor, and reading through the codebase with the type information makes it much much easier to follow. Excited to get this to the merge point!!

src/luarocks/deps.tl
```diff
@@ -324,28 +326,28 @@ function deps.fulfill_dependencies(rockspec, depskey, deps_mode, verify, deplock
...
-            util.printerr("   luarocks search "..name)
-         end
+         -- if no_upgrade then
+         --    util.printerr("This version of "..name.." is designed for use with")
+         --    util.printerr(tostring(dep)..", but is configured to avoid upgrading it")
+         --    util.printerr("automatically. Please upgrade "..dep.name.." with")
+         --    util.printerr("   luarocks install "..dep.name)
+         --    util.printerr("or look for a suitable version of "..name.." with")
+         --    util.printerr("   luarocks search "..name)
+         -- end
```
hishamhm, line 350: oh wow, looks like this codepath was broken this whole time! I think I have a fix.

src/luarocks/deps.tl
```diff
@@ -740,8 +755,8 @@ local function find_lua_incdir(prefix, luaver, luajitver)
    return nil, mainerr
 end
 
-function deps.check_lua_incdir(vars)
-   if vars.LUA_INCDIR_OK == true
+function deps.check_lua_incdir(vars: {string: string}): boolean, string, string
+   if vars.LUA_INCDIR_OK == "ok" --! fix later
```
hishamhm, line 759: you can search for `--! fix later` in this file and remove these; we'll keep it like this

src/luarocks/signing.tl
```diff
@@ -3,21 +3,21 @@ local signing = {}
 local cfg = require("luarocks.core.cfg")
 local fs = require("luarocks.fs")
 
-local function get_gpg()
+local function get_gpg(): string, string
    local vars = cfg.variables
-   local gpg = vars.GPG
+   local gpg = vars.GPG --? string from line 30
```
hishamhm, line 8: correct — you can remove the comment

src/luarocks/manif/writer.tl
```diff
@@ -46,15 +55,12 @@ end
...
-   assert(type(items) == "table")
-   assert(type(name) == "string" and not name:match("/"))
-   assert(type(version) == "string")
+local function remove_package_items(storage: {string: {string}}, name: string, version: string, items: {string: string})
+   assert(not name:match("/"))
 
    local package_identifier = name.."/"..version
 
-   for item_name, path in pairs(items) do  -- luacheck: ignore 431
+   for item_name, _ in pairs(items) do  -- luacheck: ignore 431
```
hishamhm, line 63: please change `_` to `_path` and remove the `-- luacheck` comment

src/luarocks/manif/writer.tl
```diff
@@ -259,35 +253,41 @@ local function save_table(where, name, tbl)
...
             next = {}
             walk[filename] = next
          end
          last = walk
          last_name = filename
-         walk = next
+         if not next is string then
+            walk = next
+         else
+            return nil, "Next was string, expected a table"
```
hishamhm, line 276: this is an internal consistency check, so we shouldn't bubble-up an error message as if it was a normal failure condition.

If `next` is a string here, that means it was a LuaRocks bug. So, it's actually better to assert. Teal `next` is smart enough to remember an `is` test inside an assert:

```lua
last_name = filename
assert(next is {string: Entry})
walk = next
```

src/luarocks/remove.tl
```diff
@@ -1,23 +1,28 @@
...
+
 --- Obtain a list of packages that depend on the given set of packages
 -- (where all packages of the set are versions of one program).
 -- @param name string: the name of a program
 -- @param versions array of string: the versions to be deleted.
 -- @return array of string: an empty table if no packages depend on any
 -- of the given list, or an array of strings in "name/version" format.
-local function check_dependents(name, versions, deps_mode)
-   local dependents = {}
+local function check_dependents(name: string, versions: {string: any}, deps_mode: string): {Result}
```
hishamhm, line 24: if I followed the code correctly, this should be `versions: {string: {Result}}`

src/luarocks/cmd.tl
```diff
@@ -12,10 +22,21 @@ local fun = require("luarocks.fun")
...
+local type Args = require("luarocks.core.types.args").Args
+
+local type Module = cmd.Module
+
+local type PersistableTable = require("luarocks.core.types.persist").PersistableTable
 
-local hc_ok, hardcoded = pcall(require, "luarocks.core.hardcoded")
+local type Config = cfg
+
+-- local pack = table.pack or function(...) return { n = select("#", ...), ... } end
```
hishamhm, line 37: can delete this

src/luarocks/cmd.tl
```diff
@@ -595,27 +623,27 @@ function cmd.run_command(description, commands, external_namespace, ...)
 
    local args, cmdline_vars = process_cmdline_vars(...)
    local parser = get_parser(description, cmd_modules)
-   args = parser:parse(args)
+   local argsp: Args = parser:parse(args) as Args --!
```
hishamhm, line 626: instead of naming this `argsp`, rename the other one into `cmdline_args` (because this "argsp" is the same object that appears as "args" in every other cmd module)

src/luarocks/core/types/build.d.tl
```diff
@@ -0,0 +1,66 @@
...
+         defines: {string}
+         incdirs: {string}
+         libdirs: {string}
+      end
+
+      modules: {string: (string | Module)}
+   end
+
+
+   record CMakeBuild
```
hishamhm, line 36: all these records (BuiltinBuild, CMakeBuild, etc.) should be in the luarocks/build module that uses them — let's keep only the abstract interface here

hishamhm (commented):
Phew! I went through them all. In the end, it's fewer comments than the first batch. There were a couple of more complicated changes that would be hard to explain so for those I decided to take the time and make the changes myself; those are extra commits at the bottom of this PR. Everything else here is in the same style as the comments from the first batch. Looking good!

src/luarocks/deps.tl
hishamhm: I pushed a fix, just leaving this open here so you can see it

src/luarocks/cmd.tl
```diff
@@ -623,27 +621,27 @@ function cmd.run_command(description: string, commands: {string: string}, extern
 
    local args, cmdline_vars = process_cmdline_vars(...)
    local parser = get_parser(description, cmd_modules)
-   local argsp: Args = parser:parse(args) as Args --!
+   local cmdline_args: Args = parser:parse(args) as Args --!
```
hishamhm, line 624: @V1K1NGbg  I meant doing the opposite replacement: 

`args` to `cmdline_args`
`argsp` to `args`

hishamhm:
Yay, finally got CI fully green on this branch, having the Teal code passing on CI on Lua 5.4 and LuaJIT 2.1.rolling on Linux, macOS and Windows! (the latter using both MSVC and Mingw32 -- phew!)

It took a lot of work both here and on other repositories to make that happen, to bring the Github Actions compatible with LuaJIT 2.1.rolling and to get lua-compat-5.3 compatible with Windows.

Last thing pending now is to integrate the Teal compiler itself into CI to make sure that .tl files committed get type-checked and .lua files generated. I might still require manual commits of the .lua files just to keep the build files working the same for now, to avoid requiring major refactors of the Makefiles and CI workflows, which would grow this PR even further. I'll look at what the minimal set of infra changes is needed in order to merge this safely.

But we're almost there!!!

hishamhm:
@V1K1NGbg The branch is now merged!! Congratulations on your project! :tada:

V1K1NGbg:
Wow that is amazing! I am so happy with the end result! Thank you again for everything @hishamhm !

PR #1707, opened 2024-08-28
fix: when linking Lua explicitly (Windows), use dynamic libraries

hishamhm:
The original code was inconsistent, since it demanded dynamic libraries for Cygwin but static libraries for other Windows compilers. However, that wouldn't work for the same reason as Cygwin:

From the LuaJIT makefile:
https://github.com/LuaJIT/LuaJIT/blob/ae4735f621d89d84758769b76432d2319dda9827/src/Makefile#L558-L563V

"Mixed mode [static + dynamic] is not supported on Windows. And static mode doesn't work well.
C modules cannot be loaded, because they bind to lua51.dll."

### DIFF
--- a/src/luarocks/deps.lua
+++ b/src/luarocks/deps.lua
@@ -553,15 +553,13 @@ local function check_external_dependency(name, ext_files, vars, mode, cache)
 
    for _, prefix in ipairs(prefixes) do
       prefix = resolve_prefix(prefix, dirs)
-      if cfg.is_platform("mingw32") and name == "LUA" then
+
+      if name == "LUA" and cfg.link_lua_explicitly then
          dirs.LIBDIR.pattern = fun.filter(util.deep_copy(dirs.LIBDIR.pattern), function(s)
-            return not s:match("%.a$")
-         end)
-      elseif cfg.is_platform("windows") and name == "LUA" then
-         dirs.LIBDIR.pattern = fun.filter(util.deep_copy(dirs.LIBDIR.pattern), function(s)
-            return not s:match("%.dll$")
+            return not s:match("%." .. cfg.static_lib_extension .. "$")
          end)
       end
+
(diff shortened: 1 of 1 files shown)


hishamhm:
Apparently Windows _with MSVC_ does need static libraries when linking:

From https://lua-users.org/wiki/BuildingModules (note however that this page was last edited in 2014)

<blockquote>

## Windows DLLs with MSVC

* Use the IDE to create a DLL project (File, New, Projects).
* Select the multithreaded DLL runtime library (C/C++, Category = Code Generation).
* Add the Lua include directory to the include path (C/C++, Preprocessor).
* Add the Lua directory to the library path (Link, Input).
* Add the Lua import library lua51.lib to the input libraries (Link, Input).
* Build the DLL (Build). 

note : make sure your main init function is exported with __declspec(dllexport) like this :

int __declspec(dllexport) MyModuleName (lua_State* L) { ... }

## Windows DLLs with GCC (MinGW, Cygwin or MinGW cross-compiler)

gcc -O2 -c -o module.o module.c
gcc -O -shared -o module.dll module.o -Llua_dir -lluaXX

Replace lua_dir with the directory where the luaXX.dll resides. luaXX.dll is either lua50.dll or lua51.dll, depending on which Lua version you use.
</blockquote>

tobil4sk:
> Apparently Windows with MSVC does need static libraries when linking:

@hishamhm Hi, I've been running into issues with linking on msvc recently so I've been looking into this. `.lib` on windows can be an "import" `.lib` for a `.dll`. It is still linked dynamically, but it requires both `.lib` for the definitions and `.dll` for the actual binary.

Sometimes people use `.dll.lib` for these types of `.lib` files to avoid confusion with `.lib` static libraries.

hishamhm:
@tobil4sk I have no real understanding of how linking works on Windows, and I don't even have a Windows environment to test things locally other than running things very slowly on CI. Unfortunately this means I'm unable to debug Windows issues efficiently. Any PRs helping to improve the Windows situation are much welcome.

PR #1730, opened 2024-12-21
add support for LoongArch

yetist:
Please review this PR, thanks.


src/luarocks/core/sysdetect.tl
```diff
@@ -309,6 +311,11 @@ local function read_elf_header(fd: FILE): ElfHeader, Processor
    if endian == "little" and processor == "ppc64" then
       processor = "ppc64le"
    end
+   if processor == "loongarch" and hdr.bits == 1 then
```
alerque, line 314: Shouldn't these continue an elseif chain?

yetist: The code has been modified, currently only loongarch32 and loongarch64

PR #1738, opened 2025-02-03
only create wrapper for lua scripts

Freed-Wu:


src/luarocks/fs/unix.lua
```diff
@@ -145,13 +145,14 @@ function unix.is_actual_binary(filename)
...
-   local first = file:read(2)
+   local first = file:read()
    file:close()
    if not first then
       util.warning("could not read "..filename)
       return true
    end
-   return first ~= "#!"
+   -- only create wrapper for lua scripts
+   return first:match('#!.* lua') == first
```
alerque, line 155: This is not an effective robust match. The space before `lua` is optional (it might be part of a full path), the capitalization could differ, a space could be used after the hash and before the bang, this may not be the end of the line, etc.

hishamhm:
```
Lua 5.3.6  Copyright (C) 1994-2020 Lua.org, PUC-Rio
> first = "#!/usr/bin/env luajit"
> first:match('#!.*lua') == first
false
> 
```

Also, Windows scripts don't have `#!`. You want to exclude `#!`-files which don't have `lua` anywhere. That would be a better heuristic.

Freed-Wu:
If a program needs a specific lua version, such as luajit, we shouldn't create wrapper for it, because we cannot ensure luarocks's lua and this program's lua is a same lua version.

alerque:
Isn't the wrapper created with the version of Lua potentially explicitly specified by the user at install time?

hishamhm:
@alerque It is.

@Freed-Wu what upstream called lua or luajit in their original script is never reliable in terms of the end-user's system.

No matter what we do, things will break for some people and people will complain, because things are so not standardized across the Lua world.

In some systems the luajit interpreter is called luajit-2.xyaddayadda and there is no luajit binary. Then someone installs a package lua-resty-something made "for LuaJIT". Then we apply your heuristic and don't make a wrapper. Then the script doesn't work. "LuaRocks sucks!"

Or maybe there is a luajit binary, but the user explicitly configured luarocks to use a custom luajit-2.x interpreter. Then we apply your heuristic and don't make a wrapper. Now the script is using a different interpreter from what the user specified. "LuaRocks sucks!"

Or we don't apply your heuristic. And then the scenario you described happens, because it is also possible. And then other things break. "LuaRocks sucks!"

I tried to come up with the whole wrapper thing as a convenience to make it easier for things to work out of the box most of the time. But as long as it doesn't work 100.000000% of the time, people will call it broken.

The wrapper thing was a well-intentioned idea. Because remember, in a world where people keep running and testing multiple Lua versions because of the endless 5.1/5.2/5.3/5 4/luajit debacle, doing nothing about it and just relying on "#!/usr/bin/env lua" will be broken in some system. People have no idea how many times they were "saved" by the wrappers.

You can try to make the system even more complicated to mitigate all of the scenarios I described above, by trying to guess the upstream's intention detecting luajit explicitly, and then checking if LuaRocks is managing a LuaJIT tree or not, and then adjusting luajit in the hashbang to whatever the name of the luajit binary is, and then applying other heuristics to decide what to do if the upstream hashbang called it just lua, or lua5.2, or something else. I tried to do this kind of guesswork for several things (location of headers, etc.). When it works, it works, when it doesn't, it fails badly. And it will never work every time because there's always assumptions and incorrect or insufficient metadata.

The best one can do is to have a simple rule and document it. But I don't think it's a nice thing to say "the wrappers only work if upstream calls their hashbang 'lua' and nothing else" and throw all luajit users under the bus, knowing the reality of how the luajit binary name situation is just as messed up as the lua binary situation is.

Yay, another rant! Sorry. This isn't intended as a dig at you at all -- thank you so much for the initiative to open a PR and try to make LuaRocks better! This is intended as an illustration of all the pieces I've been having to juggle in my head as I made decisions about the specific LuaRocks behaviors.

In short, I think a scenario where a specific LuaRocks tree is managing scripts for one given non-luajit Lua version bit then *also* needs to install luajit-only scripts that happen to need and work the the exact binary name _as provided by upstream_ is unrealistic.

With the change I proposed, yes, it could be that a hashbang that says "#!/usr/bin/luajit" (the interpreter chosen by upstream) gets wrapped by a wrapper that says lua5.4 (the interpreter chosed by the user), but then again that will only happen if the _rockspec author_ specified that the rock is compatible across that range. At this point, that is on them to honor that compatibility.

src/luarocks/fs/unix.lua
```diff
@@ -145,13 +145,14 @@ function unix.is_actual_binary(filename)
...
-   local first = file:read(2)
+   local first = file:read()
    file:close()
    if not first then
       util.warning("could not read "..filename)
       return true
    end
-   return first ~= "#!"
+   -- only create wrapper for lua scripts with `#!/usr/bin/env lua` or `#!/usr/bin/lua`
+   return first:match('#!.*[ /]lua') == first
```
hishamhm, line 155:
```suggestion
   return first:match('#!.*[ /]lua.*') == first
```

hishamhm (changes requested):

Freed-Wu:
Any other problem?

hishamhm (approved):

hishamhm:
@Freed-Wu thank you!

PR #1740, opened 2025-02-06
feat: Also publish json `source_digest` field

curiositycasualty:
This PR causes the `publishrelease` script to generate, then publish a sha256 digest of the source code artifact for each release (alongside the existing json fields).

~~Drafting while I prepare the changes to add digests for all previous releases.~~

See also: https://github.com/luarocks/luarocks/pull/1741

### DIFF
--- a/publishrelease
+++ b/publishrelease
@@ -43,6 +43,10 @@ do
    }
 done
 
+source_digest="$(
+   sha256sum "luarocks-${v}.tar.gz" | cut -d' ' -f1
+)"
+
 #######################################
 # utility
 #######################################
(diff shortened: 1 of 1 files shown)


gszr (approved):

curiositycasualty:
@hishamhm And this as well

hishamhm:
Does this add any additional security? Since the hashes and the files are in the same repository, both could be tampered with at the same time. The real security here is the `.asc` signature file, which can be verified with my public key.

curiositycasualty:
> Does this add any additional security? Since the hashes and the files are in the same repository, both could be tampered with at the same time. The real security here is the `.asc` signature file, which can be verified with my public key.

I don't see this as security-related, it's more just to validate that a download arrived correctly and to allow one to download the same files from some other cache.

curiositycasualty:
This is part of some work I'm doing to enabled Renovate to execute updates to `.requirements`

Issue #1700, opened 2024-07-26
Versions 3.10 and above do not install properly on windows with certain lua installs


tobil4sk:
- **Platform:**
  - Windows. (mingw-msys)

- **LuaRocks version:**
  - 3.10.0 and above (any version that includes 1ed80a15d215f5b924008f61f4a30da6139ca1bf)

- **LuaRocks output from when the issue occurred:**
 
Relevant parts of the installation logs:

```
Interpreter found, now looking for link libraries...
    checking for D:\Dev\lua\lua_install\lua5.4.lib
    checking for D:\Dev\lua\lua_install\lua54.lib
    checking for D:\Dev\lua\lua_install\liblua54.a
    checking for D:\Dev\lua\lua_install\lua5.4.dll
    checking for D:\Dev\lua\lua_install\lua54.dll
    checking for D:\Dev\lua\lua_install\liblua.dll.a
    checking for D:\Dev\lua\lua_install\lib\lua5.4.lib
    checking for D:\Dev\lua\lua_install\lib\lua54.lib
    checking for D:\Dev\lua\lua_install\lib\liblua54.a
       Found liblua54.a
```
Then from `luarocks` we see:
```
      LUA_LIBDIR : D:\Dev\lua\lua_install\lib (not found)
```

```
Error: Failed finding Lua library. You may need to configure LUA_LIBDIR.
```

- **Description:**

I installed using hererocks:
```
hererocks lua_install -r@"v3.10.0" --lua=5.4 --verbose
```
Lua is installed like this:
```
bin\lua54.dll
lib\liblua54.a
```
Since `lib` is checked first and liblua54.a is found there, `lib` is what `LUA_LIBDIR` is set to. However, since this doesn't contain the `lua54.dll`, there is an error because the dll is not found in the configured `LUA_LIBDIR` path.

- **Possible solutions:**

Removing liblua*.a from here avoids the issue:
https://github.com/luarocks/luarocks/blob/f87568a7cba69e2c19411e7a2b73016c80a14a3b/install.bat#L342

Another solution would be to prioritise `bin` over `lib`.

Alternatively, perhaps hererocks should stop building liblua$LUA_SHORTV.a at all, because the mingw instructions say to install nothing to `lib`.

```
 To install Lua properly, open the top-level Makefile and change:
...
[...]

PR #1744, opened 2025-02-19
Fix LUA_LIBDIR for msvc installations

tobil4sk:
The order here was swapped in #1701 to fix mingw builds (see #1700), but this broke MSVC installations which need to link against the `.lib`, which defines the dll imports.

### DIFF
--- a/install.bat
+++ b/install.bat
@@ -343,8 +343,10 @@ local function look_for_link_libraries(directory)
 	local directories
 	if vars.LUA_LIBDIR then
 		directories = {vars.LUA_LIBDIR}
-	else
+	else if vars.SYSTEM == "mingw" then
 		directories = {directory, directory .. "\\bin", directory .. "\\lib"}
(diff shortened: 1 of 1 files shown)


install.bat
```diff
@@ -343,8 +343,10 @@ local function look_for_link_libraries(directory)
 	local directories
 	if vars.LUA_LIBDIR then
 		directories = {vars.LUA_LIBDIR}
-	else
+	elseif vars.SYSTEM == "mingw" then
```
tobil4sk, line 346: It looks like this variable isn't defined, will have to try something else

tobil4sk: `USE_MINGW` is available here instead

hishamhm:
@tobil4sk Thank you!! I'm going through the backlog from back to front as you can imagine :)

tobil4sk:
Many thanks!

PR #1745, opened 2025-02-19
Fix make check for msys2 mingw system

tobil4sk:
On an msys2 install of mingw, make is installed as `mingw32-make`.

If `cmake` is installed, that takes priority over `mingw32-make` when matching the pattern `*make*.exe`.

This patch changes the pattern to be `*-make*.exe` instead to avoid this bug.


hishamhm:
@tobil4sk Thank you!!

PR #1746, opened 2025-02-20
I tried to fix the issue #719, as per the test results...

Vrohs:


Vrohs:
sorry for the sloppy PR message here are the test results.
Screenshot From 2025-02-20 21-13-32
Screenshot From 2025-02-20 21-13-48

PR #1747, opened 2025-02-20
Fix lua version variable issue #719

Vrohs:
Fix: Add automatic Lua version detection to Makefile (#719)

- Implemented logic to detect Lua version dynamically using the LUA interpreter.
- Fallback added to system-wide `lua` binary if `LUA` is undefined.
- Default Lua version set to 5.1 if detection fails, with a warning message.
- Ensured compatibility with existing `config.unix` structure.

This resolves issue #719 by exposing the `LUA_VERSION` variable for Makefile use, improving flexibility for environments with multiple Lua versions.
Screenshot From 2025-02-20 21-13-32
Screenshot From 2025-02-20 21-13-48

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -2,6 +2,29 @@ MAKEFLAGS += --jobs=1
 
 -include config.unix
 
+#Lua version detection
+
+ifeq ($(origin LUA_VERSION),undefined)
+  # Get Lua version directly from the LUA interpreter 
+  ifneq ($(origin LUA),undefined)
+    LUA_VERSION := $(shell $(LUA) -e "print((_VERSION or ''):match('Lua (%d+%.%d+)') or '')")
+  else
+    
+	# system lua if LUA isn't defined
+    LUA_VERSION := $(shell lua -e "print((_VERSION or ''):match('Lua (%d+%.%d+)') or '')")
+  endif
+  
+  # warning if we couldn't detect the version
+  ifeq ($(LUA_VERSION),)
+    $(warning Could not automatically detect Lua version. Please set LUA_VERSION manually.)
+    
+	LUA_VERSION := 5.1
+  endif
+endif
+
+
+
+
 datarootdir = $(prefix)/share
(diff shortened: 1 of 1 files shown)


GNUmakefile
```diff
@@ -2,6 +2,29 @@ MAKEFLAGS += --jobs=1
...
 
+#Lua version detection
+
+ifeq ($(origin LUA_VERSION),undefined)
+  # Get Lua version directly from the LUA interpreter 
+  ifneq ($(origin LUA),undefined)
+    LUA_VERSION := $(shell $(LUA) -e "print((_VERSION or ''):match('Lua (%d+%.%d+)') or '')")
+  else
+    
+	# system lua if LUA isn't defined
```
alerque, line 13: This should probably be consolidated by just setting `LUA ?= lua`. That way you don't have to check if the value is set and have two different control structures, you can just do one pass.

GNUmakefile
```diff
@@ -3,27 +3,7 @@ MAKEFLAGS += --jobs=1
...
-  ifeq ($(LUA_VERSION),)
-    $(warning Could not automatically detect Lua version. Please set LUA_VERSION manually.)
-    
-	LUA_VERSION := 5.1
-  endif
-endif
-
-
-
+LUA_VERSION ?= $(shell $(if $(LUA),$(LUA),lua) -e "print((_VERSION or ''):match('Lua (%d+%.%d+)') or '5.1')")
```
alerque, line 6: Better, but I think we can even make it a bit cleaner. I think this will accomplish the same thing, with the added advantage of keeping the LUA value around:

```suggestion
LUA ?= lua
LUA_VERSION ?= $(shell $(LUA) -e 'print((_VERSION or "5.1"):sub(-3))')
```

hishamhm:
#719 was not about adding LUA_VERSION to LuaRocks's own makefile. It was about making a `$(LUA_VERSION)` variable available to rockspecs that use the `make` build type.

Issue #1753, opened 2025-02-24
LuaRocks fails to install Lua modules written in C on MSYS2 environments which uses clang as default C compiler


luau-project:
### Please fill out the following:
- **Platform:**
  - Windows

- **LuaRocks version:**
  - Even though the version of LuaRocks installed on MSYS2 repositories is 3.9.2, I have confirmed locally that this issue still applies to 3.11.1

- **LuaRocks output from when the issue occurred:**
  It can be clearly seen on the image that LuaRocks invokes gcc rather than clang 

![Image](https://github.com/user-attachments/assets/1d572c74-99c7-48e6-bc9e-d48b7651583c)

- **Description:**

On MSYS2 MinGW-w64 environments using ```clang``` as the C toolchain (```CLANG64``` and ```CLANGARM64```), the default configuration has a fallback to use ```gcc``` as standard C compiler when the CC environment variable is not set. Without proper patches to handle this situation, such configuration makes the building of any Lua module with parts written in C to fail, because LuaRocks tries to use ```gcc``` when it should use ```clang```.

There are two manners to reproduce this issue:

1. Installing things locally and try to install a Lua module written in C through LuaRocks yourself (I'm not going to cover it, because it involves a bit of knowledge about MSYS2).
2. Through a GitHub Action (recommended)

> [!TIP]
> 
> If you don't want to run yourself the steps in the expandable section in the end of this issue, just login on GitHub and browse how it ran on a custom branch that I created for the reproduction [https://github.com/luau-project/luarocks/actions/runs/13504448310/job/37730636032](https://github.com/luau-project/luarocks/actions/runs/13504448310/job/37730636032).

> [!NOTE]
> 
> I have also reported this issue downstream [https://github.com/msys2/MINGW-packages/issues/23493](https://github.com/msys2/MINGW-packages/issues/23493).

<details>

<summary>To reproduce yourself, expand here</summary>

Create a new branch on your fork, paste the following yml code on a file ```.github/workflows/msys2-cc-issue-repro.yml``` and push it

```yaml
name: MSYS2 Issue Reproduction

on: [push, pull_request]

jobs:
  msys2-issue-repro:
    runs-on: windows-latest
    
    defaults:
      run:
        shell: msys2 {0}
    
    steps:
      - uses: actions/checkout@v4
      
      - uses: msys2/setup-msys2@v2
        with:
          msystem: CLANG64
          update: true
          install: |
            git
            mingw-w64-clang-x86_64-cc
[...]

PR #1754, opened 2025-02-24
fix: override default C compiler to cc.exe on MSYS2

luau-project:
## Description

Fixes [https://github.com/luarocks/luarocks/issues/1753](https://github.com/luarocks/luarocks/issues/1753)

Changes: 

1. on each MSYS2 MinGW-w64 environment, there is a ```cc.exe``` that points to the default C compiler for the environment. 
2. the line

```lua
defaults.variables.PWD = "cd"
```

is meant to follow a MSYS2 patch [https://github.com/msys2/MINGW-packages/blob/5fd4997c2508f0949366513f0a2dd4347f9c4e77/mingw-w64-lua-luarocks/0001-luarocks_msys2_mingw_w64.patch#L48](https://github.com/msys2/MINGW-packages/blob/5fd4997c2508f0949366513f0a2dd4347f9c4e77/mingw-w64-lua-luarocks/0001-luarocks_msys2_mingw_w64.patch#L48) that fixes previous issues for them.

> [!NOTE]
> 
> I have opened a PR on MSYS2 [https://github.com/msys2/MINGW-packages/pull/23494](https://github.com/msys2/MINGW-packages/pull/23494) that fixes this issue. However, I guess that MSYS2 maintainers expect these changes to be first accepted  upstream.

### DIFF
--- a/src/luarocks/core/cfg.lua
+++ b/src/luarocks/core/cfg.lua
@@ -468,6 +468,12 @@ local function make_defaults(lua_version, target_cpu, platforms, home)
       end
    end
 
+   if platforms.msys2_mingw_w64 then
+      defaults.variables.PWD = "cd"
+      defaults.variables.CC = os.getenv("CC") or "cc"
+      defaults.variables.LD = os.getenv("CC") or "cc"
+   end
+
    if platforms.bsd then
       defaults.variables.MAKE = "gmake"
       defaults.gcc_rpath = false


src/luarocks/core/cfg.lua
```diff
@@ -468,6 +468,12 @@ local function make_defaults(lua_version, target_cpu, platforms, home)
       end
    end
 
+   if platforms.msys2_mingw_w64 then
+      defaults.variables.PWD = "cd"
+      defaults.variables.CC = os.getenv("CC") or "cc"
+      defaults.variables.LD = os.getenv("CC") or "cc"
```
alerque, line 474: CC or LD?

luau-project: yes. These variables are initially defined for MSYS2 at [https://github.com/luarocks/luarocks/blob/c8653e5c712a3f3c4c4b3724ceda3caaf2d34851/src/luarocks/core/cfg.lua#L451-L460](https://github.com/luarocks/luarocks/blob/c8653e5c712a3f3c4c4b3724ceda3caaf2d34851/src/luarocks/core/cfg.lua#L451-L460)

However, modifying it there would be a mistake, because it would also impact users from the legacy MSYS / MinGW from [sourceforge](https://sourceforge.net/projects/mingw/), which is now hosted at [osdn](https://osdn.net/projects/mingw/). The legacy MinGW does not ship a ```cc.exe``` for gcc. That's why I moved the changes to a separate section of the code specifically for MSYS2.

alerque: Okay, fair enough.

hishamhm:
Thanks!!

Issue #1726, opened 2024-12-06
Luarocks depends on compa53 or lua>=5.3 for .tar.gz archives


bbblitz:
Hi,

[src/luarocks/tools/tar.lua](https://github.com/luarocks/luarocks/blame/master/src/luarocks/tools/tar.lua) uses [math.tointeger()](https://www.lua.org/manual/5.3/manual.html#pdf-math.tointeger), a Lua 5.3 addition. This file attempts to include compat53 at the top, but if that fails it silently ignores, and when luarocks attempts to untar a file it fails with the non-obvious error `[string "src/luarocks/tools/tar.lua"]:54: attempt to call field 'tointeger' (a nil value)`. I think the `tointeger()` calls can be replaced with with `tonumber(x,10)` and `math.floor()`.

In my case, here's the stacktrace, but I imagine it can happen in more places:
```
stack traceback:
        [string "src/luarocks/tools/tar.lua"]:91: in function 'read_header_block'
        [string "src/luarocks/tools/tar.lua"]:140: in function 'untar'
        [string "src/luarocks/fs/lua.lua"]:1282: in function 'unpack_archive'
        [string "src/luarocks/build.lua"]:185: in function 'fetch_and_change_to_source_dir'
        [string "src/luarocks/build.lua"]:397: in function 'build_rockspec'
        [string "src/luarocks/cmd/build.lua"]:82: in function 'build_rock'
        [string "src/luarocks/cmd/build.lua"]:125: in function 'do_build'
        [string "src/luarocks/cmd/build.lua"]:171: in function 'installer'
        [string "src/luarocks/deps.lua"]:237: in function 'fulfill_dependency'
        [string "src/luarocks/deps.lua"]:332: in function 'process_dependencies'
        [string "src/luarocks/build.lua"]:404: in function 'build_rockspec'
        [string "src/luarocks/cmd/build.lua"]:82: in function 'build_rock'
        [string "src/luarocks/cmd/build.lua"]:125: in function 'do_build'
        [string "src/luarocks/cmd/build.lua"]:171: in function <[string "src/luarocks/cmd/build.lua"]:138>
        [C]: in function 'xpcall'
        [string "src/luarocks/cmd.lua"]:794: in function 'run_command'
        [string "luarocks"]:34: in main chunk
        [C]: at 0x7ff6fa8b1610
```

gotten when installing `lua_cliargs` v3.0-2, which comes as a `.tar.gz` file.

Platform: Windows
[...]

PR #1757, opened 2025-02-24
Add vendored lua compat53

tobil4sk:
This fixes luarocks so that it can continue to run properly on lua 5.1/5.2 and luajit. See #1726

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -119,6 +119,16 @@ install: all install-config
 	do \
 	   $(INSTALL_DATA) "$$f" '$(DESTDIR)$(luadir)'/`echo $$f | sed 's,^src/,,'`; \
 	done
+ifeq (,$(findstring $(LUA_VERSION),"5.3" "5.4"))
+	find src/compat53/ -type d | while read f; \
+	do \
+	   mkdir -p '$(DESTDIR)$(luadir)'/`echo $$f | sed 's,^src/,,'`; \
+	done
+	find src/compat53/ -type f -name '*.lua' | while read f; \
+	do \
+	   $(INSTALL_DATA) "$$f" '$(DESTDIR)$(luadir)'/`echo $$f | sed 's,^src/,,'`; \
+	done
+endif
 
 install-config:
 	mkdir -p '$(DESTDIR)$(luarocksconfdir)/'
--- a/install.bat
+++ b/install.bat
@@ -1024,6 +1024,17 @@ end
 if not exec(S[[XCOPY /S src\luarocks\*.* "$LUADIR\luarocks" >NUL]]) then
 	die()
 end
+if vars.LUA_VERSION ~= "5.3" or vars.LUA_VERSION ~= "5.4" then
+	-- Copy the vendored lua-copmat53 source files
+	if not exists(S[[$LUADIR\compat53]]) then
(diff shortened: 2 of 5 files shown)


hishamhm:
Merged! Thank you for the initiative!

mcepl:
> Hello, thanks for the feedback.
> Would it help to add a configure option to skip installation of the vendored luacompat files? Note, luacompat would only be a dependency for lua 5.1 or 5.2.

Yes, that perfectly possible, even if the use of non-vendored compat53 would be opt-in (there are many packages with options like `--use-system-foo`).

> > Third, even if you decide to vendor [...] then put these files somewhere to private directory outside of LUA_PATH.
> 
> That would make sense. I think the issue that prevented it in the first place is that the teal compiler places in the automatic luacompat imports, which meant it wouldn't work if placed in the src/luarocks/vendor directory. Perhaps this can be solved by moving the files and adding the private import path manually into the luarocks entry point, something like:

Whatever I think about vendoring-in dependencies (and yes, `luarocks` might be an exception because of bootstrapping for normal users), it is not a bug to do so. However, taking a namespace of other package (`%{lua_noarchdir}/compat53` in RPM SPEC file syntax) is a bug and a pretty nasty one.

> @alerque Perhaps the arch luarocks package is built against lua >= 5.3, in which case the luarocks makefile doesn't install the vendored files. In this case, it would be important to test whether an installation on arch with luarocks and only lua 5.1 would even work.

Obviously, I don’t test on ArchLinux, but with compat53 version 0.14.4 (REQUIRED, lower versions, tested with 0.13.* failed) I could build plenty of packages from openSUSE [`devel:languages:lua`](https://build.opensuse.org/project/monitor/devel:languages:lua?defaults=0&succeeded=1&failed=1&unresolvable=1&broken=1&blocked=1&dispatching=1&scheduled=1&building=1&finished=1&signing=1&locked=1&deleting=1&arch_aarch64=1&arch_armv7l=1&arch_i586=1&arch_ppc64le=1&arch_s390x=1&arch_x86_64=1&repo_openSUSE_Factory_ARM=1&repo_openSUSE_Factory_PowerPC=1&repo_openSUSE_Factory_zSystems=1&repo_openSUSE_Tumbleweed=1).

alerque:
> Would it help to add a configure option to skip installation of the vendored luacompat files? Note, luacompat would only be a dependency for lua 5.1 or 5.2.

Yes

> That would make sense. I think the issue that prevented it in the first place is that the teal compiler places in the automatic luacompat imports, which meant it wouldn't work if placed in the src/luarocks/vendor directory. Perhaps this can be solved by moving the files and adding the private import path manually into the luarocks entry point, something like:
> 
> https://github.com/luarocks/luarocks/blob/99c57c8b2464550d6659cce43f84db83b17c4c15/GNUmakefile#L74

That makes way more sense that hijacking the system default package.path for what should/could be a separate package.

> @alerque Perhaps the arch luarocks package is built against lua >= 5.3, in which case the luarocks makefile doesn't install the vendored files. In this case, it would be important to test whether an installation on arch with luarocks and only lua 5.1 would even work.

This is correct, Arch's default LuaRocks package is built against 5.4, so even if you use LuaJIT (as I do) for projects LuaRocks itself is not using as it's internal interpreter and you have to have PUC Lua installed in addition to the LuaJIT runtime.

PR #1761, opened 2025-03-06
WIP: LuaRocks 4

hishamhm:
Draft PR to increase awareness of this branch.

* [x] vendor in all basic dependencies for a stand-alone binary build of the luarocks CLI
* [x] add "development mode" so that the binary can hot-load .tl files from the repo directly so that .tl files can be worked on without recompiling the binary over and over
* [x] move luarocks-loader into its own repository: https://github.com/luarocks/luarocks-loader
* [ ] remove luarocks-loader from this repo
* [ ] move luarocks-admin into its own repository
* [ ] remove luarocks-admin from this repo
* [ ] support cross-compilation of luarocks.exe binary for Windows using the main Makefile
* [ ] remove other methods of installation for Windows (install.bat, win32 tools, etc.)
* [ ] remove all other "tool based" modes of luarocks.fs, keep only the mode based on the vendored-in tools
* [ ] vendor in and statically link the Lua interpreter, so that the luarocks binary always runs on the same well-known version of Lua
* [ ] use a special tree for loading luarocks plugins (e.g luarocks.build.foo modules) that is independent from the Lua version being managed by LuaRocks (i.e. so that luarocks can manage a LuaJIT "Lua 5.1" tree but still load Lua 5.4 modules for its own extensibility).
* [ ] adjust configure script for configuring the non-Lua dependencies (libbz2, zlib, OpenSSL) - allow static linking (for shipped binaries) and dynamic linking (for distros)

(This is a WIP branch - force-pushes are expected)


mcepl:
> vendor in and statically link the Lua interpreter, so that the luarocks binary always runs on the same well-known version of Lua

Make it optional, please. Linux distributions (I package Lua for SUSE) would not be happy otherwise, because we would have to patch it out somehow.

alerque:
Yes please, very much second @mcepl's request. We'll have the same issue on Arch Linux. Even if a specific version series (e.g. 5.5) or something is required we'd still want to patch out the vendored version and use the system package to provide it.

Issue #1774, opened 2025-03-20
documentation for instaling LuaRoks for multiple Lua versions is misssing


FractalU:
There is no documentation for installing LuaRocks for multiple Lua versions. There are some systems which install LuaRocks for all Lua versions. For example [Haiku](https://github.com/haikuports/haikuports/blob/master/dev-lua/luarocks/luarocks-3.11.1.recipe) and [Arch Linux](https://gitlab.archlinux.org/archlinux/packaging/packages/luarocks/-/blob/main/PKGBUILD?ref_type=heads). These systems found out their own tricks to install LuaRocks for all Lua versions.

Issue #444, opened 2015-10-29
Make handling multiple Lua versions easier


raingloom:
I use LuaJiT and 5.3 and aim to be as compatible with anything in between as possible, because of this, I often need multiple versions of modules.

Currently I need three different luarocks installations to suit my workflow, but this is really sub-optimal and makes testing a nightmare.

One solution I found for pure Lua modules is simply symlinking, but this is an ugly workaround and the rocktrees probably won't recognize these versions.

xpol:
I currently use luaenv and setup luarocks for each version.

raingloom:
@xpol  what is your exact workflow? Do you install luarocks from source? Are there any extra configuration changes? How do you handle shell commands, eg.: the ones created by busted or ldoc?

xpol:
Yes, I manually install luarocks form source. and lua and luajit have different header path.

luaenv will hash your luarocks installed shell command such as busted and ldoc.

[Here is luaenv and how it works](https://github.com/cehoffman/luaenv) you may needs [lua-build](https://github.com/cehoffman/lua-build) to make install lua versions easy.

After that install luarocks into `~/.luaenv/versions/\<luaversion\>/` (do NOT install to `~/.luaenv/shims/`)

raingloom:
I already had the Lua install part done (from Arch repos, as they are dependencies for certain packages I use). I guess even if I install the luarocks package I could still override it with luaenv.

So this does, in fact, solve the conflicts, but not the redundancy problem.
I'll see if I can either do some package.path hackery to solve that, which may very well go ugly.

PR #1775, opened 2025-03-20
add documentation for installing LuaRocks for multiple Lua versions

FractalU:
fixes https://github.com/luarocks/luarocks/issues/1774, https://github.com/luarocks/luarocks/issues/444

### DIFF
--- a/docs/installation_instructions_for_unix.md
+++ b/docs/installation_instructions_for_unix.md
@@ -131,6 +131,23 @@ After installation, a default config file called config.lua will be installed
 at the directory defined by `--sysconfdir`. For further configuration of
 LuaRocks paths, see the [Config file format](config_file_format.md).
 
+## Configure LuaRocks for multiple Lua versions
+
+Luarocks supports multiple Lua versions via the `--lua-version <ver>` flag. For it to work, LuaRocks needs a config file for each Lua version, which are named as `config-<ver>.lua`. As an example, for Lua 5.1 the config file is named `config-5.1.lua`. Unfortunately, running `./configure` and then `make` it generates only one config file from the detected Lua version. The additional ones need to be creaded manually in the `build` directory using `make`. In addition, to install these additional config files `sudo make LUA_VERSION=<ver> install-config` needs to be run. Running `sudo make install` won't install them. Note that the latest Lua 5.4 isn't listed in the `for in` loop. This is because the latest Lua is automatically being detected by running `./configure`.
+
+To configure, build and install LuaRocks for multiple Lua versions, run the following commands.
+```
+-$ ./configure --with-lua-include=/usr/local/include
+-$ make
+-$ for v in 5.3 5.2 5.1; do
+-$ 	make LUA_VERSION=$v LUA=/usr/local/bin/lua$v LUA_INCDIR=/usr/local/include/lua$v ./build/config-$v.lua
+-$ done
+-$ sudo make install
+-$ for v in 5.3 5.2 5.1; do
+-$ 	sudo make LUA_VERSION=$v install-config
+-$ done
+```
+
 ## Next steps
 
 Once LuaRocks is installed, learn more about [using LuaRocks](using_luarocks.md).


docs/installation_instructions_for_unix.md
```diff
@@ -131,6 +131,23 @@ After installation, a default config file called config.lua will be installed
...
+-$ make
+-$ for v in 5.3 5.2 5.1; do
+-$ 	make LUA_VERSION=$v LUA=/usr/local/bin/lua$v LUA_INCDIR=/usr/local/include/lua$v ./build/config-$v.lua
+-$ done
+-$ sudo make install
+-$ for v in 5.3 5.2 5.1; do
+-$ 	sudo make LUA_VERSION=$v install-config
+-$ done
+```
+
```
hishamhm, line 150: > Luarocks supports multiple Lua versions via the `--lua-version <ver>` flag. For it to work, LuaRocks needs a config file for each Lua version, which are named as `config-<ver>.lua`. As an example, for Lua 5.1 the config file is named `config-5.1.lua`. 

This is true. But I think the recommended approach here should be to have users (or distribution maintainers) to create those additional config files manually. For example:

A user can create/update those files using `luarocks config`. For example, assuming Lua 5.2 is installed as `/usr/bin/lua52`, this works:

```
luarocks config --lua-version=5.2 variables.LUA /usr/bin/lua52
```

Likewise for LUA_INCDIR and LUA_LIBDIR. Running `luarocks --lua-version=5.2` will then print a summary of the configuration, reporting if those paths were recognized as valid.

hishamhm (changes requested):

docs/installation_instructions_for_unix.md
hishamhm: >  it exposes some internals from Makefile, which isn't meant to be used. Bui it's still most practical for distro maintainers.

I have spent many years both as an upstream maintainer and as a distro maintainer. Making life easier for distro maintainers at the expense of making it harder for upstream maintainers is not the way to go.

> In my opinion the ./configure needs to be able to configure for multiple Lua versions. 

This is not the direction I'm going. `./configure` is about the Lua version that LuaRocks itself uses to run. 

Having `./configure` try to set up multiple Lua versions is not optimal for distributions either. They have separate packages for Lua versions, so building a multi-Lua-version LuaRocks causes a mismatch to a distribution's dependency graph. 

I've added several options over the years to make LuaRocks super-configurable and adaptable at the expense of its own maintainability. I'm not adding even more. Instead, it's time to cut back and streamline it. From LuaRocks 4 onwards, the LuaRocks CLI will only build with the latest Lua version; support for other Lua versions (including LuaJIT) will continue, but only through configuration.

It is very easy for a distribution to write and ship a three-line custom LuaRocks configuration file and bundle with their Lua package, or make it into a post-install rule for either their Lua or LuaRocks package. They already do that kind of thing for several packages (e.g. pkg-config files, post-install rules for font cache refreshes, etc.). They don't need to and should not invade the upstream Makefile internals to do so.

FractalU: > It is very easy for a distribution to write and ship a three-line custom LuaRocks configuration file and bundle with their Lua package, or make it into a post-install rule for either their Lua or LuaRocks package. They already do that kind of thing for several packages (e.g. pkg-config files, post-install rules for font cache refreshes, etc.). They don't need to and should not invade the upstream Makefile internals to do so.

Honestly, I don't know how to do it. Also, there is no documentation for how to do it. I made this pull request to add the documentation. I documented similar how [Arch Linux](https://gitlab.archlinux.org/archlinux/packaging/packages/luarocks/-/blob/main/PKGBUILD?ref_type=heads) does it, which is using the internals of the Makefile. This is not how it should be done but I didn't know much better. You said that it's very easy for a distribution to write and ship a LuaRocks config file and bundle with their Lua package or make it into post-install. Maybe this might be easy to do. But there is no indication or documentation on how to do it. So each distribution needs to figure out on their own how to configure LuaRocks for multiple Lua versions with no hints or documentation. This leads to many different ways of doing and not each one is good. The [arch distribution](https://gitlab.archlinux.org/archlinux/packaging/packages/luarocks/-/blob/main/PKGBUILD?ref_type=heads) and [mysy2](https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-lua-luarocks/PKGBUILD) uses the internals of the Makefile. I don't know how [debian](https://packages.debian.org/bookworm/luarocks) does it.

If there is a little bit of documentation and guidance on how to configure LuaRocks for multiple Lua versions it would be so much better. There doesn't need to be any additional functionality or change in the implementation. Just some explanation on how to configure Luarocks for multiple Lua versions suffices.

daurnimator: > I documented similar how [Arch Linux](https://gitlab.archlinux.org/archlinux/packaging/packages/luarocks/-/blob/main/PKGBUILD?ref_type=heads) does it, which is using the internals of the Makefile. This is not how it should be done but I didn't know much better.

What do you mean "the internals of the Makefile"?
The Arch PKGBUILD uses `make` the exact way that `make` works.

There's no need to require or even desire a `./configure` to do everything for you via introspecting your system: if you know what you want, you can create/edit configuration files yourself!

FractalU: Arch Linux uses `make` to create configuration files for multiple Lua versions. Below is the `make` invokation to create these Lua config files from [Arch Linux](https://gitlab.archlinux.org/archlinux/packaging/packages/luarocks/-/blob/main/PKGBUILD?ref_type=heads).
```
  make LUA_VERSION=5.1 LUA=/usr/bin/lua5.1 LUA_INCDIR=/usr/include/lua5.1 ./build/config-5.1.lua
  make LUA_VERSION=5.2 LUA=/usr/bin/lua5.2 LUA_INCDIR=/usr/include/lua5.2 ./build/config-5.2.lua
  make LUA_VERSION=5.3 LUA=/usr/bin/lua5.3 LUA_INCDIR=/usr/include/lua5.3 ./build/config-5.3.lua
```
The variables `LUA_VERSION` and `LUA_INCDIR` are internal to `make` and not public. They are not supposed to be used. In fact, it's not supposed to be using `make` to create Lua config files for multiple Lua versions. It's instead supposed to be using `luarocks config` for that. The problem is that there is no documentation on how to configure LuaRocks for multiple Lua versions. The Arch Linux uses `make` to create them, which isn't supposed to be done that way.

PR #1776, opened 2025-03-23
ci(windows): change zlib download url

luau-project:
## Description

On Windows, to build zlib as a test dependency, at the moment, the tarball is being downloaded from [https://zlib.net/zlib-1.3.1.tar.gz](https://zlib.net/zlib-1.3.1.tar.gz). However, the main page only holds the tarball for the latest released version (1.3.1).

If you try to download the previous released zlib version (1.3) from the main page, the requested tarball is not found:

image

This means that whenever a new zlib version pops up and the cached tarball expires, the step on CI to download zlib 1.3.1 will fail, because the tarball will not be there anymore.

According to zlib's website, all released versions of zlib are stored at [https://zlib.net/fossils/](https://zlib.net/fossils/).

## Changes

Replaced [https://zlib.net/](https://zlib.net/) by [https://zlib.net/fossils/](https://zlib.net/fossils/) to avoid a failed download when a new zlib is released and the cached version is expired.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -332,7 +332,7 @@ jobs:
         if: ${{ matrix.COMPILER == 'vs' && steps.restore-zlib-tarball.outputs.cache-hit != 'true' }}
         run: |
           curl -o "${{ github.workspace }}\zlib-${{ env.LUAROCKS_DEPS_ZLIB_VER }}.tar.gz" ^
-              "https://zlib.net/zlib-${{ env.LUAROCKS_DEPS_ZLIB_VER }}.tar.gz"
+              "https://zlib.net/fossils/zlib-${{ env.LUAROCKS_DEPS_ZLIB_VER }}.tar.gz"
 
       - name: Save zlib tarball
         if: ${{ matrix.COMPILER == 'vs' && steps.restore-zlib-tarball.outputs.cache-hit != 'true' }}


hishamhm:
Thanks! Perhaps we should bump the zlib dependency instead (or in addition)

luau-project:
yes, it is a possibility too

hishamhm:
Ah, I see you've done this preventively since 1.3.1 is still the current version. Thanks!

luau-project:
> Ah, I see you've done this preventively since 1.3.1 is still the current version. Thanks!

yes

hishamhm:
The old behavior had the "benefit" of reminding us to update zlib by breaking CI whenever a new release happened. I wonder if we should look into doing some dependabot-style periodic check.

luau-project:
> The old behavior had the "benefit" of reminding us to update zlib by breaking CI whenever a new release happened. I wonder if we should look into doing some dependabot-style periodic check.

I've seen some dependabot implementations on other projects, but I never messed around with it myself.

hishamhm:
Me neither. Looks like the official Github dependabot is [designed to work with a number of package environments](https://docs.github.com/en/code-security/dependabot/ecosystems-supported-by-dependabot/supported-ecosystems-and-repositories), but for something custom like an upstream C dependency, we'd probably have to create our own bespoke Github Action running on a `schedule` basis (or maybe there's some third-party action out there that already does something similar to what we need, but I doubt it would work out-of-the-box).

luau-project:
One project that has a very complete release pipeline that I am aware of is [zstd](https://github.com/facebook/zstd), and they also use dependabot. IIRC, they use it to update GitHub Actions pinned by commit.

hishamhm:
Apparently they use it [to update their github-actions](https://github.com/facebook/zstd/blob/dev/.github/dependabot.yml)? (Which, granted, could be useful!)

luau-project:
For example, when I submitted a code there (between 1.5.6 and 1.5.7)  using

```yaml
- uses: ilammy/msvc-dev-cmd@v1
```

they changed it to

```yaml
- uses: ilammy/msvc-dev-cmd@0b201ec74fa43914dc39ae48a89fd1d8cb592756 # v1.13.0
```

for security reasons, but I guess that their dependabot tell them when a new release pops up.

PR #1785, opened 2025-04-23
Add build variables with new rockspec format

renatomaia:
This PR contains two feature commits that introduce a new rockspec format and new build variables for this new format. I can remove the second commit if the feature it adds is not desired.

The first one fixes issue [#719](https://github.com/luarocks/luarocks/issues/719) by exposing a `LUA_VERSION` variable for use by rockspec build commands, including `builtin`, `make`, and others. It follows the same approach by PR [#789](https://github.com/luarocks/luarocks/pull/789), which introduces a new rockspec format as suggested [here](https://github.com/luarocks/luarocks/pull/789#issuecomment-382099519).

The second one adds variables that expose the installed directory of the dependency rocks, so the build can access any of the dependencies' installed rocks like directories `conf` or other non-standard directories defines by `copy_directories`.

My original motivation for this support is due to library [LuaMemory](https://github.com/renatomaia/lua-memory) that exposes both a Lua module and a C API (as a C library), and the [Coutil](https://github.com/renatomaia/coutil) library that depends on this C API. I was recommended to keep the LuaMemory C API counterpart as an external dependency for both [LuaMemory rock](https://luarocks.org/manifests/maia/memory-2.0.0-1.rockspec) and [Coutil rock](https://luarocks.org/manifests/maia/coutil-2.0.0-1.rockspec), and users are required to get the LuaMemory's C library using ad-hoc approaches (make, cmake, distro packages, etc.) even though LuaRocks has all the capacity required to distribute such pure Lua libraries.

Fast-forward a couple of years, and now I found I can expose the LuaMemory's C API from its own Lua module to other modules loaded later by using `package.loadlib` (see [here](https://github.com/renatomaia/coutil/blob/ba55f2f06ebec3242a520b41db03e9ed6e3057fd/demo/console.lua#L34)). However, we still need access to its headers (and in Windows to the lib that links to the DLL) to build modules like Coutil, and ideally the headers compatible with the rock installed. Also. I tried to achieve this with current LuaRocks in the following rockspec:

```lua
local luamem_release = "2.1.0-1"

package="coutil"
version="2.0.1-1"
...
dependencies = {
	"lua == 5.4",
	"vararg >= 2.1, < 3.0",
	"memory == "..luamem_release,
}
external_dependencies = {
	LIBUV = {
		header = "uv.h",
		library = "uv",
	},
}
build = {
	type = "cmake",
	variables = {
		CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS = "ON",
		CMAKE_INSTALL_PREFIX = "$(PREFIX)",
		CMAKE_LIBRARY_PATH = "$(LUA_LIBDIR);$(ROCKS_TREE)/memory/"..luamem_release.."/library;$(LIBUV_LIBDIR)",
		LUA_INCLUDE_DIR = "$(LUA_INCDIR)",
		LUAMEM_INCLUDE_DIR = "$(ROCKS_TREE)/memory/"..luamem_release.."/include",
		LIBUV_INCLUDE_DIR = "$(LIBUV_INCDIR)",
		MODULE_DESTINATION = "$(LIBDIR)",
	},
	...
}
```

But this approach have a lot of problems like:

- The rockspec must force a specific dependency version and revision.
- I believe the `ROCKS_TREE` variable refers the location the rock being built will be installed and not necessarily where its dependencies are installed, so the path used might be inexistent or wrong.

With the proposed feature we could have:

```lua
rockspec_format = "3.1"
package="coutil"
version="2.0.1-1"
...
dependencies = {
	"lua == 5.4",
	"vararg >= 2.1, < 3.0",
	"memory >= 2.1, < 3.0",
}
external_dependencies = {
	LIBUV = {
		header = "uv.h",
		library = "uv",
	},
}
build = {
	type = "cmake",
	variables = {
		CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS = "ON",
		CMAKE_INSTALL_PREFIX = "$(PREFIX)",
		CMAKE_LIBRARY_PATH = "$(LUA_LIBDIR);$(MEMORY_ROCKDIR)/library;$(LIBUV_LIBDIR)",
		LUA_INCLUDE_DIR = "$(LUA_INCDIR)",
		LUAMEM_INCLUDE_DIR = "$(MEMORY_ROCKDIR)/include",
		LIBUV_INCLUDE_DIR = "$(LIBUV_INCDIR)",
		MODULE_DESTINATION = "$(LIBDIR)",
	},
	...
}
```

### DIFF
--- a/docs/creating_a_makefile_that_plays_nice_with_luarocks.md
+++ b/docs/creating_a_makefile_that_plays_nice_with_luarocks.md
@@ -21,6 +21,7 @@ For building:
 * `LUA_INCDIR` - where to find the Lua headers
 * `LUALIB` - the name of the Lua library. This is not available nor needed on all platforms.
 * `LUA` - the name of the Lua interpreter
+* `LUA_VERSION` - the version of Lua
 
 For installing:
 
--- a/docs/creating_a_makefile_that_plays_nice_with_luarocks.md
+++ b/docs/creating_a_makefile_that_plays_nice_with_luarocks.md
@@ -16,18 +16,18 @@ For building:
 
 * `CFLAGS` - flags for the C compiler
 * `LIBFLAG` - the flags needed for the linker to create shared libraries
-* `LUA_LIBDIR` - where to find the lua libraries
-* `LUA_BINDIR` - where to find the lua binary
-* `LUA_INCDIR` - where to find the lua headers
-* `LUALIB` - the name of the lua library. This is not available nor needed on all platforms.
-* `LUA` - the name of the lua interpreter
+* `LUA_LIBDIR` - where to find the Lua libraries
+* `LUA_BINDIR` - where to find the Lua binary
+* `LUA_INCDIR` - where to find the Lua headers
+* `LUALIB` - the name of the Lua library. This is not available nor needed on all platforms.
+* `LUA` - the name of the Lua interpreter
 
 For installing:
 
 * `PREFIX` - basic installation prefix for the module
 * `BINDIR` - where to put user callable programs or scripts
-* `LIBDIR` - where to put the shared libraries
-* `LUADIR` - where to put the lua files
+* `LIBDIR` - where to put the shared libraries implementing modules
+* `LUADIR` - where to put the Lua scripts implementing modules
 * `CONFDIR` - where to put your modules configuration
 
 Most of these variables point immediately where you'd expect them to, but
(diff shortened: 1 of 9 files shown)


renatomaia:
I didn't wrote at first the tests correctly to run in Windows with the paths with backslashes (`\`), so the tests failed when the GitActions workflow ran the first time. It should be fixed by now.

hishamhm:
Merged! Thank you!

PR #1786, opened 2025-04-23
Avoid overrwriting CMake variables from rockspec

renatomaia:
I noticed that `luarocks.build.cmake` always replace `CMAKE_LIBRARY_PATH` from the rockspec with the value from the environment variable of the same name. This way we cannot easily provide to CMake the paths provided by LuaRocks (ex `LUA_LIBDIR`, `<ext_lib>_LIBDIR`, etc.). Moreover, CMake seems to read these environment variables (see [here](https://cmake.org/cmake/help/latest/variable/CMAKE_LIBRARY_PATH.html)) anyway, so I cannot figure out why this override was even necessary in the first place. For this reason my suggestion with this PR is to simply remove this "Pass Env variables" block entirely.

Alternatively, if we actually need to pass these environment variables to CMake as CLI arguments for some reason, then at least we should do something like the following to avoid overwriting definitions from the `rockspec`:

```diff
diff --git a/src/luarocks/build/cmake.tl b/src/luarocks/build/cmake.tl
index 9a512d1e..ee793e46 100644
--- a/src/luarocks/build/cmake.tl
+++ b/src/luarocks/build/cmake.tl
@@ -28,9 +28,9 @@ function cmake.run(rockspec: Rockspec, no_install: boolean): boolean, string, st
    local variables = build.variables or {}
 
    -- Pass Env variables
-   variables.CMAKE_MODULE_PATH=os.getenv("CMAKE_MODULE_PATH")
-   variables.CMAKE_LIBRARY_PATH=os.getenv("CMAKE_LIBRARY_PATH")
-   variables.CMAKE_INCLUDE_PATH=os.getenv("CMAKE_INCLUDE_PATH")
+   variables.CMAKE_MODULE_PATH = variables.CMAKE_MODULE_PATH or os.getenv("CMAKE_MODULE_PATH")
+   variables.CMAKE_LIBRARY_PATH = variables.CMAKE_LIBRARY_PATH or os.getenv("CMAKE_LIBRARY_PATH")
+   variables.CMAKE_INCLUDE_PATH = variables.CMAKE_INCLUDE_PATH or os.getenv("CMAKE_INCLUDE_PATH")
 
    util.variable_substitutions(variables, rockspec.variables)
 
```

### DIFF
--- a/src/luarocks/build/cmake.lua
+++ b/src/luarocks/build/cmake.lua
@@ -27,11 +27,6 @@ function cmake.run(rockspec, no_install)
    local build = rockspec.build
    local variables = build.variables or {}
 
-
-   variables.CMAKE_MODULE_PATH = os.getenv("CMAKE_MODULE_PATH")
-   variables.CMAKE_LIBRARY_PATH = os.getenv("CMAKE_LIBRARY_PATH")
-   variables.CMAKE_INCLUDE_PATH = os.getenv("CMAKE_INCLUDE_PATH")
-
    util.variable_substitutions(variables, rockspec.variables)
 
    local ok, err_msg = fs.is_tool_available(rockspec.variables.CMAKE, "CMake")
--- a/src/luarocks/build/cmake.tl
+++ b/src/luarocks/build/cmake.tl
@@ -27,11 +27,6 @@ function cmake.run(rockspec: Rockspec, no_install: boolean): boolean, string, st
    local build = rockspec.build as cmake.CMakeBuild
    local variables = build.variables or {}
 
-   -- Pass Env variables
-   variables.CMAKE_MODULE_PATH=os.getenv("CMAKE_MODULE_PATH")
-   variables.CMAKE_LIBRARY_PATH=os.getenv("CMAKE_LIBRARY_PATH")
-   variables.CMAKE_INCLUDE_PATH=os.getenv("CMAKE_INCLUDE_PATH")
-
    util.variable_substitutions(variables, rockspec.variables)
 
    local ok, err_msg = fs.is_tool_available(rockspec.variables.CMAKE, "CMake")


hishamhm:
@renatomaia I think the original intention was to allow for the LuaRocks configuration file to override the CMake configuration.

Rockspecs shouldn't be defining hardcoded paths anyway, so they can be relocatable.

> I cannot figure out why this override was even necessary in the first place. 

I tried to track in Git history where those variables are from, and they're there in the initial commit that imported from CVS in 2009(!). So the history has been lost to time! :)

renatomaia:
> I think the original intention was to allow for the LuaRocks configuration file to override the CMake configuration.

I see. It sure makes sense to set this up in the LuaRocks configuration, but the code seems to fill the info from environment variables.

> Rockspecs shouldn't be defining hardcoded paths anyway, so they can be relocatable.

Good point, but LuaRocks support for CMake does not seem to set these paths by itself, so I thought it was up to the rockspec to set up accordingly,  like in the example [here](https://github.com/renatomaia/coutil/blob/ba55f2f06ebec3242a520b41db03e9ed6e3057fd/etc/coutil-scm-1.rockspec#L37). Without setting the `CMAKE_LIBRARY_PATH`, CMake is not able to find the Lua installation informed by LuaRocks (though it will find the one from the system if available).

Without the change from this PR, the user has to set this environement variable accordingly in the system, even if we try to set it up in the rockspec. I probably could circumvent this by receiving the correct library path in CMake from a different variable name and set the `CMAKE_LIBRARY_PATH` with that value inside CMake, but the LuaRocks overwriting this particular variable with `nil` in the default case seems pointless at first glance. Specially since "CMake seems to read these environment variables (see [here](https://cmake.org/cmake/help/latest/variable/CMAKE_LIBRARY_PATH.html)) anyway".

By the way, I'm not very knowledgeable with  CMake (nor LuaRocks code), so I might be missing something here.

> I tried to track in Git history where those variables are from, and they're there in the initial commit that imported from CVS in 2009(!). So the history has been lost to time! :)

Yeah, I did that as well when I was trying to figure what LuaRocks code was doing and why. ;-)

PR #1791, opened 2025-05-02
Fix markdown rendering of luarocks.md

EddieEldridge:
The formatting here was completely busted when previewing the Markdown, not sure if this is a remnant from the old docs?


hishamhm:
Yes, these docs have been converted several times over the years. There's probably various conversion issues still lurking. Thank you!

PR #1814, opened 2025-07-22
fix: deps.check_lua_libdir() should read liblua in binary mode

sewbacca:
Hi, I stumbled over this while I was doing some cross-compilation testing using luarocks on Windows. Although luarocks is seemingly fine reading a `.dll` in text mode on Windows, an ELF-formatted `.so` not so much, which leads to reporting a Lua library mismatch error.

This PR fixes the issue.


hishamhm:
@sewbacca thanks! @luau-project fixed the cause for the CI failures above, so they're not caused by your PR.

Issue #1839, opened 2025-10-25
Instructions for macOS install: No rule to make target `test'


TangentFoxy:
- **Platform:**
  - MacOS.

- **LuaRocks version:**
  - 3.12.2

- **Configuration file:**
  - N/A

- **LuaRocks output from when the issue occurred:**
  - N/A

- **Description:**
  1. Following the manual install instructions for macOS, I am instructed to follow the Unix instructions. I ignored the dependency instructions at the top because I have no idea how to do the equivalents on macOS. I already had LuaJIT installed through Homebrew, so I expect that to work.
  2. Running `make all test` tells the user to run `./configure` first. I did this. That should probably be in the instructions.
  3. Running `make all test` again after running configure returns an error ending with "No rule to make target `test'"

Full error output:
```
tangent@kibble luarocks-3.12.2 % make all test
mkdir -p "build"
mkdir -p "build"
(printf '#!/opt/homebrew/bin/luajit\n'\
	'package.loaded["luarocks.core.hardcoded"] = { '\
	"$([ -n "" ] && printf 'FORCE_CONFIG = true, ')"\
	'SYSCONFDIR = [[/usr/local/etc/luarocks]] }\n'\
	'package.path=[[/usr/local/share/lua/5.1/?.lua;]] .. package.path\n'\
	'local list = package.searchers or package.loaders; table.insert(list, 1, function(name) if name:match("^luarocks%%.") then return loadfile([[/usr/local/share/lua/5.1/]] .. name:gsub([[%%.]], [[/]]) .. [[.lua]]) end end)\n'; \
	tail -n +2 src/bin/luarocks \
	)> "build/luarocks"
mkdir -p "build"
(printf '#!/opt/homebrew/bin/luajit\n'\
	'package.loaded["luarocks.core.hardcoded"] = { '\
	"$([ -n "" ] && printf 'FORCE_CONFIG = true, ')"\
	'SYSCONFDIR = [[/usr/local/etc/luarocks]] }\n'\
	'package.path=[[/usr/local/share/lua/5.1/?.lua;]] .. package.path\n'\
	'local list = package.searchers or package.loaders; table.insert(list, 1, function(name) if name:match("^luarocks%%.") then return loadfile([[/usr/local/share/lua/5.1/]] .. name:gsub([[%%.]], [[/]]) .. [[.lua]]) end end)\n'; \
	tail -n +2 src/bin/luarocks-admin \
	)> "build/luarocks-admin"
[...]

PR #1840, opened 2025-10-25
Installation instructions for Unix should specify to run ./configure before make

TangentFoxy:
`./configure` needs to be run before make commands.

(This addresses something I noticed while opening #1839 but does not fix it.)


TangentFoxy:
I misread the instructions.

Afaik, running `./configure` is still necessary there, but I have to double check that.

TangentFoxy:
... yeeaap I am dummy xP

It's been so long since I was regularly doing this stuff that I forgot it was a bit of an outlier there. Thank you for your patience. ^^

PR #1844, opened 2025-12-07
support Lua 5.5

catwell:
This PR adds initial support for Lua 5.5. With those changes I could build LuaRocks on lua-5.5.0-rc2.

PR #1850, opened 2025-12-27
GNUmakefile: test for LUA_VERSION as 5.5

luau-project:
I think we missed it on #1844


hishamhm:
@luau-project Thank you!

PR #1853, opened 2026-01-08
docs(rockspec-format): `source.dir` can be omitted if inferrable from archive content

mrcjkb:
The RockSpec format docs convey that `source.dir` can only be omitted if it can be inferred from `source.file`.
However, there is logic in place to infer it from the archive contents:

https://github.com/luarocks/luarocks/blob/4188fdb235aca66530d274c782374cf6afba09b8/src/luarocks/fetch.tl?plain=1#L537-L571

IIRC, this is for backward compatibility with older RockSpecs.
However, it seems that quite a few people still rely on this behaviour, so I'd suggest making it part of the official RockSpec format.


hishamhm:
@mrcjkb Sounds reasonable! Thank you!

PR #1854, opened 2026-01-15
Improve flexibility around vendored libraries

tobil4sk:
compat53 is vendored since #1757 as it is required to run luarocks with lua 5.1 or 5.2.

However, this introduced some issues as the GNUmakefile install rule places these in the same place where `luarocks install compat53` would install them. This means you get conflicts if you install the actual package:
```
Warning: /.../prefix/share/lua/5.1/compat53/init.lua is not tracked by this installation of LuaRocks. Moving it to /.../prefix/share/lua/5.1/compat53/init.lua~
Warning: /.../prefix/share/lua/5.1/compat53/module.lua is not tracked by this installation of LuaRocks. Moving it to /.../prefix/share/lua/5.1/compat53/module.lua~
Warning: /.../prefix/share/lua/5.1/compat53/file_mt.lua is not tracked by this installation of LuaRocks. Moving it to /.../prefix/share/lua/5.1/compat53/file_mt.lua~
```

It is also not ideal for linux package maintainers to include a vendored package, see: https://github.com/luarocks/luarocks/pull/1757#issuecomment-3409873412.

To solve these issues, this patchset makes the following changes:
- GNUmakefile now places the compat53 files under `luarocks/vendor/compat53` (which is added internally to the luarocks script's `package.path`). This way a user's installation of compat53 does not interfere at all with luarocks one.
- Added `--with-system-compat53` option to configure script for external packaging systems.
- Fixed install.bat's logic for deciding whether to vendor compat53, as the current script includes it for every version.

install.bat already places luarocks sources outside of LUAPATH, so that part can stay as is.

I've also inverted the version check to avoid the need for future patches like: #1850.


configure
```diff
@@ -301,6 +304,9 @@ do
    --disable-incdir-check)
       DISABLE_INCDIR_CHECK=yes
       ;;
+   --with-system-compat53)
```
alerque: There was some mention of LuaRocks 4 headed towards vendoring *everything* it needs. Both myself (Arch Linux) and an OpenSUSE packager objected to that and requested that all vendoring be optional.

Perhaps this flag should be something more generic that would be an umbrella for all 3rd party Rock dependencies, e.g. `--with-system-luarocks` (as I have used for years as a configure flag for SILE).

https://github.com/sile-typesetter/sile/blob/master/configure.ac#L80-L84

tobil4sk:
I've updated to add `--with-system-rocks`, so it can be installed with no vendored files.

I realised that `make bootstrap` was still installed the vendored `compat53` files, which will still clash with the real `compat53` package. I had to move them out of `src` to prevent them from getting installed. However, this meant that they were not available to the bin script installed by make bootstrap. To fix this, I added the libraries as dependencies to the rockspec. This seems to make sense to me given what I understand `make bootstrap` to be for, but let me know.

hishamhm:
@tobil4sk looks like the failure in the BinaryBuild test is genuine: it is failing to find the `dkjson` module, which was moved by this PR. Could you take a look? I think this is the last pending thing before this PR is mergeable. Thank you!

tobil4sk:
> @tobil4sk looks like the failure in the BinaryBuild test is genuine: it is failing to find the dkjson module, which was moved by this PR. Could you take a look? I think this is the last pending thing before this PR is mergeable. Thank you!

This should now be resolved, the test is passing!

hishamhm:
@tobil4sk Thank you for this!

Issue #1257, opened 2021-01-08
LUA_LIBDIR is unset when using make build backend


gnarz:
I have a fresh install of lua 5.4.2 and luarocks 3.5.0 on a fresh install of linux MiNT 20.04. No other lua or luarocks versions are or were installed. When attempting to build my rocks, which use the make backend, with luarocks, I noticed that the LUA_LIBDIR variable is empty when luarocks calls make. The relevant output from the verbose luarocks log looks like this:

fs.execute("make ", "LUA_INCDIR=/usr/local/include", "CFLAGS=-O2 -fPIC", "LUA_BINDIR=/usr/local/bin", "LUA_LIBDIR=", "LIBFLAG=-shared", "CC=gcc")
fs.Q("LUA_INCDIR=/usr/local/include")
fs.Q("CFLAGS=-O2 -fPIC")
fs.Q("LUA_BINDIR=/usr/local/bin")
fs.Q("LUA_LIBDIR=")
fs.Q("LIBFLAG=-shared")
fs.Q("CC=gcc")
fs.execute_string("make  'LUA_INCDIR=/usr/local/include' 'CFLAGS=-O2 -fPIC' 'LUA_BINDIR=/usr/local/bin' 'LUA_LIBDIR=' 'LIBFLAG=-shared' 'CC=gcc'")

You will notice that both LUA_INCDIR and LUA_BINDIR are set, just LUA_LIBDIR is empty. This is reproducible with the sample Makefile and rockspec from here: https://github.com/luarocks/luarocks/wiki/Creating-a-Makefile-that-plays-nice-with-LuaRocks

imolein:
I just noticed the same, while I tried to install the `pulseaudio` module using luarocks 3.9.1 and lua 5.4.4.

siffiejoe:
On Linux (and I think a few other Unixes) binary extension modules must not be linked with the Lua library. The `lua` executable reexports the Lua API to them. So the `LUA_LIBDIR` and `LUALIB` variables are not needed.

tobil4sk:
It would be helpful if this was mentioned in the wiki: https://github.com/luarocks/luarocks/wiki/Creating-a-Makefile-that-plays-nice-with-LuaRocks

PR #1859, opened 2026-02-15
[docs] Fix LUA_LIBDIR handling in example Makefile

tobil4sk:
See: #1257

On most platforms LUA_LIBDIR is empty, so you get:
```
Warning: unmatched variable LUA_LIBDIR
```
And the resulting empty -L flag in the example can cause [issues](https://github.com/FourierTransformer/lua-simdjson/issues/78) like:
```
ld: warning: search path '-lpthread' not found
```

It's also not sufficient to pass the `-L` flag, you also have to provide the actual library to link.

I've updated the example to only set `LUA_LIBDIR` for windows where it is used by default. However, as noted in it is possible to do:
```sh
luarocks config link_lua_explicitly true
```

It would be better to respect this setting rather than hard-coding windows only, but I'm not sure the best way to check that config in the rockspec, I wonder if anyone else has any suggestions?

### DIFF
--- a/docs/creating_a_makefile_that_plays_nice_with_luarocks.md
+++ b/docs/creating_a_makefile_that_plays_nice_with_luarocks.md
@@ -163,16 +163,17 @@ install:
 You probably don't just want to echo stuff, so here's how to use the variables
 when actually building or installing something:
 
-```
-...
+```make
+#...
+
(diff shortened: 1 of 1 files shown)


hishamhm:
@tobil4sk Merge conflict fixed and merged manually, thank you!

tobil4sk:
Thanks!

PR #1862, opened 2026-02-26
docs: update link to the config file format

luau-project:


Tieske (approved):
lgtm, did you by any chance check other links? doing a grep or similar? should we?

luau-project:
> lgtm, did you by any chance check other links? doing a grep or similar? should we?

I used both VS code's *find in files* feature and the search bar on GitHub for the term `wiki/`. Such search queries only returned these URLs in the PR, and another ones in the change log. So, I **guess** these URLs in this PR are the only ones pointing to the wiki remaining in the source tree.

PR #1867, opened 2026-03-16
fix: check if table entry is nil or not

brunotvs:
running 

```luarocks config local_by_default``` returns ```Error: Unknown entry local_by_default```

and running

```luarocks config local_by_default true``` writes it as a string to the config file: 
```lua 
local_by_default = "true"
```

Same is true to any bool var set as false.

This pr aims to fix this issue.

### DIFF
--- a/spec/quick/config.q
+++ b/spec/quick/config.q
@@ -95,3 +95,41 @@ Warning: Failed finding Lua header lua.h (searched at /some/bad/path). You may n
 
 LuaRocks may not work correctly when building C modules using this configuration.
 --------------------------------------------------------------------------------
+
+
+
+================================================================================
+TEST: reports when getting a falsy boolean variable
+
+RUN: luarocks config local_by_default
+
+STDOUT:
+--------------------------------------------------------------------------------
+false
+--------------------------------------------------------------------------------
+
+
+
+================================================================================
+TEST: reports when setting a falsy boolean variable
+
+RUN: luarocks config local_by_default true
+
+STDOUT:
+--------------------------------------------------------------------------------
+Wrote
+local_by_default = true
+--------------------------------------------------------------------------------
+
+
+
+================================================================================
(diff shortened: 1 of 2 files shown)


hishamhm:
@brunotvs Thank you for the contribution! Could you verify the fix by adding some test(s) in `spec/quick/config.q`? They should be simple to write, following the existing examples.

(Please also rebase to the latest `main`, so that the latest fix to CI by @luau-project gets included.)

spec/quick/config.q
```diff
@@ -95,3 +95,41 @@ Warning: Failed finding Lua header lua.h (searched at /some/bad/path). You may n
...
+
+STDOUT:
+--------------------------------------------------------------------------------
+false
+--------------------------------------------------------------------------------
+
+
+
+================================================================================
+TEST: reports when setting a falsy boolean variable
```
hishamhm, line 114:
```suggestion
TEST: reports when setting a true boolean variable
```

hishamhm (changes requested):
Minor copy-paste typo, otherwise LGTM!

spec/quick/config.q
brunotvs: What I meant was the this is testing when the variable being set is falsy.

This test would work without this PR if the variable was set to true.

```lua
local_by_default = true
```

``` bash 
luarocks config local_by_default true
```
```
Wrote 
local_by_default = true -- boolean
```

But would fail if the variable was 'falsy'

```lua
local_by_default = false
```

``` bash 
luarocks config local_by_default true
```
```
Wrote 
local_by_default = "true" -- string
```

From now on, this value would always be truthy, from lua conditionals.

hishamhm: Ahh, I see! Thank you for the clarification!

hishamhm (approved):

PR #1868, opened 2026-03-18
feat: warn on module name collisions during rock installation

### LINKED ISSUE #1148: Warnings about name collision of modules
Could it be possible to have a warning message when installing a rock that contains a module that is already provided by another rock? Or is there already an option for this and I didn't find it?

At least with Luarocks version 3.2.1 I verified that there are no warnings for this case. 

However it seems that Luarocks is aware of the collision and installs the new module under another name and activates the, if the rock containing the first module is deinstalled.

sharadxsingh:
When installing a rock that provides a module already provided by another installed rock, print a warning indicating the collision and which module will take priority.

Closes #1148

### DIFF
--- a/src/luarocks/repos.lua
+++ b/src/luarocks/repos.lua
@@ -470,6 +470,13 @@ function repos.deploy_local_files(name, version, wrap_bin_scripts, deps_mode)
          local paths = get_deploy_paths(name, version, "bin", file_path, repo)
          local mode, cur_name, cur_version = check_spot_if_available(name, version, "bin", file_path)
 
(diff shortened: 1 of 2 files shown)


hishamhm:
@sharadxsingh Could you send the change to the actual `.tl` file from which the `.lua` file is generated? 

(There is a `teal.sh` script at the root of the repo that can rebuild the `.lua` files.)

Thank you!

sharadxsingh:
Updated! I've made the changes in the .tl source file (src/luarocks/repos.tl) and regenerated repos.lua using teal.sh. Please let me know if any changes are needed.

PR #1871, opened 2026-03-27
doc(cmd.write_rockspec): change `--lua-versions` description

### LINKED ISSUE #1870: Rename `--lua-versions` in `luarocks init`
I was going to write a bug report concerning how `--lua-versions` is not respected when running `luarocks init`, because it would only create the config and rocks tree for the Lua version specified by `--lua-version` or `default-lua-version.lua`. Then, I realized that it was still being used when writing the new rockspec, setting `lua >= 5.min, < 5.max+1` in `dependencies`. So that option must be for describing which versions of Lua your new rock will support.

Looking at the `--help` page, it's not very descriptive, with most of it describing accepted values.

```
   --lua-versions <ver>  Supported Lua versions. Accepted values are: "5.1",
                         "5.2", "5.3", "5.4", "5.5", "5.1,5.2", "5.2,5.3",
                         "5.3,5.4", "5.4,5.5", "5.1,5.2,5.3", "5.2,5.3,5.4",
                         "5.3,5.4,5.5", "5.1,5.2,5.3,5.4", "5.2,5.3,5.4,5.5",
                         "5.1,5.2,5.3,5.4,5.5".
```

"Supported Lua versions" doesn't describe if the dev environment supports those versions or the rock does. I thought it was the former.

It doesn't do the same thing as `--lua-version`, but the names are so similar that I thought it meant the same thing for the longest time. I don't want anyone else to have this association, so it would be nice if that option was at least aliased to something with `target` in the name, like `--target-lua-versions`, `--lua-targets`, or just `--targets`. I think that would help convey its purpose better. And maybe re-writing the description so it says something about how the rock will support those versions when published.

goldenstein64:
Closes #1870.

### DIFF
--- a/src/luarocks/cmd/write_rockspec.lua
+++ b/src/luarocks/cmd/write_rockspec.lua
@@ -54,7 +54,7 @@ function write_rockspec.cmd_options(parser)
    argname("<txt>")
    parser:option("--homepage", "Project homepage."):
    argname("<txt>")
-   parser:option("--lua-versions", 'Supported Lua versions. Accepted values are: "' ..
+   parser:option("--lua-versions", 'Lua versions supported by the rockspec. Accepted values are: "' ..
    table.concat(lua_versions, '", "') .. '".'):
    argname("<ver>"):
    choices(lua_versions)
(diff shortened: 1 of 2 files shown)


goldenstein64:
Maybe I should test this just to make sure...

goldenstein64:
Running the raw Lua files as `src/bin/luarocks write_rockspec --help` works, but `luarocks test` gave me an error?

https://gist.github.com/goldenstein64/c6b1fb6445719e46c6ae299454891b8d

Running this on Windows with MinGW, but not on MSYS2. I'm guessing that's not a supported platform given the GH workflow. I guess I shouldn't be worried and see what actually running the workflow does.

PR #1872, opened 2026-03-30
adding `luarocks nix` command

teto:
Don't write the derivation name
This is to deal with cases like cqueues that have one rockspec per lua version.
Also did some cleanup.
adding support for maintainers/longDescription
Uses special nixpkgs urls mirror://luarocks when possible
Should make nixpkgs capable of using mirrors and thus be more resilient.
Properly support busted test type
remove the newly generated "path" from the attribute passed to fetchgit.


teto:
sry wrong repo

PR #1873, opened 2026-03-30
Nix/v3.12

teto:
nerdsniped by https://github.com/NixOS/nixpkgs/pull/504707, I hope I can get up to 3.13.


teto:
I had errors when status was 0 so I changed the check from true to 0 (running from this shell) but now I get `Call to nurl --indent 2 https://github.com/mpeterv/markdown 0.33 failed with status: true` (running from nixpkgs)

so status can be either "true" or "0". Does this depend on lua version ?

PR #1881, opened 2026-05-05
luarocks upload: implement two-factor authentication

hishamhm:
Same implementation as
https://github.com/leafo/moonrocks/commit/2e9585862497e65284bfb728542e5686ebec1280 by @leafo


src/luarocks/cmd/upload.lua
```diff
@@ -52,11 +54,44 @@ local function is_dev_version(version)
...
+   end
+end
+
 function upload.command(args)
    local api, err = Api.new(args)
    if not api then
       return nil, err
    end
+   api.code = args.code
+   api.on_tfa_required = prompt_tfa
```
mrcjkb, line 94: suggestion: Ideally, we should nudge users towards enabling 2FA and requiring it for uploads.

With this retry logic, users who don't have 2FA enabled might never be made aware that luarocks.org supports it.
So I'd suggest to add an explicit `--no-tfa` flag that disables 2FA (with a strong recommendation against using it). Then, luarocks could default to prompting the user for a code if no code is set, regardless of whether it's required for luarocks upload.

This would be a breaking change, but it would eliminate the need for `/upload` retry logic and it would potentially greatly improve overall ecosystem security.

(I've [made the same proposal for Lux](https://github.com/lumen-oss/lux/issues/1651#issuecomment-4873891309)).

Issue #1851, opened 2025-12-29
v3.13.0 rockspec is broken


catwell:
The problem comes from [this commit](https://github.com/luarocks/luarocks/commit/31032c434b01c853e94ef21b8d9979436fbf81bc) - see [duplicate tag fields inside version](https://github.com/luarocks/luarocks/commit/31032c434b01c853e94ef21b8d9979436fbf81bc#diff-f0da6b3f82da7faea0ccc221d01ad751f6ce7318a8918e866f3437c0215dbbe0R6-R7).

This must be a rebase issue during release because the problem is not present on the v3.13.0 branch. However it is present inside the release tarball.

catwell:
@hishamhm in case you haven't seen it.

catwell:
Note that the release archive on GitHub is fine, it uses a different commit. The one that is broken is [the official one](https://luarocks.github.io/luarocks/releases/luarocks-3.13.0.tar.gz).

javanthropus:
Is anyone going to fix this?  I still can't install lua under asdf because of this.  Should I look for another way to install?

sify21:
> Is anyone going to fix this? I still can't install lua under asdf because of this. Should I look for another way to install?

Also faced with this bug, you can change luarocks version for https://github.com/Stratus3D/asdf-lua
```
export ASDF_LUA_LUAROCKS_VERSION=3.12.2
```

javanthropus:
> Also faced with this bug, you can change luarocks version for https://github.com/Stratus3D/asdf-lua
> 
> ```
> export ASDF_LUA_LUAROCKS_VERSION=3.12.2
> ```

Thank you for that suggestion.  It worked for me.

cstrahan:
Trying to install LuaJIT via [mise](https://mise.jdx.dev/), and running up against this problem.

Tried: `mise install asdf:mise-plugins/mise-luaJIT`

This error is printed:

> Error: Could not load rockspec file /private/var/folders/sh/z0m6y3292yj5212httmvvs240000gp/T/luarocks-3.13.0/luarocks-3.13.0-1.rockspec (Error loading file: [string "/private/var/folders/sh/z0m6y3292yj5212httmvv..."]:7: '}' expected (to close '{' at line 4) near 'tag')

This is the workaround: `LUAROCKS_VERSION=3.12.2 mise install asdf:mise-plugins/mise-luaJIT`

attehuhtakangas:
[...]

PR #1885, opened 2026-06-04
makedist: fix rockspec tag update producing invalid Lua

attehuhtakangas:
## Summary

The official `luarocks-3.13.0.tar.gz` on luarocks.org ships a syntactically invalid `luarocks-3.13.0-1.rockspec` (#1851):

```lua
source = {
   url = "git+https://github.com/luarocks/luarocks",
   tag = "v3.13.0"
   tag = "v3.12.2",
}
```

This breaks `make bootstrap` and every tool that installs LuaRocks from the official tarball (asdf, mise, etc.), since the rockspec fails `loadfile` on every Lua version with `'}' expected (to close '{' at line 4) near 'tag'`.

## Root cause

In `makedist`, the sed that sets the source tag *inserts* a new line after the `url` line — without a trailing comma, and without removing a pre-existing `tag` line:

```bash
sed -i 's/\(   url = "[^"]*",\)/\1\n   tag = "v'$version'"/' $ROCKSPEC
```

When a release is cut from a tree that carries the *previous release's* rockspec (which already has a `tag` line) rather than `luarocks-dev-1.rockspec` (which doesn't), the result is the duplicate-tag syntax error above. That's exactly what happened for 3.13.0: commit 31032c434b01c853e94ef21b8d9979436fbf81bc (the first "Release 3.13.0" commit, cut from a lineage carrying `luarocks-3.12.2-1.rockspec`) contains the broken rockspec, and the official tarball was built from that state. The rockspec was hand-fixed two minutes later in 3421bedc2ce2b64e79530bb97497531b014899a8 (the commit tagged `v3.13.0`), but the tarball was never re-rolled — which is why the GitHub release archive is fine while the luarocks.org tarball is broken.

## Fix

- Delete any existing `tag = ...` line before inserting the new one, so the substitution is correct for both the dev rockspec and a previous release's rockspec.
- Insert the new tag line with a trailing comma, matching the hand-fixed 3421bedc.
- Add a guard alongside the existing `grep -q` checks that validates the rockspec actually loads as Lua (using the Lua 5.4 prefix `makedist` already requires) before the tarball is packed — this would have caught #1851 before it shipped.

## Verification

Ran the new sed sequence against both input shapes:

- `luarocks-dev-1.rockspec` (no tag line) → valid rockspec, `loadfile` OK
- `luarocks-3.12.2-1.rockspec` (stale tag line, the 3.13.0 scenario) → valid rockspec, `loadfile` OK, byte-identical `source` table to the hand-fixed 3421bedc
- Old sed against the 3.12.2 rockspec reproduces the shipped corruption and the exact error reported in #1851; the new `loadfile` guard fails on it with exit 1.

## Note for maintainers

This PR only fixes the release tooling so it can't happen again. The broken `luarocks-3.13.0.tar.gz` itself lives on `gh-pages` with a PGP signature and a `source_digest` pinned in `releases.json`, so only a maintainer can re-roll, re-sign, and re-publish it (or publish a 3.13.1). Until then, 3.13.0 — the first release to support Lua 5.5 — remains uninstallable from the official tarball.

Fixes the tooling side of #1851.

### DIFF
--- a/makedist
+++ b/makedist
@@ -59,7 +59,8 @@ then
    fi
    sed -i 's/"Configuring LuaRocks version .*"/"Configuring LuaRocks version '$version'..."/' configure
    sed -i 's/version = "[^"]*"/version = "'$version'-1"/' $ROCKSPEC
-   sed -i 's/\(   url = "[^"]*",\)/\1\n   tag = "v'$version'"/' $ROCKSPEC
+   sed -i '/^   tag = "[^"]*",\{0,1\}$/d' $ROCKSPEC
+   sed -i 's/\(   url = "[^"]*",\)/\1\n   tag = "v'$version'",/' $ROCKSPEC
    sed -i 's/program_version = "[^"]*"/program_version = "'$version'"/' src/luarocks/core/cfg.lua
    sed -i 's/vars.VERSION = "[0-9.]*"/vars.VERSION = "'$xyversion'"/' install.bat
    echo "==============================================================================="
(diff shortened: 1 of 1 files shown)


hishamhm:
@attehuhtakangas Thank you!!

cclauss:
Did this fix:
* #1851?

PR #1858, opened 2026-02-10
Bump lua-zlib from 1.2 to 1.4 as part of Lua 5.5 support

luau-project:
## Description

Updates `lua-zlib` (1.2 &rightarrow; 1.4). Note: `lua-zlib` brought Lua 5.5 support on 1.4

> [!IMPORTANT]
> 
> CI is expected to fail due the version change (1.2 &rightarrow; 1.4) applied in the test environment [https://github.com/luarocks/luarocks/blob/47301d83aba58925e1b9594023621ebb27070cdb/spec/util/test_env.lua#L1132](https://github.com/luarocks/luarocks/blob/47301d83aba58925e1b9594023621ebb27070cdb/spec/util/test_env.lua#L1132)

* Once a new `.src.rock` upload on LuaRocks org happens for the new version ([binary/lua-zlib-1.4-0.rockspec](binary/lua-zlib-1.4-0.rockspec)) under `hisham` manifest, CI tests are expected to return to the pass state;

* A bump to OpenSSL version (3.6.0 &rightarrow; 3.6.1) was also applied to CI.

## Details

If you compare the few main `lua-zlib` changes from v1.2 tag to v1.4

```bash
diff -Naur lua-zlib-1.2/lua_zlib.c lua-zlib-1.4/lua_zlib.c
```

it can be seen that they were mostly cosmetic changes:

```diff
--- lua-zlib-1.2/lua_zlib.c	2017-10-07 12:26:37.000000000 -0300
+++ lua-zlib-1.4/lua_zlib.c	2026-01-03 21:19:29.000000000 -0300
@@ -21,7 +21,7 @@
 #endif
 
 #ifdef LZLIB_COMPAT
-/**************** lzlib compatibilty **********************************/
+/**************** lzlib compatibility **********************************/
 /* Taken from https://raw.githubusercontent.com/LuaDist/lzlib/93b88e931ffa7cd0a52a972b6b26d37628f479f3/lzlib.c */
 
 /************************************************************************
@@ -51,7 +51,7 @@
 
 /*
 ** =========================================================================
-** compile time options wich determine available functionality
+** compile time options which determine available functionality
 ** =========================================================================
 */
 
@@ -613,16 +613,16 @@
 
 static int lzstream_readline(lua_State *L) {
     lz_stream *s;
-    int sucess;
+    int success;
 
     s = lzstream_check(L, lua_upvalueindex(1), LZ_INFLATE);
-    sucess = lz_read_line(L, s);
[...]

PR #1889, opened 2026-07-04
ci: Add lua v5.5 to the test matrix

cclauss:
```diff
      matrix:
-       lua-version: ["5.4", "luajit-2.1"]
+       lua-version: ["5.4", "5.5", "luajit-2.1"]
```
https://www.lua.org/versions.html


luau-project:
From the last time I tried to upgrade LuaRocks CI to use Lua 5.5 as default version (maybe 2 or 3 months ago), IIRC, I found a few blockers:

1. `lua-compat-5.3` needs a version supporting Lua 5.5 published on luarocks.org ([https://github.com/lunarmodules/lua-compat-5.3/issues/79](https://github.com/lunarmodules/lua-compat-5.3/issues/79))
2. Not sure about this one, but I think `lua-posix` also needs a version supporting Lua 5.5 published on luarocks.org ([https://github.com/luaposix/luaposix/issues/394](https://github.com/luaposix/luaposix/issues/394))
3. A rock upload for `lua-zlib` under `hishamhm` account on luarocks.org supporting Lua 5.5 ([https://github.com/luarocks/luarocks/pull/1858](https://github.com/luarocks/luarocks/pull/1858))

For sure, the item 2 is the hardest achievement, because `luaposix` has another set of dependencies, maintained by the same dev, which also needs to support Lua 5.5. One way to work around this limitation would be to provide a small C module binding to the POSIX APIs in use by the `luarocks` CLI tool. IIRC, `hishamhm` made some work ([https://github.com/luarocks/luarocks/tree/luarocks4/vendor/miniposix](https://github.com/luarocks/luarocks/tree/luarocks4/vendor/miniposix)) on `luarocks4` branch to dettach from `luaposix` dependency at [https://github.com/luarocks/luarocks/pull/1761](https://github.com/luarocks/luarocks/pull/1761).

In short, in my opinion, LuaRocks is ready for Lua 5.5, but its dependencies need to support Lua 5.5 for this transition to happen smoothly.

cclauss:
One step forward:
`lua-compat-5.3` [v0.15.1 release](https://luarocks.org/modules/lunarmodules/compat53) now works.

One step back:
* [x] We need to upgrade `luacov` and `cluacov` because their current versions depend on lua >= 5.1, < 5.5
    * [x] #1908
* [x] spec/util/quick.lua:277: attempt to assign to const variable 'line'
    * [x] #1897
* [x] #1858
* [x] spec/quick_spec.lua:16: spec/fixtures/a_repo/manifest-5.5: No such file
* [x] spec/quick_spec.lua:16: quick tests: #quick luarocks pack: packs a namespaced rock #namespaces
* [x] ./spec/util/test_env.lua:695: failed building test dependency: luafilesystem
* [ ] ./spec/util/test_env.lua:695: failed building test dependency: lua-bz2
    * [ ] #1910

luau-project:
>     * [x]  spec/util/quick.lua:277: attempt to assign to const variable 'line'

BTW, you just spotted a valid issue here that was missed during LuaRocks 3.13.0 release. Since I read somewhere that you said to be new to Lua development, I'm going to explain in detail what the issue is and how to fix it.

## Issue

In Lua 5.5 ( [https://www.lua.org/manual/5.5/manual.html#3.3.5](https://www.lua.org/manual/5.5/manual.html#3.3.5) ), the control variable in for-loops are read-only.

So, in the file `spec/util/quick.lua`, being within the body of a for-loop defined at

[https://github.com/luarocks/luarocks/blob/ddbf1ac2e38415a48c89f09238ad9b28d0c3fa68/spec/util/quick.lua#L133](https://github.com/luarocks/luarocks/blob/ddbf1ac2e38415a48c89f09238ad9b28d0c3fa68/spec/util/quick.lua#L133)

the control variable, `line`, which is read-only for Lua 5.5, is being modified at

[https://github.com/luarocks/luarocks/blob/ddbf1ac2e38415a48c89f09238ad9b28d0c3fa68/spec/util/quick.lua#L277](https://github.com/luarocks/luarocks/blob/ddbf1ac2e38415a48c89f09238ad9b28d0c3fa68/spec/util/quick.lua#L277)

## Fix for the issue

The fix is pretty simple. Just set the control variable a different name, and place a `local` attribution right after the `do` keyword setting the old variable. In short, the patch below should fix it:

```patch
diff -Naur luarocks-orig/spec/util/quick.lua luarocks/spec/util/quick.lua
--- luarocks-orig/spec/util/quick.lua	2026-07-13 19:50:25.000000000 -0300
+++ luarocks/spec/util/quick.lua	2026-07-15 07:10:44.377485665 -0300
@@ -130,7 +130,8 @@
       input = input .. "\n"
    end
 
-   for line in input:gmatch("([^\r\n]*)\r?\n?") do
+   for inputline in input:gmatch("([^\r\n]*)\r?\n?") do
+      local line = inputline
       cur_line = cur_line + 1
 
       local state = stack[#stack]
```

## Conclusion

IMO, to keep things as clean as possible, I think it makes sense to open a separate PR for such issue.

Since you were the one that spotted this issue, can you please open a PR providing a fix for this control variable issue missed during LuaRocks 3.13.0 release?

cclauss:
macOS and Ubuntu are blocked by:
* #1910

cclauss:
macOS and Ubuntu fail with: ./spec/util/test_env.lua:695: failed building test dependency: lua-bz2

Windows fails with: .//spec\util\test_env.lua:695: failed building test dependency: lua-zlib
> lua-zlib 1.4-0 depends on lua >= 5.1, <= 5.5 (5.1-1 provided by VM: success)

luau-project:
> Windows fails with: .//spec\util\test_env.lua:695: failed building test dependency: lua-zlib
> 
> > lua-zlib 1.4-0 depends on lua >= 5.1, <= 5.5 (5.1-1 provided by VM: success)

Unfortunately, zlib 1.3.2 has created problems for people building from source code (*I have a hot open issue on zlib repo waiting for fixes since 1.3.2 was released*), but Mark Adler seems to publish releases from 2 to 3 years. So, the easiest solution is usually to set zlib back to 1.3.1.

luau-project:
This PR branch had poisoned cache entries for zlib here. I deleted the poisoned cache entry.

Then I visited your fork and saw that you also have a poisoned cache entry for zlib in the `main` branch of your fork. (tip: *the one with 178 bytes is a poisoned entry*)

cclauss:
Thank you.  I would not have found this on my own.

Do you know of any docs that I should read to learn how to recognize cache poisoning?

How should I fix my local branch?  Is this git cache?  Should I run `luarocks make` locally before pushing?

Is this related to https://docs.zizmor.sh/audits/#cache-poisoning ?

luau-project:
> Do you know of any docs that I should read to learn how to recognize cache poisoning?

I think it depends on the situation. I faced such cache poisoned entries for `zlib` investigating the problem described at [https://github.com/luarocks/luarocks/pull/1906](https://github.com/luarocks/luarocks/pull/1906).

IMO, if you download an artifact (e.g.: `zip` or `.tar.gz`) and the tools fail to extract, as happened with both of us, there is something unusual with the artifact. In the first time, I copied the exact command and reproduced on my machine. Then tried to extract and failed.

Comparing the expected archive size against the one you downloaded is a good indicator.

> Is this related to https://docs.zizmor.sh/audits/#cache-poisoning ?

Not exactly, because the poisoned entries we were getting was due the missing `-L` flag calling `curl` to download zlib tarball. In the zizmor article, it refers mostly to attacking scenarios, while ours were legit tarball providers. In other words, it was our fault to miss `-L`, not someone tricking us.

PR #1890, opened 2026-07-04
ci: Upgrade Windows OpenSSL from v3.6.2 to 3.6.3

cclauss:
```diff
-     LUAROCKS_DEPS_OPENSSL_VER: "3.6.2"
+     LUAROCKS_DEPS_OPENSSL_VER: "3.6.3"
```
Fix `WindowsTests` failures found in the `Download OpenSSL installer` job step.
* https://github.com/luarocks/luarocks/actions/runs/28606791659/job/85072099389?pr=1888
>  $installerName = "D:\a\luarocks\luarocks\build\openssl-installer-3.6.2-x64.exe";
[7](https://github.com/luarocks/luarocks/actions/runs/28606791659/job/85072099389?pr=1888#step:14:108)
Exception: D:\a\_temp\2b717ea9-6118-491a-b67d-e709f7d54951.ps1:37
Line |
  37 |    throw "Installer not found for version ${version}. Please, update O …
     |    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     | Installer not found for version 3.6.2. Please, update OpenSSL to the latest version found at
     | https://raw.githubusercontent.com/slproweb/opensslhashes/master/win32_openssl_hashes.json
Error: Process completed with exit code 1.

[win32_openssl_hashes.json](https://raw.githubusercontent.com/slproweb/opensslhashes/master/win32_openssl_hashes.json) does not contain `3.6.2` but does contain both `3.6.3` and `4.0.1`

PR #1891, opened 2026-07-05
ci: Add curl retry on zlib in binary/Makefile.windows

cclauss:
Add cURL retry with exponential backoff for a file that occasionally causes CI to fail.
`-f` fails on errors instead of saving an error page which `tar` rejects (gzip: stdin: not in gzip format).
* https://curl.se/docs/manpage.html#--fail

`--retry 5` gives us 5 tries with exponential backoff.
* https://curl.se/docs/manpage.html#--retry

`binary/Makefile.windows`:
```diff
- curl -OL https://www.zlib.net/fossils/zlib-$(ZLIB_VERSION).tar.gz
+ curl -vOL --fail-with-body --retry 5 https://www.zlib.net/fossils/zlib-$(ZLIB_VERSION).tar.gz
```
Fixes the one failing GitHub Action test in: https://github.com/luarocks/luarocks/pull/1890/checks

### DIFF
--- a/binary/Makefile.windows
+++ b/binary/Makefile.windows
@@ -46,7 +46,7 @@ $(WINDOWS_DEPS_DIR)/lib/libssl.a: $(BUILD_WINDOWS_DEPS_DIR)/openssl-$(OPENSSL_VE
 
 $(BUILD_WINDOWS_DEPS_DIR)/zlib-$(ZLIB_VERSION).tar.gz:
 	mkdir -p $(@D)
-	cd $(BUILD_WINDOWS_DEPS_DIR) && curl -OL https://www.zlib.net/fossils/zlib-$(ZLIB_VERSION).tar.gz
+	cd $(BUILD_WINDOWS_DEPS_DIR) && curl -vOL --fail-with-body --retry 5 https://www.zlib.net/fossils/zlib-$(ZLIB_VERSION).tar.gz
 $(BUILD_WINDOWS_DEPS_DIR)/zlib-$(ZLIB_VERSION): $(BUILD_WINDOWS_DEPS_DIR)/zlib-$(ZLIB_VERSION).tar.gz
 	cd $(BUILD_WINDOWS_DEPS_DIR) && tar zxvpf zlib-$(ZLIB_VERSION).tar.gz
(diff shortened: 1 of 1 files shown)


luau-project:
I think this bump on zlib version needs further investigation, specially due API breaks discussed in [https://github.com/madler/zlib/issues/1202](https://github.com/madler/zlib/issues/1202). Unfortunately, zlib 1.3.2 had a few breaking changes for Windows users (both MSVC and MinGW-w64 toolchains).

> [!NOTE]
> 
> The issue reporter (`lhmouse`) is one of MinGW-w64 collaborators, and we cross-compile 32-bit `luarocks.exe` and `luarocks-admin.exe` using 32-bit MinGW-w64 built by Ubuntu.

cclauss:
After further examination, the problem seems to be more about cURL timeouts than the version.

These changes try to fix that instead.

cclauss:
I was able to get this to fail once with `--fail --retry 5`:
* https://github.com/cclauss/luarocks/actions/runs/28750790783/job/85249378636

% `curl -fOL --retry 5 --retry-delay 0 https://www.zlib.net/fossils/zlib-1.3.2.tar.gz`
```
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0   185    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (22) The requested URL returned error: 415
```
https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/415 ???

---
Also, with verbose and fail-with-body: `curl -vOL --fail-with-body --retry 5 ...`
https://github.com/cclauss/luarocks/actions/runs/28775561660/job/85318618524?pr=1

https://curl.se/docs/manpage.html#--verbose

cclauss:
Unfortunately, we still have intermittent failures, but at least we fail on the cURL download command and not on the gzip command that follows.

Also, we have more complete error messages.

hishamhm:
@cclauss Thank you for looking into this. In cases like this, I think the solution is either to mirror the assets somewhere more reliable (or at least as reliable as GitHub itself, such as another github repo) or to use some caching action to reduce pressure on upstream.

PR #1894, opened 2026-07-13
Fix rockspec files created by the init and write_rockspec commands

Klafyvel:
Hi! The serialization of the Rockspec object was done wrong previously, which added an invalid `queries` item in the dependency table, and the generated file was sometimes invalid, using rockspec 3.0 elements when generating a rockspec 1.0 file.


hishamhm:
@Klafyvel thank you!

Klafyvel:
My pleasure :)

PR #1895, opened 2026-07-14
Add documentation for the init command

Klafyvel:
I just realized it was missing!

### DIFF
--- a/docs/luarocks_init.md
+++ b/docs/luarocks_init.md
@@ -1,0 +1,54 @@
+# luarocks init
+
+
+Initialize a directory for a Lua project using LuaRocks.
+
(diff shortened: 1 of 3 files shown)


docs/luarocks_init.md
```diff
@@ -0,0 +1,52 @@
+# luarocks init
+
+
+Initialize a directory for a Lua project using LuaRocks.
```
Tieske, line 4: I think this particular command deseves some extra explanation about the project local rockstree etc. Explaining its overall usecase.

Klafyvel: I will see what I can do, probably linking also to https://github.com/luarocks/luarocks/blob/main/docs/rocks_repositories.md

hishamhm:
We can probably merge this as-is as a starting point and then get a follow up PR, but right now GitHub is saying this PR needs to be rebased locally to fix merge conflicts. @Klafyvel could you take a look at that? Thanks!!

docs/luarocks_init.md
Klafyvel: I added a longer description below :)

Klafyvel:
Hi @hishamhm , sorry, I missed the notification. I rebased and improved the description of the command.

PR #1897, opened 2026-07-16
lua5.5: loop control variables are const in quick_parse

cclauss:
[Lua v5.5 docs](https://www.lua.org/manual/5.5/manual.html#8.1):
> The control variable in __for__ loops is read-only. If you need to change it, declare a local variable with the same name in the loop body. 

Discovered by running the tests on Lua v5.5 in:
* #1889 

Related to:
* #1844

### DIFF
--- a/spec/util/quick.lua
+++ b/spec/util/quick.lua
@@ -131,6 +131,7 @@ local function parse(filename)
    end
 
    for line in input:gmatch("([^\r\n]*)\r?\n?") do
+      local line = line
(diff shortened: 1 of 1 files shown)


cclauss:
The test failure is yet another proof of the need for:
* #1891

> gzip: stdin: not in gzip format

Is painful because we are trying to unzip a text error message.

luau-project (approved):

Issue #1896, opened 2026-07-14
Add `--dry-run` parameter to `luarocks upload` command


luau-project:
## Quick description

Add `--dry-run` parameter to `luarocks upload` command.

## Motivation

> [!NOTE]
> 
> I didn't find such parameter on [`luarocks upload` docs](https://github.com/luarocks/luarocks/blob/main/docs/luarocks_upload.md).

The main reason for this feature is to aid the development of a newer version for a Lua module. It is usually the last step before publishing a package.

This feature aligns with other package managers:

* `npm publish --dry-run` for nodejs ( [https://docs.npmjs.com/cli/v12/commands/npm-publish#dry-run](https://docs.npmjs.com/cli/v12/commands/npm-publish#dry-run) )
* `uv publish --dry-run` for Python ( [https://docs.astral.sh/uv/reference/cli/#uv-publish--dry-run](https://docs.astral.sh/uv/reference/cli/#uv-publish--dry-run) )

## What should a dry run do?

On most services that accept file uploads, `--dry-run` does all the necessary steps (e.g.: local validation, etc) to upload, but the essential behavior is that it does **NOT** perform the actual file upload. In case of LuaRocks, it should not perform the rock / rockspec upload.

> [!TIP]
> 
> Even though I think there is no consensus about the output, on some services that accept file uploads (e.g.: coveralls), the dry run mode writes the raw http request (masking secrets) to stdout.

hishamhm:
PR welcome!

luau-project:
Completed in #1899

PR #1899, opened 2026-07-18
feat(upload): add a dry-run flag

Tieske:
implements #1896


luau-project:
@Tieske wow, that's a great feature addition! Thank you for taking the time to implement this.

As an early read for this draft version, I have two suggestions:

1. In the upload spec, wouldn't be safer to assert against `check_rockspec` endpoint? as in

    ```lua
    api:method("check_rockspec", {
       package = rockspec.package,
       version = rockspec.version
    })
    ```

    I think that a check against output may be unsafe. For instance, I'm not saying it does, but in case an unintended change make them appears in the same control-flow for the usual upload, we would be in trouble;

2. Following the idea of the `check_rockspec` endpoint, I think there are two kind of essential tests to do:

    1. Let a package named `my-package` to **exist** in the mock server. Then running `luarocks upload --dry-run` for `my-package` should not include a new version in the server;
    2. Let a package named `my-new-package` to **NOT exist** in the mock server. Then running `luarocks upload --dry-run` for `my-new-package` should not create the package in the server.

Tieske:
I think that would indeed be the best approach, but...

>        I think that a check against output may be unsafe. For instance, I'm not saying it does, but in case an unintended change make them appears in the same control-flow for the usual upload, we would be in trouble;

I get the point, but the output mainly checks on strings "would upload" or "dry run", for which probability of ever appearing in regular output would be near 0 imho.

> 
>     2. Following the idea of the `check_rockspec` endpoint, I think there are two kind of essential tests to do:
>        
>        1. Let a package named `my-package` to **exist** in the mock server. Then running `luarocks upload --dry-run` for `my-package` should not include a new version in the server;
>        2. Let a package named `my-new-package` to **NOT exist** in the mock server. Then running `luarocks upload --dry-run` for `my-new-package` should not create the package in the server.

afaict the mock server serves static content and needs changes. Not worth it imho in combination with the other remark above on strings in output.

luau-project:
> afaict the mock server serves static content and needs changes.

I was not aware of that.

Reading the code, I can see that it is taking a different branch in the control-flow to skip the upload.

Thank you for the implementation @Tieske :rocket:

Issue #1884, opened 2026-05-23
[FAIL to install] The LuaBitop package gets failing on gcc compiler for Macos


ToXP:
> https://luarocks.org/modules/luarocks/luabitop
> 
> Done! Thanks for the heads up! 

 _Originally posted by @hishamhm in [#1882](https://github.com/luarocks/luarocks/issues/1882#issuecomment-4502865580)_

[FOLLOW UP!]

Hello,
unfortunately the attempt to install the LuaBitop on Luarocks fails with the following setup:

 >  Lua vs. 5.5
 >   Luarocks vs. 3.13.0
 >   gcc vs. 14.3.0
 >   MACOS vs. 10.15.8

When launching the shell command:

_$ luarocks install luabitop_

It produces the error trace:

`luabitop 1.0.3-1 depends on lua >= 5.1, < 5.6 (5.5-1 provided by VM: success) env MACOSX_DEPLOYMENT_TARGET=10.8 gcc -O2 -fPIC -I/usr/local/opt/lua/include/lua5.5 -c bit.c -o bit.o bit.c: In function 'barg': bit.c:76:2: error: #error "Unknown number type, check LUA_NUMBER_* in luaconf.h" 76 | #error "Unknown number type, check LUA_NUMBER_* in luaconf.h"`

The first and obvious first attempt was to check the correctness of the installed _luaconf.h_ file in dir: 
_/usr/local/opt/lua/include/lua5.5_

...which I verified to be correct at the Lua official page:
[https://www.lua.org/source/5.5/luaconf.h.html](url)

...then I tried to look for a solution on Stackoverflow, but with no success.
Any hint?

TX!

luau-project:
I managed to reproduce your issue on both macOS and Linux on Lua 5.4 and Lua 5.5. Through a quick inspection on `LuaBitOp` sources, I came to conclusion that it is a `LuaBitOp` issue.

Since LuaBitOp tarball is hosted on LuaJIT's website, can you please post this issue on LuaJIT's ([https://github.com/LuaJIT/LuaJIT](https://github.com/LuaJIT/LuaJIT)) repo?

ToXP:
Hello, 

I received info from the LuaJIT devs that **another** _NEWER_ vs. of the LuaBitOp package is available: 
[https://bitop.luajit.org/download.html](https://bitop.luajit.org/download.html)

[...]

PR #1901, opened 2026-07-19
ci: Can a package be installed on Lua 5.4 and 5.5 on three OSes?

cclauss:
Add an on-demand workflow to test whether a package on https://luarocks.org can be installed on both Lua 5.4 and 5.5 on macOS, Ubuntu, and Windows.
 
Unlike most other workflows, this one is NOT run on pushes and pull requests.  Instead, it is run only ON DEMAND by MAINTAINERS of this repo when they want to know whether a Lua package can be installed on both Lua 5.4 and 5.5.

* Go to https://github.com/luarocks/luarocks/actions/workflows/luarocks_install.yml
* Click the "Run workflow" button at the upper right of that page.
    * You must be a MAINTAINER of this repo for that button to appear.
* Enter the name of the package on https://luarocks.org to be tested.

That package is then installed on both Lua 5.4 and 5.5 on macOS, Ubuntu, and Windows.

image

### DIFF
--- a/.github/workflows/luarocks_install.yml
+++ b/.github/workflows/luarocks_install.yml
@@ -1,0 +1,48 @@
+# Can a package be installed on Lua 5.4 and 5.5 on macOS, Ubuntu, and Windows?
+# Unlike most other workflows, this one is NOT run on pushes and pull requests.
+#
+# This workflow is only run ON DEMAND by MAINTAINERS of this repo when they
+# want to know if a Lua package can be installed on both Lua 5.4 and 5.5.
+#
+# https://github.com/luarocks/luarocks/actions/workflows/luarocks_install.yml
+# * Click the "Run workflow" button at the upper right of the page.
+#     * You must be a MAINTAINER of this repo for that button to appear.
+# * Enter the name of the package on https://luarocks.org to be tested.
+#
+# That package is installed on both Lua 5.4 and 5.5 on macOS, Ubuntu, and Windows.
+
+name: luarocks install
+run-name: luarocks install ${{ inputs.package }}
+on:
+  workflow_dispatch:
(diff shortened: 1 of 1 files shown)


Tieske:
What's the use case here? why would luarocks maintainers test other peoples packages?

Also I doubt whether it works, many packages are build for unix only for example, or on Windows for a specific toolchain. So how would luarocks maintainers be handling that?

cclauss:
> What is the use case here?

[18 issues mention `5.5`](https://github.com/luarocks/luarocks/issues?q=is%3Aissue%205.5) so maintainers can do bespoke work when trying to replicate and triage each of these issues, or they can run a handy quick test.  Python 3.15 is currently in its final beta, so I use a similar workflow to discover which dependencies are ready and which need porting.  In the Lua space, the packages `lyaml`, `optparse`, and `specl` all have failing workflows for lua5.5 so the maintainer knows when the port is complete.  These workflows were created manually, but the `workflow_dispatch` approach is better for quick trials.

Of course, when new Lua versions are released, the version matrix can be updated, and the process is repeated.

> many packages are built for Unix only

This utility is not on the critical path, so irrelevant failures do not slow down the process.  For a Unix-only package, I would merely ignore all macOS and Windows failures (or successes).

---
Examples:
* Test issue #1884 because [`LuaBitOp v1.0.3`](https://luarocks.org/modules/luarocks/luabitop) != [`v1.0.4`](https://bitop.luajit.org/download.html) the [tests are red](https://github.com/cclauss/itinerant-tester/actions/runs/29882038952), but should go green when rerun after the v1.0.4 upload to Luarocks.
* Test `luaposix` to discover that #1884 is the initial blocker for `luaposix` in #1889.

PR #1903, opened 2026-07-21
chore(ci): bump deps and other minor changes

luau-project:
Changes worth to note:

* following `leafo/gh-actions-lua`, `luarocks/gh-actions-lua` and `luarocks/gh-actions-luarocks`, changed the action to setup MSVC from `ilammy/msvc-dev-cmd` to `step-security/msvc-dev-cmd` due warnings related to the use of EOL nodejs versions;
* replaced actions from `hishamhm` acc to `luarocks` org acc.

### DIFF
--- a/.github/workflows/pr-gh-actions.yml
+++ b/.github/workflows/pr-gh-actions.yml
@@ -51,7 +51,7 @@ jobs:
 
     steps:
 
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           repository: ${{ env.TARGET_REPOSITORY }}
           token: ${{ secrets.GH_ACTIONS_LUAROCKS_TOKEN }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -79,7 +79,7 @@ jobs:
       run: luacov -c "testrun/luacov.config"
 
     - name: Setup Python
-      uses: actions/setup-python@v5
+      uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
       with:
         python-version: '3.11'
 
@@ -522,7 +522,7 @@ jobs:
           )
 
       - name: Setup Python
-        uses: actions/setup-python@v5
+        uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
         with:
           # employ the latest Python version
           # described at https://pypi.org/project/hererocks/
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -81,7 +81,7 @@ jobs:
     - name: Setup Python
       uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
       with:
-        python-version: '3.11'
+        python-version: '3.14'
 
     - name: Install Codecov
       run: pip install codecov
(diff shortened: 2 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -79,9 +79,9 @@ jobs:
       run: luacov -c "testrun/luacov.config"
 
     - name: Setup Python
-      uses: actions/setup-python@v5
+      uses: actions/setup-python@v7
       with:
-        python-version: '3.11'
+        python-version: '3.14'
```
cclauss, line 84: I like 3.x here so that GitHub Actions finds bugs in new Python releases (each October) before users do.

.github/workflows/test.yml
```diff
@@ -522,11 +522,9 @@ jobs:
           )
 
       - name: Setup Python
-        uses: actions/setup-python@v5
+        uses: actions/setup-python@v7
         with:
-          # employ the latest Python version
-          # described at https://pypi.org/project/hererocks/
-          python-version: '3.11'
+          python-version: '3.14'
```
cclauss, line 527: 3.x?

cclauss (commented):
Nice changes.  I approve.

.github/workflows/pr-gh-actions.yml
```diff
@@ -51,7 +51,7 @@ jobs:
 
     steps:
 
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@main
```
hishamhm: I think the current recommendation for GitHub Actions in general is to avoid using moving-target versions such as `@main`. 

```suggestion
      - uses: actions/checkout@v7
```

.github/workflows/test.yml
hishamhm: We need to make sure that the Python version used here is compatible with the one used by hererocks (according the the comment that this PR is deleting... (why?) )

hishamhm:
@luau-project thanks for the cleanup! I left some comments

.github/workflows/test.yml
luau-project: I just went to Python's website and looked for the latest release. Then, it is 3.14. The role of Python on LuaRocks is pretty small. AFAICT, Python serves to:

* fetch a recent SSL cert
* to run `hererocks` in order to install Lua + LuaRocks on Windows.

I don't have any objections to set it to `3.X`, because `hererocks` is very conservative and works on almost all Python versions. It's up to `hishamhm`, really.

.github/workflows/test.yml
luau-project: like I commented to `cclauss` above, `hererocks` is very conservative. A few releases ago, they used to support even Python 2. I think it is unlikely to break on new versions. If you want, I can restore such comments.

.github/workflows/pr-gh-actions.yml
luau-project: in my personal taste, I prefer to pin each action to the latest release's `sha`. v7 is also a moving tag within the v7.X.Y series.

Do you want to pin to the latest `sha`? It would make the process to setup dependabot easier, in case it is wanted.

hishamhm: The sha for the latest v7 tagged version would be fine.

.github/workflows/test.yml
hishamhm: Let's keep it 3.14 then, so that any upgrades can be tested prior to applying.

luau-project:
> The sha for the latest v7 tagged version would be fine.

not sure if I understood it correctly, but I pinned each action usage to the latest tag/release. If you wanted only the `actions/checkout` to be pinned, I still have a local copy of the old branch to restore the previous actions to unpinned mode.

I saw that I screwed up the commit message for the 'pin actions to save and restore'. Maybe it is easier to fix while merging in order to avoid an entire CI run again.

hishamhm:
@luau-project looks good, thank you!

PR #1906, opened 2026-08-07
fix(ci): use `-L` flag with `curl` to follow redirects

luau-project:
## Description

Fixes the download of artifacts with `curl` by adding `-L` flag to follow redirects. At the moment, it was downloading an empty tarball for zlib, and thus CI is failing.

> [!IMPORTANT]
> 
> The actions cache were poisoned by empty cache entries for `zlib` tarball due the lack of `-L` flag to follow redirects. I deleted such poisoned entries manually to test these changes (I did it on my fork and it worked).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -279,7 +279,7 @@ jobs:
       - name: Download zlib
         if: ${{ matrix.COMPILER == 'vs' && steps.restore-zlib-tarball.outputs.cache-hit != 'true' }}
         run: |
-          curl -o "${{ github.workspace }}\zlib-${{ env.LUAROCKS_DEPS_ZLIB_VER }}.tar.gz" ^
+          curl -Lo "${{ github.workspace }}\zlib-${{ env.LUAROCKS_DEPS_ZLIB_VER }}.tar.gz" ^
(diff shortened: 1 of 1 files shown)


Tieske:
Is this the right solution? The nice thing about not having -L is that it fails loud. Since in the long run, at some point also the redirects will be gone, since they are a transitional thing usually.

If adding -L, shouldn't we then at least in this PR also update the location (since the current hardcoded one is outdated)?

luau-project:
> Is this the right solution? The nice thing about not having -L is that it fails loud. Since in the long run, at some point also the redirects will be gone, since they are a transitional thing usually.

as long as the URL is legit, I don't see any issues following redirects.

In my experience, some services (e.g.: `sourceforge`) host their tarballs behind a chain of redirects (probably due CDN).

> If adding -L, shouldn't we then at least in this PR also update the location (since the current hardcoded one is outdated)?

The zlib URL affected by this PR, which was failing during the last few runs, is from a release tarball of the upstream zlib repository.

If there are security concerns and `hishamhm` wants, I think we can improve the process through the addition of a checksum verification (.e.g.: SHA256) for each downloaded archive.

PR #1907, opened 2026-08-08
ci: Use recommended way to detect the operating system

cclauss:
GitHub Actions docs: [`Detecting the operating system`](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-variables#detecting-the-operating-system)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -51,11 +51,11 @@ jobs:
         luaRocksVersion: "3.13.0"
 
     - name: 'Setup macOS deps'
-      if: ${{ contains(matrix.os, 'macos') }}
+      if: runner.os == 'macOS'
       run: brew install openssl bzip2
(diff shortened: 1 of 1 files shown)


luau-project (approved):
LGTM

@cclauss

I know the syntax `${{ }}` within `if:` directive is optional in simple checks like this, but I think it helps to keep consistency with all the uses of `if:` in the same workflow.

**for reviewers**: possible `runner.os` values are described at [https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#runner-context](https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#runner-context)

cclauss:
There important Actions security rules for not encouraging contributors to use that syntax everywhere.
https://docs.zizmor.sh/audits/#template-injection

PR #1909, opened 2026-08-09
ci: GitHub Actions zizmor remove persist-credentials

cclauss:
[zizmor](https://docs.zizmor.sh) is a static analysis tool that can find and fix security issues in common CI/CD setups, including GitHub Actions, Dependabot, and pre-commit.

This PR fixes https://docs.zizmor.sh/audits/#artipacked

Zizmor finds other issues, but let's leave those for a separate pull request.  Careful review of `pr-gh-actions.yml` please, because it performs git operations that may require `persist-credentials` ~but I think not~.

### DIFF
--- a/.github/workflows/pr-gh-actions.yml
+++ b/.github/workflows/pr-gh-actions.yml
@@ -56,7 +56,7 @@ jobs:
           repository: ${{ env.TARGET_REPOSITORY }}
           token: ${{ secrets.GH_ACTIONS_LUAROCKS_TOKEN }}
           path: gh-actions-luarocks
-          persist-credentials: false
+          persist-credentials: true  # Needed for git push.
 
       - name: Create a new branch
         working-directory: gh-actions-luarocks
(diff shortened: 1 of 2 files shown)


.github/workflows/pr-gh-actions.yml
```diff
@@ -56,6 +56,7 @@ jobs:
           repository: ${{ env.TARGET_REPOSITORY }}
           token: ${{ secrets.GH_ACTIONS_LUAROCKS_TOKEN }}
           path: gh-actions-luarocks
+          persist-credentials: false
```
alerque: This one at least won't work, it will block the `git push` that happens later. Before opening the pull request it has to create the remote branch, and if the API token it used to clone isn't retained it won't be able to push.

cclauss: Thanks much for the review!
```suggestion
          persist-credentials: true  # Needed for git push.
```

Issue #1829, opened 2025-08-25
3.12.2 fails to install packages on Windows


magneto538:
Presumably connected to https://github.com/luarocks/luarocks/issues/1828 as the logged output is identical
 
- **Platform:**
  - `win32-x86_64` on GitHub Actions agent

- **LuaRocks version:**
  - 3.12.2

- **Configuration file:**
  - default config

- **LuaRocks output from when the issue occurred:**
```
Error: LuaRocks 3.12.2 bug (please report at https://github.com/luarocks/luarocks/issues).
Arch.: win32-x86_64 (fs_use_modules = true) (binary)
[string "src/luarocks/fetch.lua"]:144: attempt to concatenate a nil value (local 'name')
stack traceback:
	[string "src/luarocks/fetch.lua"]:144: in upvalue 'download_with_mirrors'
	[string "src/luarocks/fetch.lua"]:201: in function 'luarocks.fetch.fetch_url'
	[string "src/luarocks/fetch.lua"]:90: in function 'luarocks.fetch.fetch_caching'
	[string "src/luarocks/fetch.lua"]:248: in function 'luarocks.fetch.fetch_url_at_temp_dir'
	[string "src/luarocks/fetch.lua"]:352: in function 'luarocks.fetch.fetch_and_unpack_rock'
	[string "src/luarocks/cmd/build.lua"]:66: in upvalue 'build_rock'
	[string "src/luarocks/cmd/build.lua"]:125: in upvalue 'do_build'
	[string "src/luarocks/cmd/build.lua"]:171: in function 'luarocks.cmd.build.command'
	(...tail calls...)
	[C]: in function 'xpcall'
	[string "src/luarocks/cmd.lua"]:794: in function 'luarocks.cmd.run_command'
	[string "luarocks"]:34: in main chunk
	[C]: in ?

```

- **Description:**
  - Install Luarocks 3.12.2 on a GitHub Actions agent via `luarocks/gh-actions-luarocks@v6`
  - Attempt to install a package
  - Observe the error occurring

luau-project:
There are many ways to install LuaRocks on Windows. Please, provide more info to help the reproduction of your issue:

* Which method did you use to install LuaRocks?
    1. standalone LuaRocks
        * 32-bit LuaRocks from [https://luarocks.github.io/luarocks/releases/luarocks-3.12.2-windows-32.zip](https://luarocks.github.io/luarocks/releases/luarocks-3.12.2-windows-32.zip);
[...]

Issue #1828, opened 2025-08-23
3.12.2 fails to install packages on macOS


vadi2:
### Please fill out the following:
- **Platform:**
  - MacOS.

- **LuaRocks version:**
  - `luarocks --version`: 3.12.2

- **Configuration file:**
  - default confoguration

- **LuaRocks output from when the issue occurred:**
```
Error: LuaRocks 3.12.2 bug (please report at https://github.com/luarocks/luarocks/issues).
Arch.: macosx-x86_64
.../Cellar/luarocks/3.12.2/share/lua/5.4/luarocks/fetch.lua:144: attempt to concatenate a nil value (local 'name')
Installing https://luarocks.org/luafilesystem-1.8.0-1.src.rock
stack traceback:
	.../Cellar/luarocks/3.12.2/share/lua/5.4/luarocks/fetch.lua:144: in upvalue 'download_with_mirrors'
	.../Cellar/luarocks/3.12.2/share/lua/5.4/luarocks/fetch.lua:201: in function 'luarocks.fetch.fetch_url'
	.../Cellar/luarocks/3.12.2/share/lua/5.4/luarocks/fetch.lua:90: in function 'luarocks.fetch.fetch_caching'
	.../Cellar/luarocks/3.12.2/share/lua/5.4/luarocks/fetch.lua:248: in function 'luarocks.fetch.fetch_url_at_temp_dir'
	.../Cellar/luarocks/3.12.2/share/lua/5.4/luarocks/fetch.lua:352: in function 'luarocks.fetch.fetch_and_unpack_rock'
	...lar/luarocks/3.12.2/share/lua/5.4/luarocks/cmd/build.lua:66: in upvalue 'build_rock'
	...lar/luarocks/3.12.2/share/lua/5.4/luarocks/cmd/build.lua:125: in upvalue 'do_build'
	...lar/luarocks/3.12.2/share/lua/5.4/luarocks/cmd/build.lua:171: in function 'luarocks.cmd.build.command'
	(...tail calls...)
	[C]: in function 'xpcall'
	...al/Cellar/luarocks/3.12.2/share/lua/5.4/luarocks/cmd.lua:794: in function 'luarocks.cmd.run_command'
	/usr/local/bin/luarocks:38: in main chunk
	[C]: in ?
Error: Process completed with exit code 99.
```
- **Description:**
[...]

PR #1911, opened 2026-08-12
download_with_mirrors: Safely build errors with falsely names

cclauss:
The variable `name` must be falsely in this `else` clause so stringify values like `false`,  `nil `,  `{}`, etc., so we can safely build an `err` message instead of merely crashing.
* #1828
* #1829


hishamhm:
Could you add the edit to the .tl source as well in the PR? Thank you!

cclauss:
Thanks, yes.

hishamhm (approved):

hishamhm:
@cclauss Thank you for these fixes and updates!!

